PR #3, opened 2016-11-02
Added observer + callback for unit.run.

petevg:
Allows us to wait until an action has actually been completed before
returning a result.

@tvansteenburgh @johnsca I'm not sure whether this is the right approach. Should we push responsibility to manage the observer to the code making the call? Or should we generalize some of the techniques here, and push them further into the codebase? (Also, it's intuitive to me to return stdout, stderr, and code for what is essentially a remote bash command, but I'm open to changing what we return.)

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -535,6 +535,30 @@ class Model(object):
             if o.cares_about(delta):
                 asyncio.ensure_future(o(delta, old_obj, new_obj, self))
 
+    async def _wait(self, entity_type, entity_id, action, predicate=None):
+        """
+        Block the calling routine until a given action has happened to the
+        given entity
+
+        :param entity_type: The entity's type.
+        :param entity_id: The entity's id.
+        :param action: the type of action (e.g., 'add' or 'change')
+        :param predicate: optional callable that must take as an
+            argument a delta, and must return a boolean, indicating
+            whether the delta contains the specific action we're looking
+            for. For example, you might check to see whether a 'change'
+            has a 'completed' status. See the _Observer class for details.
+
+        """
+        q = asyncio.Queue(loop=self.loop)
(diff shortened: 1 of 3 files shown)


juju/unit.py
```diff
@@ -108,21 +109,58 @@ def resolved(self, retry=False):
...
         """
         action = client.ActionFacade()
         action.connect(self.connection)
 
         log.debug(
             'Running `%s` on %s', command, self.name)
 
-        # TODO this should return an Action
-        return await action.Run(
+        action_status = asyncio.Queue(loop=self.model.loop)
```
tvansteenburgh: While this approach works, we have a helper in place to make it more concise. Something like this;

```python
res = await action.Run(
    [],
    command,
    [],
    timeout,
    [self.name],
)
tag = res.results[0].action.tag

def predicate(delta):
    if delta.data['status'] in ('completed', 'error'):
        return True

return await self.model._wait_for_new('action', tag, predicate)
```

There are some other examples of _wait_for_new() in the codebase.

petevg: Aha. I missed that I could pass the tag to _wait_for_new as the entity_id. That makes a lot of sense, and definitely saves on all my boilerplate. Making the change locally now. Thx :-)

juju/model.py
```diff
@@ -543,14 +567,20 @@ def _stop_watching(self):
         This coroutine blocks until the new object appears in the model.
 
         """
-        entity_added = asyncio.Queue(loop=self.loop)
+        return await self._wait(entity_type, entity_id, predicate)
```
tvansteenburgh: Shouldn't this be:

    return await self._wait(entity_type, entity_id, 'add', predicate)

PR #11, opened 2016-11-17
Added juju.model.connect_model()

sfeole:
Added juju.model.connect_model(arg) 

Allowing the developer to test against an already existing model 

Example Usage:

    from juju.model import Model

    model = Model()
    await model.connect_model("mycontroller:admin/testmodel")


tvansteenburgh:
@sfeole Thanks!

PR #12, opened 2016-11-17
Adding comment in example/controller.py regarding cloudcred format

sfeole:
Possible Typo in the Example/Controller.py
I updated cloud-creds:  :s/@/_

'cloudcred-aws_tvansteenburgh_external_aws-tim',

For LXD I used:
'cloudcred-localhost_admin_default'

'cloudcred-localhost_admin@default' fails, unless the @ is intentional for AWS

### DIFF
--- a/examples/controller.py
+++ b/examples/controller.py
@@ -27,7 +27,7 @@ async def run():
     model = await controller.add_model(
         'libjuju-test',
         'cloud-aws',
-        'cloudcred-aws_tvansteenburgh_external@aws-tim',
+        'cloudcred-aws_tvansteenburgh@external_aws-tim',
     )
     await model.deploy(
         'ubuntu-0',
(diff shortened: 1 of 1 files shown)


tvansteenburgh:
Yeah this is a great observation, but it's not actually a typo. In my scenario it was required (shared controller on aws). At least I think it was required - didn't try it w/o the '@external'. But at the very least, it worked. Maybe change the PR to add a comment in the example about the cloudcred format instead, to assist future users?

sfeole:
Hey Tim, I looked at this again, You are infact following the cloudcred format as in: 
cloudcred-<cloudname>_<user>_<credentialname>
The '@' completely threw me off :) 

I updated the example comments to add a 2 line note there, if that works?

sfeole:
I think this is something that documentation would fix, when that becomes available.

tvansteenburgh:
@sfeole Thanks!

PR #13, opened 2016-11-21
Added .is_leader method to Unit class.

petevg:
As leadership is not exposed via deltas in the current version of the
juju websocket api, we have to spin up a ClientFacade and dig through
FullStatus for the information we need.

A ticket for adding leadership info to the deltas lives here:
https://bugs.launchpad.net/juju/+bug/1643691

@tvansteenburgh @johnsca (Gave the method a funky name because I don't want it to collide with .is_leader in the future, which will be called without an `await`)

### DIFF
--- a/examples/leadership.py
+++ b/examples/leadership.py
@@ -1,0 +1,26 @@
+"""
+This example:
+
+1. Connects to the current model.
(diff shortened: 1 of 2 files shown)


tvansteenburgh:
lgtm, thanks @petevg

PR #19, opened 2016-11-22
added juju.controller.destroy_models()

sfeole:
This small update adds the ability to destroy models. I've been using this in my qa framework throughout the day, it seems to be working as expected.

This was tested using the the following snippet from a script:

```
import asyncio
import logging

async def run():
    controller = Controller()
    await controller.connect_current()
    await controller.destroy_model(
                       'model-ceb0e596-dc12-4f44-81c2-f33f0eddd546',
                       'model-24b024e0-b631-4c34-8701-db4d64eb85a4')
                       'model-616bd305-f3df-407f-8414-eca0397c87a1')
...
```

The user can pass as many models as they desire using the format "model-<UUID>"  1,2,3 , etc... 
I'm not catching any exceptions if the model does not exist it will simply move on.  Do we care? 

Let me know if any updates are needed.

### DIFF
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -63,6 +63,20 @@ class Controller(object):
 
         return model
 
+    async def destroy_models(self, *args):
+
+        """Destroy a model to this controller.
+
+        :param str : model-<UUID>
+
+        """
+        model_facade = client.ModelManagerFacade()
+        model_facade.connect(self.connection)
+
+        for arg in args:
+            log.debug('Destroying Model %s', arg)
+            await model_facade.DestroyModels([client.Entity(arg)])
+
     def add_user(self, username, display_name=None, acl=None, models=None):
         """Add a user to this controller.
 
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -67,15 +67,27 @@ class Controller(object):
 
         """Destroy a model to this controller.
 
-        :param str : model-<UUID>
+        :param str : <UUID> of the Model
+        param accepts string of <UUID> only OR `model-<UUID>`
+
 
         """
         model_facade = client.ModelManagerFacade()
         model_facade.connect(self.connection)
 
-        for arg in args:
+        #Generate list of args, pre-pend 'model-'
+        prependarg = list(args)
+        for index, item in enumerate(prependarg):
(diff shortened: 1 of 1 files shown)


juju/controller.py
```diff
@@ -63,6 +63,20 @@ class Controller(object):
...
+
+        :param str : model-<UUID>
+
+        """
+        model_facade = client.ModelManagerFacade()
+        model_facade.connect(self.connection)
+
+        for arg in args:
+            log.debug('Destroying Model %s', arg)
+            await model_facade.DestroyModels([client.Entity(arg)])
```
tvansteenburgh: I would suggest making a list of client.Entity objects from args so that only one call to DestroyModels is needed.

sfeole: Ack, makes sense and I +1 agree..   Will make the update shortly

juju/controller.py
```diff
@@ -63,6 +63,20 @@ class Controller(object):
 
         return model
 
+    async def destroy_models(self, *args):
+
+        """Destroy a model to this controller.
+
+        :param str : model-<UUID>
```
sfeole: ack, simple enough and good idea! will make the update

sfeole:
I just pushed an update. The update was tested with the following script:

```
async def run():
    controller = Controller()
    await controller.connect_current()
    await controller.destroy_models(
                '69238efa-e551-402e-854b-0fe07a605576',
                '2c1a88a0-cad2-4bb8-8300-2a185bd9bf2d',
                '6df5a5ed-04b2-46ad-8425-a7d731e48ffa')
    await controller.disconnect()
```

This provides the following output upon model destruction:

DEBUG:asyncio:Using selector: EpollSelector
INFO:websocket:Driver connected to juju wss://10.0.8.126:17070/api
DEBUG:juju.controller:Destroying Model model-69238efa-e551-402e-854b-0fe07a605576
DEBUG:juju.controller:Destroying Model model-2c1a88a0-cad2-4bb8-8300-2a185bd9bf2d
DEBUG:juju.controller:Destroying Model model-6df5a5ed-04b2-46ad-8425-a7d731e48ffa
DEBUG:juju.controller:Closing controller connection

PR #20, opened 2016-11-24
Add get metrics

simonklb:
This implements support for retrieving metrics stored on the controller for either specific units, all units of an applicaiton or for the whole model.


tvansteenburgh:
@simonklb This is fantastic, thanks!

Issue #10, opened 2016-11-17
Wrong params in auto-generated code


tvansteenburgh:
See ClientFacade.AddMachines() for one example. It accepts arg 'params' of type AddMachineParams, but it should actually accept 'params' of type AddMachines.

Not sure what the proper fix is for this yet. Either the expected type should change, or the auto-generated method body (of AddMachines(), in this example) would need to wrap the passed 'params' in a call to AddMachines.

petevg:
A follow-up on this: I'm still running into trouble, even when I pass in AddMachines as the params.

For example, when I deploy the `hadoop-processing` bundle from the store, I still get the error `juju.errors.JujuAPIError:** json: cannot unmarshal object into Go value of type []params.AddMachineParams` despite applying the fix in the below diff, which I believe is correct (note that I dump the params that I've made .to_json() to make the AddMachines class happy -- it expects to get an array of json objects representing params):

```
diff --git a/juju/model.py b/juju/model.py
index 9d14f82..59dd50b 100644
--- a/juju/model.py
+++ b/juju/model.py
@@ -1255,8 +1255,8 @@ class BundleHandler(object):
         await self.client_facade.AddCharm(None, entity_id)
         return entity_id
 
-    async def addMachines(self, series, constraints, container_type,
-                          parent_id):
+    async def addMachines(self, series=None, constraints=None, container_typNo
ne,                                                                           
+                          parent_id=None):
         """
         :param series string:
             Series holds the optional machine OS series.
@@ -1281,7 +1281,8 @@ class BundleHandler(object):
             container_type=container_type,
             parent_id=self.resolve(parent_id),
         )
-        results = await self.client_facade.AddMachines(params)
+        results = await self.client_facade.AddMachines(
+            client.AddMachines([params.to_json()]))
         log.debug('Added new machine %s', results[0].machine)
         return results[0].machine
```

tvansteenburgh:
Have you tried it without the `.to_json()`? i.e.:

    results = await self.client_facade.AddMachines(client.AddMachines([params]))

To be clear, there is still a bug here (in the facade generation). But afaict your workaround should work in the meantime. But it looks me like you don't need the `.to_json()` in your example.

petevg:
@tvansteenburgh Without the to_json, it raises an exception in the AddMachines class, as that class attempts to unpack params from json.

petevg:
@tvansteenburgh Trace, for posterity:

```
Traceback (most recent call last):
[...]

PR #21, opened 2016-11-28
Fixed addMachines.

petevg:
We now accept a single dict called params, which is what we get back
from the planner, rather than expecting expanded arguments.

We do some parsing of constraints, to put them in a format that should
make juju core happy.

@tvansteenburgh @johnsca: This is a partial fix for deploying hadoop processing. It gets us past all the errors when adding machines. We still get an error unmarshalling the instance.Placement object further down the line, but that error does not appear to stem from this part of the code (I'll work on fixing it next).

### DIFF
--- a/juju/constraints.py
+++ b/juju/constraints.py
@@ -11,6 +11,9 @@
 # constraints keys are valid, and that we can successfully dump the
 # constraints dict to json.
 #
+# Once https://bugs.launchpad.net/juju/+bug/1645402 is addressed, this
+# module should be deprecated.
+#
 
 import re
 
@@ -33,6 +36,10 @@ def parse(constraints):
     """
     if constraints is None:
         return None
+
+    if type(constraints) is dict:
+        # Fowards compatibilty: already parsed
+        return constraints
 
     constraints = {
         normalize_key(k): normalize_value(v) for k, v in [
--- a/juju/constraints.py
+++ b/juju/constraints.py
@@ -1,0 +1,64 @@
+#
+# Library to parse constraints
+#
+# The current version of juju core expects the client to take
+# constraints given in the form "mem=10G foo=bar" and parse them into
+# json that looks like {"mem": 10240, "foo": "bar"}. This module helps us
+# accomplish that task.
+#
+# We do not attempt to duplicate the checking done in
+# client/_client.py:Value here. That class will verify that the
+# constraints keys are valid, and that we can successfully dump the
+# constraints dict to json.
+#
+
+import re
+
+# Matches on a string specifying memory size
+MEM = re.compile('^[1-9][0-9]*[MGTP]$')
+
+# Multiplication factors to get Megabytes
+FACTORS = {
+    "M": 1,
+    "G": 1024,
+    "T": 1024 * 1024,
+    "P": 1024 * 1024 * 1024
+}
+
+def parse(constraints):
+    """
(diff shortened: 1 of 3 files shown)


juju/constraints.py
```diff
@@ -0,0 +1,64 @@
+#
+# Library to parse constraints
```
tvansteenburgh: s/Library/Module

juju/model.py
```diff
@@ -1275,15 +1275,21 @@ def resolve(self, reference):
             case this machine is a container, in which case also ContainerType
             is set.
         """
```
tvansteenburgh: This docstring needs to be updated.

juju/constraints.py
petevg: Done.

petevg:
@tvansteenburgh I agree with your assessment that #10 wasn't a real bug (or rather, it had a real bug initially, where it was accepting a "params" argument, then clobbering it by saying "params=dict()", but @johnsca fixed that earlier).

That said, there is a bug here, which I filed with core: the client shouldn't be responsible for parsing the constraints. constraints.py in this PR is a hopefully short lived workaround, with a future version of the Facade eliminating the need for it.

petevg:
@tvansteenburgh nits addressed. Also added a link to the bug that I filed w/ juju core, along with a notice that constraints.py should eventually go away. What do you think?

juju/model.py
```diff
@@ -1255,35 +1256,38 @@ def resolve(self, reference):
...
+    async def addMachines(self, params):
+        """
+        :param params dict:
+            Dictionary specifying the machine to add. Keys include:
+
+            series: string specifying the machine OS series.
+            constraints: string holding optional machine constraints. We'll
+                parse this into the json friendly dict that the juju api
+                expects.
+            Container_type: string holding the type of the container (for
```
tvansteenburgh: Is this really Container_type and not container_type?

juju/model.py
```diff
@@ -1255,35 +1256,38 @@ def resolve(self, reference):
...
-            container_type=container_type,
-            parent_id=self.resolve(parent_id),
-        )
-        results = await self.client_facade.AddMachines(params)
-        log.debug('Added new machine %s', results[0].machine)
-        return results[0].machine
+    async def addMachines(self, params):
+        """
+        :param params dict:
+            Dictionary specifying the machine to add. Keys include:
```
tvansteenburgh: Are all of the keys in `params` optional? Would be helpful to say which, if any, are required.

petevg: @tvansteenburgh Added a note to the effect that all params are optional (as far as I can tell).

Also made the params dict optional, though I think that you'd only call this method if you had some params to specify.

juju/model.py
```diff
@@ -1255,35 +1256,41 @@ def resolve(self, reference):
...
+        """
+        :param params dict:
+            Dictionary specifying the machine to add. All keys are optional.
+            Keys include:
+
+            series: string specifying the machine OS series.
+            constraints: string holding machine constraints, if any. We'll
+                parse this into the json friendly dict that the juju api
+                expects.
+            Container_type: string holding the type of the container (for
```
tvansteenburgh: Should this be `container_type` instead?

petevg: @tvansteenburgh Yep. That's what I get for copy-pasting without double checking to make sure the original author was right. Fixed. :-)

PR #23, opened 2016-11-29
Added Placement parser.

petevg:
Interim workaround for https://bugs.launchpad.net/juju/+bug/1645480,
where the websocket API will not accept the placement string that the
planner generates.

@tvansteenburgh @johnsca: This PR closes the loop on deploying hadoop-processing, resolving the final error that we get back from the websocket api when we run `await model.deploy('hadoop-processing')`

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -3,6 +3,7 @@ import logging
 
 from . import model
 from .client import client
+from .placement import parse as parse_placement
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


juju/application.py
```diff
@@ -83,7 +84,7 @@ def status_message(self):
 
         result = await app_facade.AddUnits(
             application=self.name,
-            placement=to,
+            placement=[parse_placement(to)],
```
tvansteenburgh: It's not clear to me why the placement param became a list?

petevg: @tvansteenburgh If you take a look at the docstring for AddUnits (line 7575 in _client.py -- it's too big to link), it wants a Sequence for placement. Apparently, like AddMachines, it expects that we might bundle up multiple placement commands in one call.

tvansteenburgh:
Thanks @petevg

PR #29, opened 2016-12-02
Added option to use a non system Juju.

seman:
We don't usually use system Juju for testing. By providing a full path to the Juju binary, it makes it easier to use a non system Juju.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -161,11 +161,11 @@ class Connection:
             "Couldn't authenticate to %s", endpoint)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


tvansteenburgh:
I don't really like this. The library, when necessary, uses whatever juju is on the PATH. If you want to use something other than `/usr/bin/juju`, you can just put it on the PATH before running your program. I think this is a better solution than passing a juju path to the library.

seman:
I thought this will make a little easier than setting up PATH. Feel free to remove PR. I don't have a strong opinion about this.

PR #33, opened 2016-12-13
Improved datetime parsing for unit.py

petevg:
Now uses dateutils lib to handle timezones, nanoseconds, and other
headaches.

@tvansteenburgh @johnsca This adds a dependency. But per IRC discussion, we can't guarantee that timestamps coming from the juju api are UTC, so creating naive datetime objects is probably a bad idea ...

### DIFF
--- a/juju/unit.py
+++ b/juju/unit.py
@@ -1,5 +1,6 @@
 import logging
-from datetime import datetime
+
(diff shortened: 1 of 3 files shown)


petevg:
Bleh. Mistake in the code. Will fix and resubmit.

PR #34, opened 2016-12-14
Improved datetime parsing for unit.py

petevg:
Now uses dateutils lib to handle timezones, nanoseconds, and other
headaches.

@johnsca @tvansteenburgh Fixed my earlier mistake. Resubmitting.

### DIFF
--- a/juju/unit.py
+++ b/juju/unit.py
@@ -1,5 +1,6 @@
 import logging
-from datetime import datetime
+
+from dateutil.parser import parse as parse_date
 
 from . import model
 from .client import client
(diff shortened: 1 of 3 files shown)


petevg:
@tvansteenburgh Forgot to add the justification to this version of the pull request.

Basically, juju core tries to consistently return UTC timestamps, but the timestamps are stored in mongodb as they come back from the machine, so it's possible that you have the case where someone is running a machine that isn't set to UTC, and a non UTC timestamp makes it back to us through the interface.

In light of that, I think that it's best to return timezone aware datetime objects.

It also makes for cleaner code, with less inline chomping of nanoseconds and such.

tvansteenburgh:
Even if the machine is non-utc, doesn't juju normalize timestamps to utc before sending them over the wire? I thought that's what anastasiamac said?

To be clear, I'm okay with this change, I'm just trying to understand why it's needed. We're doing this to workaround bugs that might exist in juju, but we're not sure?

johnsca:
@tvansteenburgh We were told that, in general Juju does convert things to UTC, but that we couldn't have a definite guarantee.  I don't think we've actually hit a case where it wasn't, yet, but since the change to guard against it is cleaner code with only one additional dependency, it seems safer and better all around to go this route.

PR #39, opened 2016-12-22
Fixed detecting controller for non-superuser

johnsca:
The `show-controller` command can only be used by superusers, so usually
fails on shared controllers.

https://bugs.launchpad.net/juju/+bug/1652106


petevg:
+1, based on code readthrough.

PR #43, opened 2017-01-05
Add ability to get cloud for controller

johnsca:
Also use this as default when just a credential is given


petevg:
I am also +1 after Travis.

PR #44, opened 2017-01-06
Add support for local charms

tvansteenburgh:
- Adds support for deploying local charms.
- Adds support for using local charm paths in bundles.


johnsca:
:+1:

PR #46, opened 2017-01-12
Add loop helpers and simplify deploy example

johnsca:


examples/deploy.py
```diff
@@ -2,65 +2,33 @@
...
+
 
-logging.basicConfig(level=logging.DEBUG)
-ws_logger = logging.getLogger('websockets.protocol')
-ws_logger.setLevel(logging.INFO)
-loop = asyncio.get_event_loop()
-loop.set_debug(False)
-loop.create_task(run())
-loop.run_forever()
+loop.run(main)
```
tvansteenburgh: Shouldn't this be `loop.run(main())` ?

johnsca: Yes.  Yes it should.

johnsca:
The interrupt handling doesn't work like I expected and we get a bunch of error dumps.  Also now getting an error on normal run that I don't understand:

```
Connecting to model
Deploying ubuntu
Waiting for active
Removing ubuntu
Disconnecting from model
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

juju/loop.py
```diff
@@ -0,0 +1,31 @@
...
+
+def init_event_loop(interruptable=True, debug=False):
+    """
+    Helper to get the event loop and set some common options.
+    """
+    loop = asyncio.get_event_loop()
+    loop.set_debug(debug)
+    if interruptable:
+        for signame in ('SIGINT', 'SIGTERM'):
+            loop.add_signal_handler(getattr(signal, signame), stop_loop)
```
tvansteenburgh: `stop_loop` could be changed to `loop.stop`

tvansteenburgh:
I don't really understand the "signal handler must be callable" error either. Regarding errors at termination, I assume those are coming from still-running coroutines. You'll always get tracebacks if you stop the loop while any coroutines are still running. You can try canceling all running coroutines in your signal handler.

johnsca:
@tvansteenburgh I tried cancelling the coroutines in various ways but couldn't avoid the errors, which is actually why I had the separate `stop_loop` functionality.  I added one of my attempts, but it doesn't help.

johnsca:
I got it working gracefully and (had to) boiled it down to a single helper function instead of split across three.

tvansteenburgh:
LGTM, thanks @johnsca

PR #47, opened 2017-01-25
Dropped in fixes for juju ssh.

petevg:
Workaround for https://bugs.launchpad.net/juju/+bug/1643076

This requires dropping to the cli client, but it does give us machines
that actually have the juju admin's ssh key.

@tvansteenburgh @johnsca

### DIFF
--- a/examples/add_model.py
+++ b/examples/add_model.py
@@ -1,0 +1,56 @@
+"""
+This example:
+
+1. Creates a model on the current controller
+2. Deploys a charm to it.
+3. Attempts to ssh into the charm
+
+"""
+from juju import loop
+from juju import utils
+from juju.controller import Controller
+import asyncio
+from logging import getLogger
+
+LOG = getLogger(__name__)
+
+
+async def main():
+    controller = Controller()
+    print("Connecting to controller")
+    await controller.connect_current()
+
+    try:
+        model = await controller.add_model("quux")
+
+        print('Deploying ubuntu')
+        application = await model.deploy(
+            'ubuntu-10',
+            application_name='ubuntu',
+            series='trusty',
+            channel='stable',
+        )
+
+        print('Waiting for active')
+        await asyncio.sleep(10)
+        await model.block_until(
+            lambda: all(unit.workload_status == 'active'
+                        for unit in application.units))
+
+        print("Verifying that we can ssh into the created model")
+        ret = utils.execute_process('juju', 'ssh', 'ls /', log=LOG)
+        assert ret
+
+        print('Removing ubuntu')
+        await application.remove()
+
+        print("Destroying model")
+        await controller.destroy_model(model.info.uuid)
+
+    finally:
+        print('Disconnecting from controller')
+        await model.disconnect()
+        await controller.disconnect()
+
+
+loop.run(main())
--- a/examples/add_model.py
+++ b/examples/add_model.py
@@ -21,7 +21,8 @@ async def main():
     await controller.connect_current()
 
     try:
-        model = await controller.add_model("quux")
+        model_name = "quux"
+        model = await controller.add_model(model_name)
 
         print('Deploying ubuntu')
         application = await model.deploy(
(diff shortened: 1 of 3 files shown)


examples/add_model.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+        print('Waiting for active')
+        await asyncio.sleep(10)
+        await model.block_until(
+            lambda: all(unit.workload_status == 'active'
+                        for unit in application.units))
+
+        print("Verifying that we can ssh into the created model")
+        ret = utils.execute_process('juju', 'ssh', 'ls /', log=LOG)
```
tvansteenburgh: This call needs to be awaited.

juju/utils.py
```diff
@@ -0,0 +1,39 @@
...
+def read_ssh_key():
+    '''
+    Attempt to read the local juju admin's public ssh key, so that it
+    can be passed on to a model.
+
+    '''
+    default_data_dir = Path(Path.home(), ".local", "share", "juju")
+    juju_data = os.environ.get("JUJU_DATA", default_data_dir)
+    ssh_key_path = Path(juju_data, 'ssh', 'juju_id_rsa.pub')
+    with ssh_key_path.open('r') as ssh_key_file:
```
tvansteenburgh: It would be better to do file IO in a thread executor so it doesn't block the event loop.

johnsca: Or use [aiofiles](https://pypi.python.org/pypi/aiofiles).  There are a couple of other places we do blocking file IO, but since the files are small, we didn't bother converting them to aiofiles.

tvansteenburgh: @johnsca I knew I could count on you to suggest another dependency!

petevg: @johnsca I was initially going to use aiofiles, but I backed out because I foresaw that @tvansteenburgh would object to it :-p

Since it's a pretty small file, I figured that blocking would be semi-okay. I can refactor to happen in a thread instead, though. That is the best practice :-)

examples/add_model.py
petevg: Ack. Thx for the catch. It would definitely be better if the test actually verified the thing that it was meant to be testing ...

petevg:
@johnsca @tvansteenburgh I dropped in fixes based on your PR comments. Thank you.

In double-checking everything, I ran across a bug. I think that it's an edge case, and I'm not sure how to fix it. Here's how to reproduce:

1) Tweak examples/add_model.py so that it goes back to the original behavior of creating a model with a fixed name, rather than creating a model with a unique id each time.
2) Run the script. Note that it succeeds.
3) Watch juju list-models, and wait until the model is cleaned up (it may lag a little behind the script).
4) Run the script again. Watch it fail :-/

This doesn't happen when you create a model with a new name, or if you manually destroy the model that got created during the failed run, and then re-run the script with that same model name again.

Do either of you have any ideas what might cause this?

juju/utils.py
petevg: Fixed. I call this method in a thread executor context now.

tvansteenburgh: Thanks. 2 things for the record: you can pass None as the first arg to run_in_executor to use the default, which is the threadpoolexecutor. And, I'm fine with using aiohttp and aiofiles in this project. :)

tvansteenburgh:
@petevg Thanks for catching that. I haven't looked into it, but my guess is the cli is blocking until the model is actually destroyed, whereas our client is sending the destroy cmd to the api and then returning immediately. Please open a separate bug for that, if you don't mind.

This actually raises an important question about when we should block and when we shouldn't - a question that applies to lots of stuff in this library. I think in general we should not block by force, but give the user an easy way to block if they want to.

juju/utils.py
petevg: @tvansteenburgh I went ahead with the run_with_executor(None, ...) option (see below -- I also moved the run_in_executor into utils. so that you can naively call read_ssh_key without blocking).

 Maybe next time we'll add the libs :-)

petevg:
@tvansteenburgh I filed https://github.com/juju/python-libjuju/issues/48 to track the "should block" issue. Thx.

tvansteenburgh:
@petevg Thanks for your work on this.

PR #56, opened 2017-02-21
Implement Model.add_machine()

### LINKED ISSUE #51: Implement Model.add_machines()
We have BundleHandler.addMachines() but that's not a public api. Maybe the logic can be relocated to Model.add_machines() and then BundleHandler.addMachines() can call that.

tvansteenburgh:
Also fixes a bug that was causing the 'to' parameter to
Model.deploy() to not be handled correctly.

Add docs and examples for adding machines and containers
and deploying charms to them.

Fixes #51.

### DIFF
--- a/examples/add_machine.py
+++ b/examples/add_machine.py
@@ -13,8 +13,8 @@ import logging
 from juju import loop
 from juju.model import Model
 
-MB = 1024 * 1024
-GB = MB * 1024
+MB = 1
+GB = 1024
 
 
 async def main():
--- a/examples/add_machine.py
+++ b/examples/add_machine.py
@@ -54,9 +54,10 @@ async def main():
                         for unit in application.units))
 
         await application.remove()
-        await machine3.destroy()
-        await machine2.destroy()
-        await machine1.destroy()
+
+        await machine3.destroy(force=True)
+        await machine2.destroy(force=True)
+        await machine1.destroy(force=True)
     finally:
         await model.disconnect()
 
--- a/examples/add_machine.py
+++ b/examples/add_machine.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3.5
+
+"""
+This example:
+
+1. Connects to the current model
+2. Creates two machines and a lxd container
+3. Deploys charm to the lxd container
+
+"""
+import logging
+
+from juju import loop
+from juju.model import Model
+
+MB = 1024 * 1024
+GB = MB * 1024
+
+
+async def main():
+    model = Model()
+    await model.connect_current()
+
+    try:
+        # add a new default machine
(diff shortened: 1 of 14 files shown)


johnsca:
Should we also change `BundleHandler` to use this?

juju/model.py
```diff
@@ -930,9 +966,7 @@ def debug_log(
 
         """
         if to:
-            placement = [
-                client.Placement(**p) for p in to
-            ]
+            placement = parse_placement(to)
```
petevg: This makes a lot of sense. Thank you.

You can technically get rid of the call to parse_placement in application.py (though the "backwards compatibility" line in placement.py means that the multiple calls won't break anything).

examples/add_machine.py
```diff
@@ -0,0 +1,68 @@
...
+3. Deploys charm to the lxd container
+
+"""
+import logging
+
+from juju import loop
+from juju.model import Model
+
+MB = 1024 * 1024
+GB = MB * 1024
```
petevg: If you take a look in constraints.py, these conversions are already built, and they use megabytes, rather than bytes, as a unit, per the Go code on the core side. I'm fairly certain that these numbers will give you machines with much more memory than expected (unless I'm missing a line where you downsize stuff before passing it to the API.)

juju/placement.py
```diff
@@ -41,10 +41,10 @@ def parse(directive):
...
 
     if "/" in directive:
-        machine, container, container_num = directive.split("/")
+        # e.g. "0/lxd/0"
+        # https://github.com/juju/juju/blob/master/instance/placement_test.go#L29
         return [
-            client.Placement(scope=MACHINE_SCOPE, directive=machine),
-            client.Placement(scope=container, directive=container_num)
+            client.Placement(scope=MACHINE_SCOPE, directive=directive),
         ]
```
petevg: Thx for the fix.

juju/model.py
tvansteenburgh: You mean the one in Application.add_unit()? AFAICT that's still needed, what am I missing?

sparkiegeek:
Unit tests 🙏?

examples/add_machine.py
tvansteenburgh: Thanks, fixed.

tvansteenburgh:
@johnsca Re: BundleHandler, I thought about it, but they are different enough that we'd need to factor some stuff out into a 3rd (private) method anyway. Since these are the only two callers, I don't think it's worth it. Maybe later if we end up with more code using the AddMachines api.

tvansteenburgh:
@sparkiegeek Yep, coming. I knew you'd be watching. ;) Thanks for keeping me honest.

PR #63, opened 2017-03-01
Fixes for conjure-up

johnsca:
Includes:
* Implement `Model().get_status`
* Fixes `Model.connect_model()` failing in 2.1 due to model names now including username in `models.yaml`
* Fixes explicit loop instance not being used in several places, causing the default loop to be used for some things but not others

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -34,6 +34,8 @@ class Connection:
         # Connect to the currently active model
         client = await Connection.connect_current()
 
+    Note: Any connection method or constructor can accept an optional `loop`
+    argument to override the default event loop from `asyncio.get_event_loop`.
     """
     def __init__(
             self, endpoint, uuid, username, password, cacert=None,
@@ -265,20 +267,24 @@ class Connection:
         jujudata = JujuData()
 
         if ':' in model:
+            # explicit controller given
             controller_name, model_name = model.split(':')
         else:
+            # use the current controller if one isn't explicitly given
             controller_name = jujudata.current_controller()
             model_name = model
+
+        accounts = jujudata.accounts()[controller_name]
+        username = accounts['user']
+        # model name must include a user prefix, so add it if it doesn't
+        if '/' not in model_name:
+            model_name = '{}/{}'.format(username, model_name)
 
         controller = jujudata.controllers()[controller_name]
         endpoint = controller['api-endpoints'][0]
         cacert = controller.get('ca-cert')
-        accounts = jujudata.accounts()[controller_name]
-        username = accounts['user']
         password = accounts.get('password')
         models = jujudata.models()[controller_name]
-        if '/' not in model_name:
-            model_name = '{}/{}'.format(username, model_name)
         model_uuid = models['models'][model_name]['uuid']
         macaroons = get_macaroons() if not password else None
 
@@ -339,7 +345,9 @@ class JujuData:
         return output.get('current-controller', '')
 
     def current_model(self, controller_name=None):
-        models = self.models()[controller_name or self.current_controller()]
+        if not controller_name:
+            controller_name = self.current_controller()
+        models = self.models()[controller_name]
         if 'current-model' not in models:
             raise JujuError('No current model')
         return models['current-model']
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -260,12 +260,17 @@ class Connection:
     async def connect_model(cls, model, loop=None):
         """Connect to a model by name.
 
-        :param str model: <controller>:<model>
+        :param str model: [<controller>:]<model>
 
         """
-        controller_name, model_name = model.split(':')
-
         jujudata = JujuData()
+
+        if ':' in model:
+            controller_name, model_name = model.split(':')
+        else:
+            controller_name = jujudata.current_controller()
+            model_name = model
+
         controller = jujudata.controllers()[controller_name]
         endpoint = controller['api-endpoints'][0]
         cacert = controller.get('ca-cert')
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -267,6 +267,7 @@ class Connection:
         username = accounts['user']
         password = accounts.get('password')
         models = jujudata.models()[controller_name]
+        model_name = '{}/{}'.format(username, model_name)
         model_uuid = models['models'][model_name]['uuid']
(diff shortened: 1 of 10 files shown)


juju/client/connection.py
```diff
@@ -267,11 +273,13 @@ def https_connection(self):
         username = accounts['user']
         password = accounts.get('password')
         models = jujudata.models()[controller_name]
+        if '/' not in model_name:
+            model_name = '{}/{}'.format(username, model_name)
```
petevg: Will this break in older versions of juju?

johnsca:
@tvansteenburgh It gets passed to the websocket.  Without that, the websocket events get processed by the wrong loop.

I'm currently working on a functional test for this, btw.

juju/client/connection.py
johnsca: It actually looks like this has always been the case in 2.x but we weren't using `Connection.connect_model()` directly before, so this is really more of a convenience that I thought was a bug before.

tvansteenburgh:
@johnsca Ah, thanks I missed that line.

juju/client/connection.py
petevg: Cool. +1 to the change, then.

tvansteenburgh:
Not sure if there are any more commits coming on this, but I'm +1 on what's here so far.

juju/client/connection.py
```diff
@@ -36,13 +37,14 @@ class Connection:
     """
     def __init__(
             self, endpoint, uuid, username, password, cacert=None,
-            macaroons=None):
+            macaroons=None, loop=None):
```
mitechie: Doe sthis need some help text and some usage docs updates? It looks like this changes how this can be fundamentally be used.

juju/client/connection.py
```diff
@@ -248,30 +253,37 @@ def https_connection(self):
...
-        :param str model: <controller>:<model>
+        :param str model: [<controller>:]<model>
 
         """
-        controller_name, model_name = model.split(':')
-
         jujudata = JujuData()
+
+        if ':' in model:
+            controller_name, model_name = model.split(':')
```
mitechie: isn't there a third case where model is passed in but w/o the : ?

juju/client/connection.py
mitechie: this looks like it should be part of determining the model_name logic above. This could include a : ?

juju/client/connection.py
```diff
@@ -326,6 +338,12 @@ def current_controller(self):
         output = yaml.safe_load(output)
         return output.get('current-controller', '')
 
+    def current_model(self, controller_name=None):
+        models = self.models()[controller_name or self.current_controller()]
```
mitechie: I think this would be better as two steps vs combined like this. It's kind of unclear what's going on at first glance.

juju/model.py
```diff
@@ -1005,16 +1005,26 @@ def debug_log(
...
-                ])
+                ], loop=self.loop)
             return [app for name, app in self.applications.items()
                     if name in handler.applications]
         else:
-            log.debug(
-                'Deploying %s', entity_id)
-
             if not is_local:
+                parts = entity_id[3:].split('/')
```
mitechie: we should update is_local to follow juju logic where it has to be a filepath starting with a '.'. 

Then here we'd just do something like auto prefix with cs: and append the rest to the charmstore api?

juju/model.py
```diff
@@ -1005,16 +1005,26 @@ def debug_log(
...
+                    parts.pop(0)
+                if not application_name:
+                    application_name = parts[-1].split('-')[0]
+                if not series:
+                    if len(parts) > 1:
+                        series = parts[0]
+                    else:
+                        entity = await self.charmstore.entity(entity_id)
+                        ss = entity['Meta']['supported-series']
+                        series = ss['SupportedSeries'][0]
```
mitechie: we can let the charmstore do this right? Do we need to deal with the not series and such?

juju/model.py
```diff
@@ -1027,26 +1037,52 @@ def debug_log(
...
+                constraints=constraints,
+                endpoint_bindings=bind,
+                resources=resources,
+                storage=storage,
+                channel=channel,
+                num_units=num_units,
+                placement=parse_placement(to),
+            )
+
+    async def _deploy(self, charm_url, application, series, config,
```
mitechie: are there any comments as to why we now have a _deploy here?

juju/model.py
johnsca: Just verified that a local charm path doesn't have to start with `./` (at least, absolute paths work fine)

mitechie: Awesome, but local: is dead from the cli I believe.

juju/model.py
johnsca: The series param is required by the API, and the charm store doesn't tell us what, if any, series was provided in the URL.  So we have to parse the URL ourselves and if not provided, get the first supported series from the charm store data.  I  will refactor this a little bit, though, because it currently makes two requests when it could make one.  I will also add some more comments.

johnsca:
@tvansteenburgh I should have made multiple PRs but I was lazy and just kept adding to this one.  If everyone is ok with this as is, I'd like to get it merged and I will be better about separate PRs for future changes.

tvansteenburgh:
@johnsca Latest commit LGTM. Feel free to merge if @mitechie is +1.

juju/model.py
```diff
@@ -1005,18 +1008,36 @@ def debug_log(
...
-                ])
+                ], loop=self.loop)
             return [app for name, app in self.applications.items()
                     if name in handler.applications]
         else:
-            log.debug(
-                'Deploying %s', entity_id)
-
             if not is_local:
+                if not application_name:
```
mitechie: There's no test changes for this application parsing stuff that I think would be useful.

johnsca: It should be covered by https://github.com/juju/python-libjuju/pull/63/files#diff-2e7ca22cc563b1b07795f08e3ee7cf42R103 shouldn't it?

mitechie:
It looks like that tests the default expected case but if there's a slash
but no cs:, is there a test that it verifies. Basically there's a bunch of
if's through there but only the one case tested. If it's solidly tested
elsewhere I withdraw my comment but just looking it felt like a lot of
conditionals without a set of tests verifying the combinations.

PR #64, opened 2017-03-01
Add integration tests.

### LINKED ISSUE #27: Integration tests
I'd very much like it if python-libjuju had integration tests with the following properties:

1) They don't run by default, but can be run with "tox integration" or similar. (This is because they may take a long time to run.)
2) They all inherit from a class that creates a new, uniquely named model, runs tests, then tears down that model (this way, the tests won't accidentally clobber your production code.)
3) They sanity check deploys for wiki-simple and a bundle with placement and machine directives. (This would save us from my laziness, where I tested a bundle with placement and machine directives, but didn't test the simpler case, and managed to break addUnits.)

I'll try to set aside and write something like the above -- I wanted to think aloud here in case other people had feedback, or somebody has the spare time right now to write them, though :-)

tvansteenburgh:
- Covers most of what is demonstrated in the examples/ directory.
- The integration tests are not run by default (with the 'tox' command),
  but they are run by TravisCI.
- Each integration test gets its own clean model in which to run, which
  is destroyed at the end of the test. See tests.base.CleanModel.
- Includes charmstore and local bundle tests (local bundle has machine
  and placement directives).

Fixes #27.


petevg:
My machine falls over when running this tests, with an error that suggests that something wacky is happening with my path or site-packages: 

```
Obtaining file:///home/petevg/Code/python-libjuju
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/home/petevg/Code/python-libjuju/setup.py", line 15, in <module>
        from pathlib import Path
    ImportError: No module named pathlib
```

I'm on Xenial, with a minimally polluted system Python. Will keep poking at it, to see if I can figure out what broke.

petevg:
@tvansteenburgh Do you have /usr/bin/python pointed at python3 rather than python2 on your machine, by any chance?

petevg:
Figured it out: https://github.com/juju/python-libjuju/pull/68

PR #66, opened 2017-03-02
Use config-yaml by default when deploying a bundle.

petevg:
This invokes better/more modern behavior for blank strings in the
options for a given charm in the bundle.yaml. Specifically, it allows us
to deploy landscap-dense-maas without running into a KeyError.

This has been tested by using matrix to successfully deploy
landscape-dense-maas, using the code in this PR.

@tvansteenburgh @johnsca @abentley 

Discussion on the justification for this here: https://bugs.launchpad.net/juju/+bug/1651260

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1650,14 +1650,18 @@ class BundleHandler(object):
         """
         # resolve indirect references
         charm = self.resolve(charm)
-        # stringify all config values for API
+        # stringify all config values for API, and convert to YAML
         options = {k: str(v) for k, v in options.items()}
+        options = yaml.dump({application: options}, default_flow_style=False)
         # build param object
         app = client.ApplicationDeploy(
             charm_url=charm,
             series=series,
             application=application,
-            config=options,
+            # Pass options to config-yaml rather than config, as
+            # config-yaml invokes a newer codepath that better handles
+            # empty strings in the options values.
+            config_yaml=options,
             constraints=parse_constraints(constraints),
             storage=storage,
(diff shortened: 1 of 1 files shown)


petevg:
@tvansteenburgh If I can get the integration tests to work on my machine, I will definitely see what I can do about adding a config-yaml check to them. (You merged your PR before I could complain about the errors I'm seeing :-p )

petevg:
@tvansteenburgh ... the tricky thing about testing this in an integration test is that the error in landscape-dense-maas doesn't show up until much later, when a "config changed" hook runs. This means that it doesn't manifest as an error in python-libjuju -- you need to monitor the juju logs and/or unit statuses in order to catch it -- and it makes the tests take a long time.

If I were to write a test to verify that I could deploy landscape-dense-maas without errors, I'd write an integration test that look a lot like an implementation of matrix, complete with health check, to catch the errored state. But we already have matrix :-)

In the meantime, the integration tests that deploy bundles *do* exercise the refactored code path, and they validate that we're using config_yaml correctly -- a poorly formatted string passed to config_yaml creates immediate errors that do show up in python-libjuju, either as JujuAPIErrors, or as a hang.

Given that the integration tests pass, and that a more thorough integration test isn't really practical, and duplicates work done in matrix, are you good with merging this PR?

tvansteenburgh:
Okay, fair enough. I thought it would be simpler, like deploying any random bundle with an empty string option, and then using `app.get_config()` to verify the option was set to empty string rather than not set at all.

PR #74, opened 2017-03-07
Raise errors in a centralized fashion.

petevg:
Fixes error with annotations that's exposed when we actually run tests
with the "raise errors" fix.

Added tests to verify that errors get raised.

~~@tvansteenburgh @johnsca (work in progress because I need to write one more test, though I'd welcome any ideas on what request to use for the final test.)~~


petevg:
@tvansteenburgh No longer a WIP, as @johnsca helped me write the final test.

Issue #73, opened 2017-03-07
Implement Controller.get_model()


tvansteenburgh:
Should return a connected Model instance.

mitechie:
I've got a branch with this started in #75

PR #75, opened 2017-03-07
WIP - Adding list_models and get_model

mitechie:


juju/controller.py
```diff
@@ -206,15 +206,21 @@ def kill(self):
...
 
         """
-        pass
+        model_facade = client.ModelManagerFacade()
+        model_facade.connect(self.connection)
+        username = username or self.connection.info['user-info']['identity']
+
+        results = await model_facade.ListModels(tag.user(username))
+        models = results.user_models
+        return [(m.model.name, m.model.uuid) for m in models]
```
petevg: I think that I have a slight preference for returning `[m.model for m in models]`, with a note that you can get .name, .uuid, and .owner from each model object. That way, if the websocket api ever decides to return more info about a model (e.g., we add some major model relevant feature), we get it for free.

Not a blocker, though, because I can see the argument against, as well as for, this comment :-)

examples/deploy.py
```diff
@@ -18,7 +18,7 @@
     try:
         print('Deploying ubuntu')
         application = await model.deploy(
-            'ubuntu-10',
+            'ubuntu',
```
petevg: I liked having the int in the string, just to demonstrate that you can deploy specific versions of things.

If this was breaking things, though, I understand taking it out ...

juju/controller.py
```diff
@@ -257,13 +263,20 @@ def logout(self, force=False):
...
         """Get a model by name.
 
         :param str name: Model name
 
         """
-        pass
+        model_facade = client.ModelManagerFacade()
+        model_facade.connect(self.connection)
+        results = await model_facade.ModelInfo([client.Entity(tag.model(uuid))])
+        if results and results.results and results.results[0].result:
```
petevg: We really need a better general pattern for digging through results like this. Not a complaint about this particular PR, though :-)

juju/controller.py
```diff
@@ -257,13 +263,20 @@ def logout(self, force=False):
...
+            return results.results[0].result
+        else:
+            log.debug('No result found for model: ', uuid)
+            return None
```
petevg: I'd be tempted to raise an errors.JujuError here. This isn't a "model.exists" call, but a "get me the model!" command, and I think that it is only under Exceptional circumstances that this would come back empty.

Bonus points for writing a NotFoundError, and raising that instead :-)

juju/controller.py
```diff
@@ -257,13 +263,20 @@ def logout(self, force=False):
         """
         pass
 
-    def get_model(self, name):
+    async def get_model(self, uuid):
```
tvansteenburgh: -1 here. I think Controller.get_model() should return a connected Model instance. Would've mentioned this sooner but I thought this PR was still WIP.

mitechie: It is still WIP. I had hoped to work on it during this sprint and wanted to note there was some code in flight in the issue #73 

This isn't ready to land, has no tests, etc.

johnsca:
Is this PR still being worked on?  I think there's some overlap with the just-landed #89

PR #78, opened 2017-03-08
Don't raise ConnectionClosed error in receiver task

johnsca:
It pointlessly fills the log with unretrieved exception messages.


petevg:
+1. Will merge after tests complete.

PR #82, opened 2017-03-09
Updated examples to use juju.loop

petevg:
Also updated the README.

Made a few other tweaks to make them more consistent in style.

@tvansteenburgh @abentley @johnsca: I wanted to make sure that when people encounter this library the first time, they encounter our loop helper, and aren't turned off by the details of standing up an asyncio event loop.

### DIFF
--- a/examples/action.py
+++ b/examples/action.py
@@ -10,6 +10,7 @@ This example:
 import asyncio
 import logging
 
+from juju import loop
 from juju.model import Model
 
 
@@ -24,7 +25,7 @@ async def run_action(unit):
     logging.debug("Action results: %s", action.results)
 
 
-async def run():
+async def main():
     model = Model()
     await model.connect_current()
     await model.reset(force=True)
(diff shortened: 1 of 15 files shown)


examples/action.py
```diff
@@ -40,13 +41,10 @@
...
-ws_logger = logging.getLogger('websockets.protocol')
-ws_logger.setLevel(logging.INFO)
-loop = asyncio.get_event_loop()
-loop.set_debug(False)
-loop.create_task(run())
-loop.run_forever()
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG)
+    ws_logger = logging.getLogger('websockets.protocol')
+    ws_logger.setLevel(logging.INFO)
```
johnsca: Perhaps we should go ahead and make a helper for this common logging config.

examples/fullstatus.py
```diff
@@ -2,14 +2,11 @@
...
 from juju.client.client import ClientFacade
-
-
-loop = asyncio.get_event_loop()
-conn = loop.run_until_complete(Connection.connect_current())
-
+from juju import loop
 
 async def status():
     client = ClientFacade()
```
johnsca: I didn't realize we had an example exposing the raw API facades.   This should probably be added as a method on the object layer.

tvansteenburgh: Yeah, this has been implemented in Model.get_status() since this example was created. But I don't think it hurts to have some examples of calling APIs directly too. Although this might be a bad example to use for that now since there's a way to do it in the object layer.

examples/action.py
petevg: @johnsca I actually had the helper half written. The problem is that the values we set in the examples aren't really values that you'd set for production. I wound up backing it out, and thinking maybe I'd add it in another pass.

Issue #90, opened 2017-03-17
python-libjuju is broken against the latest alpha


petevg:
If you call the FullStatus facade, python-libjuu breaks due to a missing model-status key. It looks like we have a breaking change in the latest version of the websocket api.

Pushing a test that reproduces this shortly.

@tvansteenburgh @johnsca

petevg:
Repro in this PR: https://github.com/juju/python-libjuju/pull/91

petevg:
This is fixed!

Issue #49, opened 2017-01-30
Figure out how to handle version changes


tvansteenburgh:
See https://lists.ubuntu.com/archives/juju/2017-January/008479.html for context.

Loading the correct facade version dynamically is the way to go. The schema json file will need to be appended to instead of overwritten so that old facade versions are preserved. API classes will need to be namespaced so that two versions of the same facade (with the same name) don't collide.

That's the easy part. The tricky part is the coupling between the OO and api layers. For example, Application.add_units() calls ApplicationFacade.AddUnits() with parameters. What happens when the next version of that facade changes the parameters? The OO layer needs some way to deal with that...

petevg:
I'm tempted to do some munging at import time. Set an environment variable (or some other flag) before importing juju, and then have client.py import _client<version>.py, depending on the flag.

Another option would be to have a set of wrappers (one per version) sitting in front of the object layer. Depending on which wrapper you imported your objects from, we'd do something different with the _client import. That would break if people tried to dig deeper into the library for their imports, though.

petevg:
@tvansteenburgh @johnsca @abentley I have a proposal for a solution, with some psuedo code to illustrate how it might work.

It lives in a pastebin: http://paste.ubuntu.com/24320954/

What do you think. Is it a workable solution?

johnsca:
@petevg I'm not a big fan of having the implementation be dynamically loaded at run-time.  It makes it hard to know what params to pass to the facades, whether by introspection, reading the source, or reading generated docs.  Also, if / when https://bugs.launchpad.net/juju/+bug/1679438 happens, we should be using that schema definition directly and shouldn't even need schemagen or schema JSON files in the libjuju repo anymore.

Instead, I'd prefer to split `_client.py` into separate packages for each facade with a file for each version of the facade but still have them pre-generated.  We could even have an optimization that only looks for new versions of facades that don't have implementations and don't bother regenerating existing versions.  I think this gives us a natural place to put generated docstrings, makes the params very explicit, and should give us nice commits that make it very clear what's new.  It would also make it easy to diff two versions of a facade to see what changed.  I think the import logic would work pretty much how you described for loading the versioned schema in the pastebin.

Does that seem like a reasonable counter-proposal?

johnsca:
[...]

PR #91, opened 2017-03-17
Added test to reproduce bug with FullStatus against latest juju.

petevg:
If the current python-libjuju is run against juju-2.2-alpha1 (or the
daily builds), client.FullStatus breaks due to a missing 'model-status'
key.

This is a repro for https://github.com/juju/python-libjuju/issues/90

@johnsca @tvansteenburgh

### DIFF
--- a/tests/integration/test_connection.py
+++ b/tests/integration/test_connection.py
@@ -1,6 +1,7 @@
 import pytest
 
 from juju.client.connection import Connection
+from juju.client import client
 from .. import base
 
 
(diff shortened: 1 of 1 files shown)


johnsca:
The test is good, and is catching the issue, but I'm unsure how to proceed with this PR since I don't want the build to be broken.  AFAIU, we don't have any reasonable way to fix this without addressing #49.

petevg:
@johnsca Yeah. I don't want us to merge this.

Having an open PR kind of sucks, though, and we understand the issue pretty thoroughly. I'm going to lose this out for now -- I can resubmit once we have the bug fixed.

PR #92, opened 2017-03-17
Regen schema to pick up missing APIs

johnsca:
Due to some portions of the API living in a different location in the Juju source, they were missed.  The schema gen code has been factored out into its own repo (https://github.com/juju/schemagen) and the
missing APIs fixed.  Some minor fixes to the code gen were also required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 BIN := .tox/py35/bin
 PY := $(BIN)/python3.5
 PIP := $(BIN)/pip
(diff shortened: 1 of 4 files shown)


petevg:
This looks good, and works for me locally.

Since this contains an updated schema, generated based on the latest juju alpha, I wanted to test for backwards compatibility with the current stable release of juju 2 before merging.

Doing that now ...

PR #102, opened 2017-04-03
Include resources from store when deploying

### LINKED ISSUE #77: Deploy needs to include resources
See https://github.com/conjure-up/conjure-up/blob/master/conjureup/juju.py#L458

johnsca:
Fixes #77

### DIFF
--- a/juju/client/overrides.py
+++ b/juju/client/overrides.py
@@ -1,9 +1,14 @@
 from collections import namedtuple
 
-from .facade import Type
+from .facade import ReturnMapping, Type
+from .import _client
 
 __all__ = [
     'Delta',
+]
+
+__patches__ = [
+    'ResourcesFacade',
 ]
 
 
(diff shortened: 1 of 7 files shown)


juju/client/overrides.py
```diff
@@ -43,3 +48,35 @@ def __init__(self, deltas=None):
...
     def from_json(cls, data):
         return cls(deltas=data)
+
+
+class ResourcesFacade(Type):
+    """Patch parts of ResourcesFacade to make it work.
+    """
+
+    @ReturnMapping(_client.AddPendingResourcesResult)
+    async def AddPendingResources(self, application_tag, charm_url, resources):
```
battlemidget: We should start with this but open a bug to the juju team to have these put under the right api code so the generator will pick these up

johnsca: I had discussed this with @mjs so there might be an issue already opened.  I'm following up to confirm or get one opened if not.

PR #103, opened 2017-04-03
Support placement lists

johnsca:


petevg:
+1 for recursion :-) LGTM after tests pass.

Issue #99, opened 2017-03-29
The websocket connection can fail silently


petevg:
I don't have a clean repro yet, but I believe this to be the underlying cause of this bug in matrix: https://github.com/juju-solutions/matrix/issues/92

Basically, matrix periodically checks on the "health" of a cluster by checking the status for each unit in each application in Model.applications, where Model is the Model class from python-libjuju's model.py. That data is supposed to be updated by a listener inside the Model object.

If matrix is running on a flaky connection, or if the Controller times the websocket connection out, the listener will stop receiving updates, however, with no Exceptions being thrown. I've verified that this is what's happening by running matrix in a debugger, and checking the .is_open property of the connection. Eventually, it gets set to false because we have been disconnected from the Controller.

We need to either come up with an automagic reconnection scheme, or a recommended pattern for dealing with the case where the listener gets disconnected. 

Note that this is semi-related to https://github.com/juju/python-libjuju/issues/98, though making a Pinger would only address Controller initiated timeouts; a flaky connection would still kill the listener.

petevg:
I've been giving this one a lot of thought over the past couple of days.

I don't think that an auto-resume will work. There doesn't seem to be a facility in the websocket api for resuming a watcher, and there are a lot of ways to break and miss messages if we try to hack something together on our end.

I also don't think that an elaborate Exception surfacing scheme is necessarily called for. At least, I haven't come up with one that isn't a headache for developers who aren't me to understand.

I think that the best solution would be to attach a "monitor" class to each connection object. The monitor class would have a "status" property that would be in one of three states:

```
    1. connected
    2. errored
     3. closed
```

"connected" means that everything is healthy. "closed" means that the websocket connection is closed due to the connection.close() method being called. "errored" means that the receiver method raised an Exception for whatever reason, and our connection is no longer functional (and probably disconnected due to an unexpected network issue).

The monitor could be used in the following ways:

1) The "health" check in matrix can see that `connection.monitor.status == 'errored'`, and raise an InfraFailure.
2) A tool with an interactive user interface could have a little "connection" icon that goes from green to red when the connection fails, prompting the user to take action to reconnect.
3) A tool could track all its watchers, possibly via helpers in python-libjuju, and be able to restart them if the connection.monitor.status goes bad. The tool would be responsible for ensuring that any actions it takes on deltas in a watcher are idempotent. An auto-reconnect helper in the matrix tasks could be an eventual model for this pattern.

[...]

PR #105, opened 2017-04-04
Added Monitor class to Connection.

petevg:
Calling Monitor.status allows a client to determine whether the
connection is "connected", "disconnected" or in an "error" state, and
take appropriate action as a result.

This is meant to address the issue of watchers failing silently when the
websocket connection goes away due to a network issue.

@tvansteenburgh @johnsca @abentley Addresses https://github.com/juju/python-libjuju/issues/99. WIP because I want to write some matrix code that uses this and validate that it actually solves my problem. Edit: no longer WIP, as this works quite nicely in matrix :-)

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -21,6 +21,83 @@ from juju.utils import IdQueue
 log = logging.getLogger("websocket")
 
 
+class Monitor:
+    """
+    Monitor helper class for our Connection class.
+
+    Contains a reference to an instantiated Connection, along with a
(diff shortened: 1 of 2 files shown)


petevg:
@johnsca if you drill down the AllWatcher task ends up repeatedly calling connection.recv (via Type.rpc, via AllModelWatcherFacade.Next), which, thanks to a change you made, raises an Exception if the connection unexpectedly closes.

So the future should make it into the "done" state, with an .exception on it. I'm going to go ahead and merge this :-)

PR #107, opened 2017-04-14
[WIP] Feature/api version support

petevg:
WIP because the code needs a lot of cleanup, and it still isn't working!

@tvansteenburgh @johnsca I would appreciate any insight that you two have on a blocking bug: after my refactor, we no longer seem to be writing out arguments for the methods in the Facade classes.

For example, the ActionFacade has a method called Actions. The old code wrote this out:

```
    @ReturnMapping(ActionResults)
    async def Actions(self, entities):
        '''
        entities : typing.Sequence<+T_co>[~Entity]<~Entity>
        Returns -> typing.Sequence<+T_co>[~ActionResult]<~ActionResult>
        '''
        # map input types to rpc msg
        _params = dict()
        msg = dict(type='Action', request='Actions', version=2, params=_params)
        _params['entities'] = entities
        reply = await self.rpc(msg)
        return reply
```

... and my refactor writes this:

```
 @ReturnMapping(ActionResults)
    async def Actions(self):
        '''

        Returns -> typing.Sequence<+T_co>[~ActionResult]<~ActionResult>
        '''
        # map input types to rpc msg
        _params = dict()
        msg = dict(type='Action', request='Actions', version=2, params=_params)

        reply = await self.rpc(msg)
        return reply
```

(Note the missing "entities" argument, along the the missing line in the docstring about it, and the missing call to set it up in params['entitites'].)

The weird thing is that I'm not sure how this was working in the first place. makeFunc is the thing that builds those methods. It is called by _buildMethod, and it has a "params" argument that it uses to construct the args for each method. But buildMethod sets params to "None", and then doesn't do anything with them! There's some magic, maybe with the _types registry, that happens in the original code, and I appear to have broken it, but I'm not sure _how_ I broke it :-/


Makefile
```diff
@@ -15,8 +15,7 @@ client:
 ifndef SCHEMAGEN
 	$(error "schemagen is not available, please install from https://github.com/juju/schemagen")
 endif
-	schemagen > juju/client/schemas.json
-	$(PY) -m juju.client.facade -s juju/client/schemas.json -o juju/client/_client.py
+	$(PY) -m juju.client.facade -s "juju/client/schemas*" -o juju/client/_client
```
petevg: Dropped the call to automatically invoke schemagen, because we're writing out files based on a pattern, rather than writing schemas.json each time.

petevg:
@tvansteenburgh @johnsca Following up on this, the same thing (missing subroutine arguments) happens when I try to checkout master, and run `python3 -m juju.client.facade -s juju/client/schemas.json -o juju/client/_client.py`

That's ... weird. How confident are we that our _client.py building code was actually working correctly? (I'm gonna dig through git logs now ...)

petevg:
Aha! I found it. This commit, from April 6th, blows up the params arg that makeFunc is expecting:

```
diff --git a/juju/client/facade.py b/juju/client/facade.py
index a7083e6..a6fd9f6 100644
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -401,7 +401,11 @@ def _buildMethod(cls, name):
         prop = method['properties']
         spec = prop.get('Params')
         if spec:
-            params = _types.get(spec['$ref'])
+            result = _types.get(spec['$ref'])
+            if '$ref' in spec:
+                result = _types.get(spec['$ref'])
+            else:
+                result = SCHEMA_TO_PYTHON[spec['type']]
         spec = prop.get('Result')
         if spec:
             if '$ref' in spec:

```

 @mhmoerma can you give me some insight into what you were attempting to do with that change? I'd like to unbreak the _client generate code, but I don't want to break other bits in the process :-p

petevg:
@mhmoerma I think that the change I referenced above was just the result of a merge conflict. I'm going to assume that's the case and just revert it, unless I hear differently from you. Sorry to bug ya :-)

petevg:
I'm currently blocked by https://bugs.launchpad.net/juju/+bug/1682925, so I'm going to go ahead and close out this PR for now. Once I've worked around the above, I'll either be in a place where I need some feedback on working around the next blocker, or have a non WIP PR to review :-)

mhmoerma:
Sorry for the late reply. @petevg Yes, this was just a merge conflict.

PR #108, opened 2017-04-17
[WIP] Feature/api version support

petevg:
Still working on smoothing out the rough edges and remaining test failures on this. Getting close, though ...


petevg:
So the thing that is broken about this build is that I seem to have made an error when generating the schemas -- they all have an AllWatcher version of 1, which is not correct.

Will re-generate schemas, and try this again ...

PR #113, opened 2017-04-26
Workaround or missing redirect_info in JaaS.

petevg:
JaaS deployments were throwing a JujuAPIError because the JaaS
controller does not appear to have implemented redirect_info.

Dropped in a change to catch the error and ignore it in that case.

@tvansteenburgh @johnsca I'm not sure how to test JaaS stuff effectively. To validate that my code works, you can run examples/add_model.py, after editing the file to pass `cloud_name="aws"` into the call to .add_model.

(Without this change, add_model will fail due to a JujuAPIErrror.)

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -323,7 +323,16 @@ class Connection:
                      loop)
         await client.open()
 
-        redirect_info = await client.redirect_info()
+        try:
+            redirect_info = await client.redirect_info()
+        except JujuAPIError as e:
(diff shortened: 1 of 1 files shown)


johnsca:
I created a more comprehensive refactor in #116 which I'd like to close this PR in favor of if it looks ok to you.

petevg:
Agree that @johnsca's refactor is more comprehensive. Closing this out ...

PR #121, opened 2017-05-08
Strip local:-prefix from local charm urls

simonklb:
Minor fix for local charms when using local: prefix. The prefix was never removed which corrupted the path.

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1000,7 +1000,7 @@ class Model(object):
             os.path.isdir(entity_url)
         )
         if is_local:
-            entity_id = entity_url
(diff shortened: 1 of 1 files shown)


johnsca:
I don't think the `local:` prefix is technically allowed by the CLI any more, but there was an issue with the bundle API returning something that included the prefix for internal use which I thought was required.  The bundle API originally called this user-facing method directly, before it was refactored to use the `_deploy` internal method instead.  I think this handling of that prefix was left over from before that refactor.

Regardless, I think this change is fine, but we should really get some test cases added to the integration suite to vet local charm handling.

PR #128, opened 2017-05-12
Add model context manager

simonklb:
This implements a context manager for the Model class which automatically connects to the current model and disconnects when you exit the scope. I found myself repeating the connect and disconnect functions a lot in charm tests so I thought this could be beneficial to someone else as well.

Example usage:
```
with Model() as model:
    model.deploy(...)
```

I also added the asynctest package to the testsuite because pytest-asyncio doesn't support TestCase classes.
See: https://github.com/pytest-dev/pytest-asyncio/issues/15
asynctest made it extremely easy to write tests for async functions, it could be useful to check out for other tests where asyncio is heavily involved.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -407,7 +407,11 @@ class Connection:
 
         """
         jujudata = JujuData()
+
         controller_name = jujudata.current_controller()
+        if not controller_name:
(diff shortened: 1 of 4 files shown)


tvansteenburgh:
LGTM

PR #131, opened 2017-05-31
Accept new / unknown API fields gracefully

### LINKED ISSUE #130: Definitions should gracefully accept unknown params
From discussions in Montreal, it seems that new fields in the Juju schema definitions are expected to be handled gracefully even without a bump to the facade version.  That is, facade versions will only be bumped if a field is removed or changed.

johnsca:
From discussions in Montreal, it seems that new fields in the Juju schema definitions are expected to be handled gracefully even without a bump to the facade version. That is, facade versions will only be bumped if a field is removed or changed.

Fixes #130
Also fixes conjure-up/conjure-up#903


docs/contributing/index.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Once you have a new schema file, you can update the Python code
+using the `client` make target:
+
+.. code:: bash
+
+  make client
+
+You should expect to see updates to the `juju/client/_definitions.py` file,
+as well as one or more of the `juju/client/_clientX.py` files, dependning on
```
tvansteenburgh: depending

docs/contributing/index.rst
```diff
@@ -0,0 +1,66 @@
...
+which facades were touched.
+
+
+Overrides
+---------
+
+It should be quite rare, but occasionally the generated Python code is does
```
tvansteenburgh: s/is//

docs/contributing/index.rst
```diff
@@ -0,0 +1,66 @@
...
+not capture all of the logic needed properly parse the output from the API
```
tvansteenburgh: needed to properly

docs/index.rst
```diff
@@ -16,6 +16,7 @@ Table of Contents
 
    narrative/index
    API Docs <api/modules>
+   schema/index
```
tvansteenburgh: Should this be contributing/index?

juju/client/overrides.py
```diff
@@ -96,5 +101,94 @@ class AllWatcherFacade(Type):
...
+                raise TypeError('Unable to parse Binary version string: '
+                                '{}'.format(data))
+        d = {}
+        for k, v in (data or {}).items():
+            d[cls._toPy.get(k, k)] = v
+
+        return cls(**d)
+
+    def serialize(self):
+        return "%s-%s-%s" % (self.number.serialize(), self.series, self.arch)
```
tvansteenburgh: Number and Binary should at least have a happy path test case for parse/serialize.

docs/contributing/index.rst
battlemidget: i wasn't going to nitpick but also `Python code is does` should be fixed

battlemidget:
I just tested a new built snap pulling in this branch for conjure-up and deployments succeed again.

johnsca:
For reference, the tests were ported from https://github.com/juju/version/blob/master/version_test.go

PR #132, opened 2017-06-01
Implement Application.upgrade_charm method

Cynerva:
Partial implementation of `Application.upgrade_charm` method.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -346,6 +346,8 @@ class Application(model.ModelEntity):
 
         if switch is not None:
             charm_url = switch
+            if not charm_url.startswith('cs:'):
+                charm_url = 'cs:' + charm_url
         else:
             charm_url = self.data['charm-url']
             charm_url = charm_url.rpartition('-')[0]
--- a/juju/application.py
+++ b/juju/application.py
@@ -372,8 +372,9 @@ class Application(model.ModelEntity):
             storage_constraints=None
         )
 
-        while self.data['charm-url'] != charm_url:
-            await asyncio.sleep(1)
+        await self.model.block_until(
+            lambda: self.data['charm-url'] == charm_url
+        )
 
     async def get_metrics(self):
         """Get metrics for this application's units.
--- a/juju/application.py
+++ b/juju/application.py
@@ -3,6 +3,7 @@ import logging
 
 from . import model
 from .client import client
+from .errors import JujuError
 from .placement import parse as parse_placement
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


juju/application.py
```diff
@@ -331,7 +331,49 @@ def upgrade_charm(
...
+            charm_url = self.data['charm-url']
+            charm_url = charm_url.rpartition('-')[0]
+            if revision is not None:
+                charm_url = "%s-%d" % (charm_url, revision)
+            else:
+                charmstore = self.model.charmstore
+                entity = await charmstore.entity(charm_url, channel=channel)
+                charm_url = entity['Id']
+
+        await client_facade.AddCharm(
```
tvansteenburgh: If  charm_url turns out to be the same as what's already deployed (i.e. there's not a newer revision), will AddCharm() fail? We should probably just return in that case, right?

juju/application.py
```diff
@@ -331,7 +331,49 @@ def upgrade_charm(
...
+            charm_url=charm_url,
+            config_settings=None,
+            config_settings_yaml=None,
+            force_series=force_series,
+            force_units=force_units,
+            resource_ids=None,
+            storage_constraints=None
+        )
+
+        while self.data['charm-url'] != charm_url:
```
tvansteenburgh: Consider using:

```python
await self.model.block_until(
    lambda: self.data['charm-url'] == charm_url
)
```

Minor difference, but the latter will always use the correct event loop in a multi-threaded scenario.

juju/application.py
Cynerva: I just tested this scenario, and the result was that upgrade_charm completed without error.

The Juju CLI, however, raises an error in the same scenario:
```
$ juju deploy cs:~containers/easyrsa
Located charm "cs:~containers/easyrsa-9".
Deploying charm "cs:~containers/easyrsa-9".

$ juju upgrade-charm easyrsa --revision 9
ERROR already running specified charm "cs:~containers/easyrsa-9"
```

I'll add a check here and raise an exception accordingly. Thanks for the feedback!

Cynerva:
Pushed changes to address PR feedback. Also fixed a bug uncovered by the integration tests (whew!).

Thanks again for looking. Anything else?

tvansteenburgh:
LGTM once tests pass - thanks!

johnsca:
If it hangs forever without the cs: prefix, is it possible for other invalid URLs to cause it to hang?

Cynerva:
> If it hangs forever without the cs: prefix, is it possible for other invalid URLs to cause it to hang?

Possibly, yeah. It'd have to be one that the Juju API accepts, but then normalizes to something else. I can't think of any URLs like this off the top of my head, but they're probably lurking about somewhere. I'll look into it.

PR #133, opened 2017-06-02
Add machine status properties

johnsca:
Add properties to access the machine provisioning and Juju machine agent status information.

Unfortunately, there seems to be a bug with the AllWatcher related to these fields (https://bugs.launchpad.net/juju/+bug/1695335) that causes the deltas to contain stale data, so this also includes a hack to work around that by fetching newer data from the FullStatus API.

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -1,12 +1,86 @@
 import logging
 
-from . import model
+from dateutil.parser import parse as parse_date
+
+from . import model, utils
 from .client import client
 
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


tvansteenburgh:
Looks like the test timeout needs to be extended, but other than that this LGTM, given that it's temporary until the core bug is fixed.

Cynerva:
I like that you have the workaround clearly called out and isolated for easy removal. Thanks for that!

battlemidget:
This also addresses the first part of this bug: https://github.com/conjure-up/conjure-up/issues/856

PR #144, opened 2017-06-14
Fix deploying non-stable channels and explicit revs

### LINKED ISSUE #143: Handling of channels and explicit revs is broken
If you call `model.deploy()` with an explicit, non-stable channel, the stable channel still gets deployed.  If you try to use an explicit rev without specifying the channel, it fails with a "charm not found" error.

johnsca:
Fixes #143

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1002,7 +1002,7 @@ class Model(object):
         :param dict bind: <charm endpoint>:<network space> pairs
         :param dict budget: <budget name>:<limit> pairs
         :param str channel: Charm store channel from which to retrieve
-            the charm or bundle, e.g. 'development'
+            the charm or bundle, e.g. 'edge'
         :param dict config: Charm configuration dictionary
         :param constraints: Service constraints
         :type constraints: :class:`juju.Constraints`
--- a/juju/model.py
+++ b/juju/model.py
@@ -1040,7 +1040,7 @@ class Model(object):
         if is_local:
             entity_id = entity_url.replace('local:', '')
         else:
-            entity = await self.charmstore.entity(entity_url)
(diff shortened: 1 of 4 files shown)


juju/model.py
```diff
@@ -1072,8 +1072,6 @@ def _get_series(self, entity_url, entity):
                     application_name = entity['Meta']['charm-metadata']['Name']
                 if not series:
                     series = self._get_series(entity_url, entity)
-                if not channel:
-                    channel = 'stable'
```
tvansteenburgh: Why was this removed?

johnsca: Because it's wrong.  ;)  If an explicit rev is given, it may not be the stable channel.  There's also no actual need to provide a value for the channel, as we already resolve the charm to an explicit rev by querying the charm store.

tests/integration/test_model.py
```diff
@@ -36,6 +36,22 @@
 
 @base.bootstrapped
 @pytest.mark.asyncio
+async def test_deploy_channels_revs(event_loop):
```
wwwtyro: :heart:

juju/model.py
tvansteenburgh: Okay, makes sense, I just wanted to make sure the subsequent call to AddCharm wouldn't fail when passed a None channel.

johnsca: Sure.  The test does cover that case.

kwmonroe: Sorry to hijack this @johnsca, but while you're in here, would you please fix the docstring on line 1005 s/development/edge?

johnsca: Sure.  Done.

PR #148, opened 2017-06-20
Improve handling of closed connections

### LINKED ISSUE #147: Watcher dies with uncatchable ConnectionClosed error
Originally reported via [conjure-up's Sentry](https://sentry.io/canonical-pj/conjure-up/issues/296142667/):

```
ConnectionClosed: WebSocket connection is closed: code = 1006, no reason.
  File "juju/model.py", line 648, in _start_watch
    self.loop)
  File "juju/utils.py", line 94, in run_with_interrupt
    result = [f.result() for f in done if f is not event_task]
  File "juju/utils.py", line 94, in <listcomp>
    result = [f.result() for f in done if f is not event_task]
  File "juju/client/facade.py", line 392, in wrapper
  File "juju/client/_client1.py", line 59, in Next
    reply = await self.rpc(msg)
  File "juju/client/overrides.py", line 104, in rpc
  File "juju/client/connection.py", line 238, in rpc
    result = await self.recv(msg['request-id'])
  File "juju/client/connection.py", line 175, in recv
    return await self.messages.get(request_id)
  File "juju/utils.py", line 63, in get
    raise value
  File "juju/client/connection.py", line 183, in receiver
    loop=self.loop)
  File "juju/utils.py", line 94, in run_with_interrupt
    result = [f.result() for f in done if f is not event_task]
  File "juju/utils.py", line 94, in <listcomp>
    result = [f.result() for f in done if f is not event_task]
  File "websockets/protocol.py", line 299, in recv
    raise ConnectionClosed(self.close_code, self.close_reason)
```

This should be reported to the Monitor.  However, we probably also need to have `block_until` surface closed connections instead of blocking indefinitely.

johnsca:
This fixes #147 by closing the watcher gracefully when the connection is closed unexpectedly so that the `Monitor` can report it as intended.  It also fixes a circular reference between the `Connection` and `Monitor`, improves the logic in the monitor status, and makes `Model.block_until` raise a `websockets.ConnectionClosed` exception instead of blocking indefinitely.

**Edit:** Also fixes conjure-up/conjure-up#965

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -656,6 +656,18 @@ class Model(object):
                             allwatcher.Next(),
                             self._watch_stopping,
                             self.loop)
+                    except JujuAPIError as e:
+                        if 'watcher was stopped' not in str(e):
+                            raise
+                        # controller stopped our watcher for some reason
+                        # (we never actually call AllWatcherFacade.Stop(),
+                        # so there's no reason we should see this, but the
+                        # controller does occasionally send it anyway; so
+                        # if we get it, just start a new watcher)
+                        log.warning(
+                            'Watcher: watcher stopped, restarting')
+                        del allwatcher.Id
+                        continue
                     except websockets.ConnectionClosed:
                         monitor = self.connection.monitor
                         if monitor.status == monitor.ERROR:
--- a/juju/model.py
+++ b/juju/model.py
@@ -659,11 +659,14 @@ class Model(object):
                     except JujuAPIError as e:
                         if 'watcher was stopped' not in str(e):
                             raise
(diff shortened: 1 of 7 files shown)


juju/model.py
```diff
@@ -659,11 +659,14 @@ def _watch(self):
                     except JujuAPIError as e:
                         if 'watcher was stopped' not in str(e):
                             raise
+                        if self._watch_stopping.is_set():
+                            # this shouldn't ever actually happen, because
+                            # the event should trigger before the controller
+                            # has a chancel to tell us the watcher is stopped
```
battlemidget: chancel -> chance

PR #149, opened 2017-06-22
Implement scp functionality

Cynerva:
This adds scp methods to the Unit and Machine classes. I had to redefine the interfaces a little bit since the previous scp stubs didn't account for push vs. pull.

Also added a couple properties along the way, `unit.machine` and `machine.public_address`.

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -140,7 +140,7 @@ class Machine(model.ModelEntity):
         if proxy:
             raise NotImplementedError('proxy option is not implemented')
 
-        address = self.public_address
+        address = self.dns_name
         destination = '%s@%s:%s' % (user, address, destination)
         await self._scp(source, destination, scp_opts)
 
@@ -157,7 +157,7 @@ class Machine(model.ModelEntity):
         if proxy:
             raise NotImplementedError('proxy option is not implemented')
 
-        address = self.public_address
+        address = self.dns_name
         source = '%s@%s:%s' % (user, address, source)
         await self._scp(source, destination, scp_opts)
 
@@ -247,10 +247,11 @@ class Machine(model.ModelEntity):
         return parse_date(self.safe_data['instance-status']['since'])
 
     @property
-    def public_address(self):
-        """Get a public address for this machine.
+    def dns_name(self):
+        """Get the DNS name for this machine. This is a best guess based on the
+        addresses available in current data.
 
-        May return None if no public address is found.
+        May return None if no suitable address is found.
         """
         for scope in ['public', 'local-cloud']:
             addresses = self.safe_data['addresses'] or []
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -124,18 +124,27 @@ class Machine(model.ModelEntity):
         )
         return await self.ann_facade.Set([ann])
 
-    def scp(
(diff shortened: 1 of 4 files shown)


johnsca:
The failure from Travis is:

```
tests/integration/test_machine.py::test_scp Warning: Identity file ~/.local/share/juju/ssh/juju_id_rsa not accessible: No such file or directory.
ssh: Could not resolve hostname none: Name or service not known
```

Further down is the command executed:

```
E           juju.errors.JujuError: command failed: ['scp', '-i', '~/.local/share/juju/ssh/juju_id_rsa', '-o', 'StrictHostKeyChecking=no', '/tmp/tmpf5khpgto', 'ubuntu@None:testfile']
juju/machine.py:178: JujuError
```

There also seem to be some subsequent SSL-related failures on the websocket, but I suspect those are spurious.

johnsca:
I'm not clear on what the need is for separate `scp_to` vs `scp_from` commands?  The Juju CLI is able to handle both with a single command, as is the underlying `scp` command, so I'd much prefer to preserve that.  It just means being able to detect unit names in either param.  It might also be worth supporting multiple source files, like the `scp` client does.

johnsca:
Regarding the failures in Travis, it looks like you need to use `os.path.expanduser()` (or `pathlib.Path().expanduser()`) and the host ended up being `None` so there's an issue with the logic for that.

Cynerva:
> I'm not clear on what the need is for separate scp_to vs scp_from commands?

The interface we had before didn't make sense to me:
```
unit = model.units['ubuntu/0']
unit.scp('localpath', 'ubuntu/0:remotepath')
```

Having a `Unit.scp` method where you have to specify the name of the unit seems silly. Here are the alternatives I came up with:

1. Split it up into `scp_to` and `scp_from`:
```
unit.scp_to('localpath', 'remotepath')
unit.scp_from('remotepath', 'localpath')
```
2. Use a simpler prefix like `remote`:
```
unit.scp('localpath', 'remote:remotepath')
```
3. Move the `scp` method to `Model`:
```
model.scp('localpath', 'ubuntu/0:remotepath')
```

I'm not a fan of APIs that require string manipulation -- seems error prone, and harder to read when you have to build strings dynamically -- so I went for option 1. I'm open to changing it however.

Cynerva:
Added the `os.path.expanduser` bit, thanks!

Regarding the hostname=None bit, I'm able to repro locally. The problem is that LXD machines are never assigned a public address. Here's what I see in `machine.data['addresses']` instead:
```
[{'scope': 'local-cloud', 'type': 'ipv4', 'value': '10.44.92.183'},
 {'scope': 'local-machine', 'type': 'ipv4', 'value': '127.0.0.1'},
 {'scope': 'local-machine', 'type': 'ipv6', 'value': '::1'}]
```

I'mma go stare at the Juju CLI golang for a while and see how they handle it.

Cynerva:
Rebased to resolve conflicts. Added a couple commits to get this working on LXD. This solution doesn't seem quite proper, but I can't come up with anything better.

Cynerva:
@johnsca The tests failed, but it doesn't look related to anything from this PR as far as I can tell. Any idea what's up?

Issue #157, opened 2017-09-19
upgrade_charm does not upgrade resources


Cynerva:
I wrote the upgrade_charm method, so this one's my fault, but:

The [Application.upgrade_charm](https://github.com/juju/python-libjuju/blob/master/juju/application.py#L318) method does not handle resource upgrades properly. I've observed this while upgrading from cs:~containers/flannel-26 to cs:~containers/flannel-31: the new `flannel-amd64` resource is not found by the charm.

This only occurs when upgrading via libjuju. The Juju CLI does not have the same problem.

I believe this is the step from the CLI implementation that we're missing: https://github.com/juju/juju/blob/7ddc8fb8ed8cbea361bd5867d9188af1b9e7cc92/cmd/juju/application/upgradecharm.go#L318-L327

Issue #145, opened 2017-06-14
libjuju fails on unknown facade versions


johnsca:
If libjuju is connected to a controller which supports a newer facade version than libjuju supports, it will fail instead of gracefully falling back to the maximum version that both it and the controller support (which is what the Juju client does).

We will also need to include a test which explicitly checks for any unknown facade versions so that can be run against upcoming Juju releases or PRs and inform us when new facades are available.

johnsca:
![](https://github.trello.services/images/mini-trello-icon.png) [Error when controller supports newer facade versions](https://trello.com/c/zmNDJ8CS/93-error-when-controller-supports-newer-facade-versions)

PR #158, opened 2017-09-25
Make Application.upgrade_charm upgrade resources

Cynerva:
Don't merge, this still needs integration tests. Opening for early review.

Fixes https://github.com/juju/python-libjuju/issues/157

The client is responsible for determining which resources to upgrade, determining which revisions to upgrade to, uploading the resources via `AddPendingResources`, and finally passing the new resource IDs to `SetCharm`. For reference, see [the same code in golang Juju](https://github.com/juju/juju/blob/7ddc8fb8ed8cbea361bd5867d9188af1b9e7cc92/cmd/juju/application/upgradecharm.go#L349-L463).

The `Resource` override adds missing `name` and `origin` fields. I would have expected to find that data within the `.charmresource` field, but in my testing, that field is always set to `None`. I don't know if this is the proper way to get the missing data, but it works in my testing at least.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -346,6 +346,9 @@ class Application(model.ModelEntity):
             self.connection)
         app_facade = client.ApplicationFacade.from_connection(self.connection)
 
+        charmstore = self.model.charmstore
+        charmstore_entity = None
+
         if switch is not None:
             charm_url = switch
             if not charm_url.startswith('cs:'):
@@ -356,9 +359,9 @@ class Application(model.ModelEntity):
             if revision is not None:
                 charm_url = "%s-%d" % (charm_url, revision)
             else:
-                charmstore = self.model.charmstore
-                entity = await charmstore.entity(charm_url, channel=channel)
-                charm_url = entity['Id']
+                charmstore_entity = await charmstore.entity(charm_url,
+                                                            channel=channel)
+                charm_url = charmstore_entity['Id']
 
         if charm_url == self.data['charm-url']:
             raise JujuError('already running charm "%s"' % charm_url)
@@ -370,9 +373,10 @@ class Application(model.ModelEntity):
         )
 
         # Update resources
-        response = await self.model.charmstore.entity(charm_url,
-                                                      channel=channel)
-        store_resources = response['Meta']['resources']
+        if not charmstore_entity:
+            charmstore_entity = await charmstore.entity(charm_url,
+                                                        channel=channel)
+        store_resources = charmstore_entity['Meta']['resources']
 
         request_data = [client.Entity(self.tag)]
         response = await resources_facade.ListResources(request_data)
--- a/juju/application.py
+++ b/juju/application.py
@@ -342,6 +342,8 @@ class Application(model.ModelEntity):
             raise ValueError("switch and revision are mutually exclusive")
 
         client_facade = client.ClientFacade.from_connection(self.connection)
+        resources_facade = client.ResourcesFacade.from_connection(
+            self.connection)
         app_facade = client.ApplicationFacade.from_connection(self.connection)
 
         if switch is not None:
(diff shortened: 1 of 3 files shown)


juju/application.py
```diff
@@ -361,11 +363,57 @@ def update_allocation(self, allocation):
...
 
+        # Update charm
         await client_facade.AddCharm(
             url=charm_url,
             channel=channel
         )
 
+        # Update resources
+        response = await self.model.charmstore.entity(charm_url,
+                                                      channel=channel)
```
johnsca: This is potentially duplicating a store request a few lines up.  Given how unreliable and slow these requests can be, we should avoid them as much as possible.  Given that this one is unconditionally required, we should move it up above where the other call is and change them to both use the result of a single call.

johnsca: Though, maybe instead (or in addition) we should look in to setting up a caching system for charmstore requests by revision.  I think entities can't really be modified without increasing their revision, so we shouldn't have to worry too much about cache expiration.  We could even potentially support an on-disk cache to speed things up across multiple runs, though then we would have to worry about pruning the cache at some point.

juju/application.py
```diff
@@ -361,11 +363,57 @@ def update_allocation(self, allocation):
...
+        response = await resources_facade.ListResources(request_data)
+        existing_resources = {
+            resource.name: resource
+            for resource in response.results[0].resources
+        }
+
+        resources_to_update = [
+            resource for resource in store_resources
+            if resource['Name'] not in existing_resources or
+            existing_resources[resource['Name']].origin != 'upload'
```
johnsca: Do we need to worry about cases where the revision hasn't changed, to avoid unnecessary fetches, or will the controller already handle detecting those?

juju/client/overrides.py
```diff
@@ -273,3 +274,26 @@ def __repr__(self):
...
+    def __init__(self, charmresource=None, application=None, id_=None, pending_id=None, timestamp=None, username=None, name=None, origin=None, **unknown_fields):
+        '''
+        charmresource : CharmResource
+        application : str
+        id_ : str
+        pending_id : str
+        timestamp : str
+        username : str
+        name: str
+        origin : str
```
johnsca: Why does this need to be in overrides?  Are these new fields from more recent builds of juju?  If that's the case, we should go ahead and follow the [upstream updates](https://pythonlibjuju.readthedocs.io/en/latest/upstream-updates/index.html) process to get all of the changes.  If these fields are not new, I'd like to figure out why they weren't captured already.

Cynerva: These aren't new, I've observed the same issue with Juju controllers running 2.0.2 and 2.2.3.

The schema is wrong. In particular, this bit of the Resource definition states that there's a field named CharmResource, defined by #/definitions/CharmResource:

```
"Resource": {
  "type": "object",
  "properties": {
    "CharmResource": {
      "$ref": "#/definitions/CharmResource"
    },
    ...
  }
  ...
}
```

where #/definitions/CharmResource defines the `name` and `origin` fields:

```
"CharmResource": {
  "type": "object",
  "properties": {
    ...
    "name": {
      "type": "string"
    },
    "origin": {
      "type": "string" 
    },
    ...
  },
  ...
},
```

However, here's the actual data I get in the response:

```
"resources": [
  {
    "username": "test/0",
    "type": "file",
    "pending-id": "",
    "name": "test-new",
    "fingerprint": "XwyivFD/IYBYXbtVo2VwSITDq8/yv+tncsHIPTnEY46+DjYpyL2JOupZXFTjSkz7",
    "revision": 3,
    "size": 10,
    "id": "test/test-new",
    "timestamp": "2017-09-27T15:07:53.923Z",
    "path": "test",
    "origin": "store",
    "application": "test"
  }
]
```

So Resource does not have a field named CharmResource. Rather, Resource _includes the fields from the CharmResource definition_. But that's not what the schema says.

After reading the [JSON Schema doc page that discusses $ref](https://spacetelescope.github.io/understanding-json-schema/structuring.html), it looks to me like the Resource definition should be using `allOf` to extend from CharmResource properly.

@johnsca Seems like I should open a bug against juju/schemagen, yeah?

johnsca: That seems to be how it's [defined](https://github.com/juju/juju/blob/develop/apiserver/params/resources.go#L78) in the juju source, though I guess the lack of an explicit `` `json:"<name>"` `` is supposed to make it be included inline.  We're still using a [fork](https://github.com/bcsaller/jsonschema) of the go [jsonschema](https://github.com/alecthomas/jsonschema) library which actually does the reflection, so this might already fixed upstream.  I created juju/schemagen#1 to track getting us off the fork or at least getting it updated.

juju/application.py
Cynerva: I'm gonna punt on this one, sorry. FWIW I didn't see the golang code comparing revisions so we're probably okay.

Cynerva:
@johnsca I added an integration test and removed the duplicate call to the charm store. Can you take another look?

johnsca:
Travis failures were due to charm store outage.  Restarted.

johnsca:
Tests are now failing on #145 and #162.  We need to address those, but I'm +1 to merging this now anyway, since those failures aren't the fault of this PR.

PR #161, opened 2017-09-28
removing cli command to add ssh keys

sebastienpattyn93:
I removed the cli call that was causing issue #160. The issue occured because we hadn't used the cli in more than 24 hours so the macaroons expired. That's why the cli command to add ssh-key was not responding. I could easily reproduce it by deleting the macaroons and check what happens if I used a cli command.

`please enter password for admin on google-controller: ERROR cannot log in: context deadline exceeded`
 The ERROR only came after 10 minutes so that would explain why it took so long to get the response.

So removing the cli call fixed it. I was wondering if it is necessary to afterwords call the `add_ssh_key` after connecting to the model? or should this not be included into the `add_model` wrapper?

### DIFF
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -116,7 +116,7 @@ class Controller(object):
             loop=self.loop,
         )
         ssh_key = await utils.read_ssh_key(loop=self.loop)
-        model.add_ssh_key(owner, ssh_key)
+        await model.add_ssh_key(owner, ssh_key)
 
         return model
 
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -95,6 +95,10 @@ class Controller(object):
             credential = None
 
         log.debug('Creating model %s', model_name)
+        
+        if not config or 'authorized-keys' not in config:
+            config = config or {}
+            config['authorized-keys'] = await utils.read_ssh_key(loop=self.loop)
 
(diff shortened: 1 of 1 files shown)


johnsca:
I think the tests failures are due to #145 and #162 and not related to this PR.

sebastienpattyn93:
@johnsca Thank you for your respond. I was already debugging and could not find the issue since everything was working fine locally. Should the `model.add_ssh_key(user,key)` be implemented in the add_model wrapper?

johnsca:
It looks like the SSH key is actually supposed to be [injected](https://github.com/juju/juju/blob/develop/cmd/juju/controller/addmodel.go#L589) into the model config passed in to `CreateModel` and it was something we originally missed and subsequently misattributed to the upstream bug and "work-around" by falling back to the CLI.

We should fix it correctly using something like:

```python
if not config or 'authorized-keys' not in config:
    config = config or {}
    config['authorized-keys'] = await utils.read_ssh_key(loop=self.loop)
```

petevg:
+1 to @johnsca's suggestion. That looks like the correct thing to do.

sebastienpattyn93:
Adding @johnsca solutions works great, Not getting issue #160 anymore. Thanks a lot!

PR #163, opened 2017-09-28
Fix test failures

### LINKED ISSUE #145: libjuju fails on unknown facade versions

### LINKED ISSUE #162: test_explicit_loop fails on Python 3.5.3 & 3.6.0
Starting with Python 3.6.0 and backported to 3.5.3 (which Travis is now using, apparently), the `test_explicit_loop` test fails.  It might be a [bug](https://stackoverflow.com/a/42566336) but it's still affecting 3.6.2, so I'm guessing it's intended.  We should remove that test, and just keep `test_explicit_loop_threaded` since that still passes and was the intended use case.

johnsca:
Fix a few test errors that have cropped up.

First is due to changes in asyncio and incorrect understanding of how multiple event loops should be used, and fixes #162.

Second is due to the addition of the user name to the fully-qualified model name and `Controller.add_model` not adding anything to the on-disk client yaml data.

Third is tests failing on Juju edge due to new facade versions that we don't yet support.  This is addressed by gracefully downgrading to the highest version that we do support, which fixes #145.


tests/base.py
```diff
@@ -34,28 +34,51 @@ def __init__(self):
...
         self.controller = Controller()
-        await self.controller.connect_current()
+        juju_data = JujuData()
+        self.controller_name = juju_data.current_controller()
+        self.user_name = juju_data.accounts()[self.controller_name]['user']
+        await self.controller.connect_controller(self.controller_name)
 
-        model_name = 'model-{}'.format(uuid.uuid4())
-        self.model = await self.controller.add_model(model_name)
+        self.model_name = 'model-{}'.format(self.user_name, uuid.uuid4())
```
petevg: Should this throw a fit if self.user_name is not set?

tests/integration/test_model.py
```diff
@@ -131,18 +131,6 @@
 
 @base.bootstrapped
 @pytest.mark.asyncio
-async def test_explicit_loop(event_loop):
```
tvansteenburgh: Can you explain why this test is invalid?

juju/controller.py
```diff
@@ -39,9 +40,11 @@ def __init__(self, loop=None,
         """Connect to the current Juju controller.
 
         """
-        self.connection = (
-            await connection.Connection.connect_current_controller(
-                max_frame_size=self.max_frame_size))
+        jujudata = connection.JujuData()
```
tvansteenburgh: What's the reasoning behind this change? Generally I think we have tried to read juju data from disk as little as possible, and only when there was no other option, because relying on it basically means that you need the cli installed in order to use the python lib.

tests/base.py
```diff
@@ -34,28 +34,51 @@ def __init__(self):
...
         self.controller = None
+        self.controller_name = None
         self.model = None
+        self.model_name = None
+        self.model_uuid = None
 
     async def __aenter__(self):
         self.controller = Controller()
-        await self.controller.connect_current()
+        juju_data = JujuData()
```
tvansteenburgh: Same question as above.

tests/integration/test_model.py
johnsca: The error ("Cannot run the event loop while another loop is running") occurs because you're not actually supposed to have multiple loops running in a single thread.  It wasn't explicitly checked for until 3.5.3, but we should not have been doing this in the first place.

juju/controller.py
johnsca: The juju data was already being read in `Connection.connect_current_controller` but then the controller name wasn't being saved on the `Controller` instance.  This was actually from an earlier approach to fixing the test issue I'll mention below, but I think it's a bit nicer to ensure that the `Controller.controller_name` is set when it can be.

tests/base.py
johnsca: This is to fix the test failing with `KeyError`, e.g. https://travis-ci.org/juju/python-libjuju/jobs/281013483

When the model is added with `Controller.add_model`, the juju data isn't updated.  But since a few of the tests use `connect_current`, it was trying to read the data about the current model.  We patched in a portion of that, but didn't have the model UUID in there.  TBH, I'm not sure how those tests ever worked, or at what point they broke.

Some of the tests affected by this could be cleaned up by just getting the connection from the `CleanModel` directly, but at least the test for `connect_current` itself requires this.

johnsca:
[Error when controller supports newer facade versions](https://trello.com/c/zmNDJ8CS/93-error-when-controller-supports-newer-facade-versions)

johnsca:
[Test failures due to upstream changes](https://trello.com/c/sw9IiCqY/94-test-failures-due-to-upstream-changes)

PR #164, opened 2017-10-05
Removing a user with libjuju

sebastienpattyn93:
I added a wrapper in `controller.py` to remove a user on a specific controller. I also wrote an integration test to test the call but I'm not sure if the test is completely right. I'm not sure if line 26-28 are necessary.


sebastienpattyn93:
apparently my commits from last week still appear here, which is strange since there are no changes in the first 6 commits since this was merged already.

johnsca:
@sebastienpattyn93 It's because we use "Squash & Merge" for PRs and you didn't use separate feature branches on your fork.  As long as it's not causing merge conflicts, it's a non-issue, but if you wanted to clean it up, you could create a feature branch, cherry-pick the relevant commits in, and force-reset your master to the upstream master.  You'd need to recreate the PR, I think, though, so I wouldn't bother with it for now.  I would recommend force-resetting your fork's master to upstream once this is merged and using feature branches in the future, though.

The integration test is failing on Travis with "permission denied" from Juju on the `get_user()` call after the user is removed.  Should that be wrapped in a `with self.assertRaises(JujuError):`?

sebastienpattyn93:
@johnsca thank you for the information on the PR's. 

About the `with self.assertRaises(JujuError)`. Can I just change my test to this => `async def test_remove_user(self, event_loop)`? 
Because otherwise I can't use the `self`. Isn't it a possibility to wrap this in a `try/except` and catch the JujuError there?

johnsca:
Sorry, I forgot we're not using `unittest.TestCase`.  You'll want to use [`with pytest.raises(JujuError):`](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) instead.

johnsca:
I'm also working on a PR that will change the return value of `get_user`, because we really shouldn't be directly; we should be wrapping that in an object layer class.  But I can resolve the conflicts on that when the time comes.

johnsca:
I ended up incorporating this change into #170 so this should be superseded by that.  Please take a look and make sure it works for you.

sebastienpattyn93:
@johnsca  that looks great! thanks a lot for changing the user handling in general. I will be able to start testing this tomorrow and I'll keep you updated on my progress.

PR #165, opened 2017-10-05
Support 'applications' key in bundles

### LINKED ISSUE #110: Suggestion: using applications instead of services
In `model.py` I found the following code on line 1644, but I was wondering if it wouldn't be better to change everything to applications (same counts for lines 1589 and 1618) when bundles have to be deployed, since juju is now using Applications instead of Services. I am using the word Applications in my json bundles so I keep getting KeyErrors when I want to deploy my Json with libjuju. And I would prefer to use the same bundle for libjuju as I'm using for juju.
```
# line 1589
    app_dict = bundle['services'][app_name]

# line 1618
   bundle['services'][app_name]['charm'] = charm_url

# line 1644
    @property
    def applications(self):
        return list(self.bundle['services'].keys())
```

johnsca:
Fixes #110

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1658,8 +1658,9 @@ class BundleHandler(object):
         apps, args = [], []
 
         default_series = bundle.get('series')
+        apps_dict = bundle.get('applications', bundle.get('services', {}))
         for app_name in self.applications:
-            app_dict = bundle['services'][app_name]
+            app_dict = apps_dict[app_name]
             charm_dir = os.path.abspath(os.path.expanduser(app_dict['charm']))
             if not os.path.isdir(charm_dir):
                 continue
@@ -1688,7 +1689,7 @@ class BundleHandler(object):
             ], loop=self.model.loop)
             # Update the 'charm:' entry for each app with the new 'local:' url.
             for app_name, charm_url in zip(apps, charm_urls):
-                bundle['services'][app_name]['charm'] = charm_url
+                apps_dict[app_name]['charm'] = charm_url
 
         return bundle
 
@@ -1714,7 +1715,9 @@ class BundleHandler(object):
 
     @property
     def applications(self):
-        return list(self.bundle['services'].keys())
+        apps_dict = self.bundle.get('applications',
+                                    self.bundle.get('services', {}))
+        return list(apps_dict.keys())
 
     def resolve(self, reference):
         if reference and reference.startswith('$'):


tvansteenburgh:
LGTM but Travis is upset about it.

johnsca:
There seems to be something up with the edge channel of Juju.  The failure was:

```
    @base.bootstrapped
    @pytest.mark.asyncio
    async def test_get_models(event_loop):
        async with base.CleanController() as controller:
>           result = await controller.get_models()
tests/integration/test_controller.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
juju/controller.py:234: in get_models
    return await controller_facade.AllModels()
juju/client/facade.py:392: in wrapper
juju/client/_client3.py:892: in AllModels
    reply = await self.rpc(msg)
juju/client/facade.py:514: in rpc
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <juju.client.connection.Connection object at 0x7f67d4056da0>
msg = {'params': {}, 'request': 'AllModels', 'request-id': 2, 'type': 'Controller', ...}
encoder = <class 'juju.client.facade.TypeEncoder'>
    async def rpc(self, msg, encoder=None):
        self.__request_id__ += 1
        msg['request-id'] = self.__request_id__
        if'params' not in msg:
            msg['params'] = {}
        if "version" not in msg:
        for attempt in range(3):
            try:
                await self.ws.send(outgoing)
                break
            except websockets.ConnectionClosed:
                if attempt == 2:
                    raise
                log.warning('RPC: Connection closed, reconnecting')
                # the reconnect has to be done in a separate task because,
                # if it is triggered by the pinger, then this RPC call will
                await asyncio.wait([self.reconnect()], loop=self.loop)
        result = await self.recv(msg['request-id'])
    
        if not result:
            return result
    
        if 'error' in result:
            # API Error Response
>           raise JujuAPIError(result)
E           juju.errors.JujuAPIError: model bf33dece-a932-4f6c-85ad-2b215fb39c6b has been removed
```

It seems that another test being run in parallel was cleaning up its model which caused Juju to blow up.  We could retry that in the `get_models()` call, perhaps, but maybe we should open it as a bug against Juju.  It might also be an issue in the stable channel and we just haven't hit it before.

PR #170, opened 2017-10-08
Refactor users handling, add get_users

### LINKED ISSUE #95: UserInfo
Is there a call on controller level to get all the users on the controller?
On model level, the users attribute of a model object returns an empty list, despite the fact there are users with access to this model. 
Is there a way to get the model access for a certain user?

johnsca:
Fixes #95 

Also adds linting and fixes all the lint errors.

### DIFF
--- a/juju/client/overrides.py
+++ b/juju/client/overrides.py
@@ -275,10 +275,28 @@ class ConfigValue(_definitions.ConfigValue):
                                                 repr(self.source),
                                                 repr(self.value))
 
+
 class Resource(Type):
-    _toSchema = {'application': 'application', 'charmresource': 'CharmResource', 'id_': 'id', 'pending_id': 'pending-id', 'timestamp': 'timestamp', 'username': 'username', 'name': 'name', 'origin': 'origin'}
-    _toPy = {'CharmResource': 'charmresource', 'application': 'application', 'id': 'id_', 'pending-id': 'pending_id', 'timestamp': 'timestamp', 'username': 'username', 'name': 'name', 'origin': 'origin'}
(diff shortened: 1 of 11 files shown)


johnsca:
Test failure is #172.  Unrelated to this, so I'm trying re-running the job.

juju/client/overrides.py
```diff
@@ -275,10 +275,28 @@ def __repr__(self):
                                                 repr(self.source),
                                                 repr(self.value))
 
+
 class Resource(Type):
-    _toSchema = {'application': 'application', 'charmresource': 'CharmResource', 'id_': 'id', 'pending_id': 'pending-id', 'timestamp': 'timestamp', 'username': 'username', 'name': 'name', 'origin': 'origin'}
-    _toPy = {'CharmResource': 'charmresource', 'application': 'application', 'id': 'id_', 'pending-id': 'pending_id', 'timestamp': 'timestamp', 'username': 'username', 'name': 'name', 'origin': 'origin'}
-    def __init__(self, charmresource=None, application=None, id_=None, pending_id=None, timestamp=None, username=None, name=None, origin=None, **unknown_fields):
+    _toSchema = {'application': 'application',
```
battlemidget: Is it worth a doc string here? I'm guessing the Resource class is heavily used?

johnsca: This is essentially an override of generated code.  At least some sort of comment to indicate what we're overriding and why makes sense.

johnsca: FWIW, the changes here are strictly for fixing lint errors.  Might have been better to just exclude this file, since it's mostly generated code; but it's also a mix of manually created code as well.

johnsca: On second thought, since this is just lint fixes, I'm going to punt on adding comments to this for now.

PR #171, opened 2017-10-08
Refactor model handling

### LINKED ISSUE #54: controller.destroy-models should accept names of the models
I've created a script that adds a new model but then I went to clean it up and found the destroy call takes the uuid. I think it's worthwhile to allow the user to supply the name and leave it to the function call to deal with uuids as internal implementation details. To the user the model names is what they use/live off of via juju models calls and such.

johnsca:
This changes get_models to list_models, implements get_model, and allows destroy_models to accept model names (fixes #54).

Unfortunately, this makes us much more likely to hit https://bugs.launchpad.net/juju/+bug/1721786 with the tests.

### DIFF
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -177,24 +177,28 @@ class Controller(object):
 
         return model
 
-    async def destroy_models(self, *uuids):
+    async def destroy_models(self, *models):
         """Destroy one or more models.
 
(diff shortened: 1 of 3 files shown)


johnsca:
Upstream bug has been fixed.  :tada:

PR #182, opened 2017-11-13
juju/client: factor out JujuData class

rogpeppe:
We move JujuData into its own file because it's logically
independent of the Connection class.

We also make it cache the data that's been read so that we don't need
to read and parse the YAML file every time a method
is called.

We also change it so that it doesn't shell out to Juju to obtain the
current controller name - the name can be retrieved from the controllers.yaml
file.

In a subsequent PR, we'll factor those Connection methods
that depend on JujuData out of the Connection class so
that Connection doesn't need to depend on the details of
the juju configuration files.

### DIFF
--- a/juju/client/jujudata.py
+++ b/juju/client/jujudata.py
@@ -1,0 +1,129 @@
+import os
+import yaml
+import io
+
+from juju import tag
+import juju.client.client as jujuclient
+from juju.errors import JujuError
+
+
+class JujuData:
+    '''Provide access to the Juju client configuration files.
+    Any configuration file is read once and then cached.'''
+    def __init__(self):
+        self.path = os.environ.get('JUJU_DATA') or '~/.local/share/juju'
+        self.path = os.path.abspath(os.path.expanduser(self.path))
+        # _loaded keeps track of the loaded YAML from
+        # the Juju data files so we don't need to load the same
+        # file many times.
+        self._loaded = {}
+
+    def refresh(self):
+        '''Forget the cache of configuration file data'''
+        self._loaded = {}
+
+    def current_controller(self):
+        '''Return the current controller name'''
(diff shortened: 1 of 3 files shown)


juju/controller.py
```diff
@@ -90,11 +91,10 @@ def __init__(self, loop=None,
             raise errors.JujuError('Name must be provided for credential')
 
         if not credential:
-            name, credential = connection.JujuData().load_credential(cloud,
-                                                                     name)
+            name, credential = JujuData().load_credential(cloud, name)
             if credential is None:
-                raise errors.JujuError('Unable to find credential: '
-                                       '{}'.format(name))
+                raise errors.JujuError(
```
frankban: Thanks for reformatting like that.

juju/client/jujudata.py
```diff
@@ -0,0 +1,129 @@
...
+class JujuData:
+    '''Provide access to the Juju client configuration files.
+    Any configuration file is read once and then cached.'''
+    def __init__(self):
+        self.path = os.environ.get('JUJU_DATA') or '~/.local/share/juju'
+        self.path = os.path.abspath(os.path.expanduser(self.path))
+        # _loaded keeps track of the loaded YAML from
+        # the Juju data files so we don't need to load the same
+        # file many times.
+        self._loaded = {}
```
frankban: Nice.

rogpeppe:
[libjuju: factor out JujuData](https://trello.com/c/Q3oW6T87/27-libjuju-factor-out-jujudata)

rogpeppe:
FWIW I see the same test errors when running `tox -e integration` on the current master (e64a5d1759e826c857021ddbf081132323e5a109), so I don't think this PR has introduced the problem.

PR #187, opened 2017-11-23
refactor connections prior to bakery authentication

rogpeppe:
We do a bunch of refactoring in advance of actually adding the bakery authentication. No additional functionality as yet (hence no new tests) but this is a breaking API change in a few ways.

API changes
----

Model and Controller now use the connect method for both connecting to a named model and for connecting to the current model (if the name is empty or None); the connect_current and connect_model methods are removed.

Connection.connect is now a class method that returns a new connected connection.

Various fields in Connection, Model and Controller that are really internal-only are given a leading underscore.

Controller and Model now have connection methods not properties, and they raise an exception if there is no current connection.

Other changes that should not break the API
----

Add a bakery_client argument to Connection.connect so that the client can have control over what interaction methods are used and provide a cookie jar.

Use "-q -B" options to scp so that it doesn't print spurious output when it's used.

Avoid unnecessary explicit inheritance from object.

Add is_connected method to Controller and Model.

Made new Connector class factoring out code in common between Controller and Model.

Factor Connection._connect so that we can do the bakery authentication.

Connection has a new connect_params method that returns parameters suitable for passing to Connection.connect.

Factor out all JujuData knowledge from connection.py.

Factored out commonality from Connection.pinger and Connection.receiver to _Task, making it easier to disable or mock those tasks in tests.

### DIFF
--- a/examples/action.py
+++ b/examples/action.py
@@ -27,7 +27,7 @@ async def run_action(unit):
 
 async def main():
     model = Model()
-    await model.connect_current()
+    await model.connect()
     await model.reset(force=True)
 
     app = await model.deploy(
--- a/juju/client/_client.py
+++ b/juju/client/_client.py
@@ -41,7 +41,13 @@ class TypeFactory:
         @param connection: initialized Connection object.
 
         """
-        version = connection.facades[cls.__name__[:-6]]
+        facade_name = cls.__name__
+        if not facade_name.endswith('Facade'):
+           raise TypeError('Unexpected class name: {}'.format(facade_name))
+        facade_name = facade_name[:-len('Facade')]
+        version = connection.facades.get(facade_name)
+        if version is None:
+            raise Exception('No facade {} in facades {}'.format(facade_name, connection.facades))
 
         c = lookup_facade(cls.__name__, version)
         c = c()
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -2,19 +2,17 @@ import asyncio
 import base64
 import json
 import logging
-import random
 import ssl
-import string
+import urllib.request
 import weakref
 from concurrent.futures import CancelledError
 from http.client import HTTPSConnection
-from pathlib import Path
 
+import macaroonbakery.httpbakery as httpbakery
 import websockets
-from juju import tag, utils
+from juju import utils
 from juju.client import client
-from juju.client.jujudata import JujuData
-from juju.errors import JujuAPIError, JujuConnectionError, JujuError
+from juju.errors import JujuAPIError, JujuError
 from juju.utils import IdQueue
 
 log = logging.getLogger("websocket")
@@ -25,7 +23,7 @@ class Monitor:
     Monitor helper class for our Connection class.
 
     Contains a reference to an instantiated Connection, along with a
-    reference to the Connection.receiver Future. Upon inspecttion of
+    reference to the Connection.receiver Future. Upon inspection of
     these objects, this class determines whether the connection is in
     an 'error', 'connected' or 'disconnected' state.
 
@@ -43,10 +41,6 @@ class Monitor:
         self.connection = weakref.ref(connection)
         self.reconnecting = asyncio.Lock(loop=connection.loop)
         self.close_called = asyncio.Event(loop=connection.loop)
-        self.receiver_stopped = asyncio.Event(loop=connection.loop)
-        self.pinger_stopped = asyncio.Event(loop=connection.loop)
-        self.receiver_stopped.set()
-        self.pinger_stopped.set()
 
     @property
     def status(self):
@@ -76,7 +70,7 @@ class Monitor:
             return self.DISCONNECTING
 
         # connection closed uncleanly (we didn't call connection.close)
-        if self.receiver_stopped.is_set() or not connection.ws.open:
+        if connection._receiver_task.stopped.is_set() or not connection.ws.open:
             return self.ERROR
 
         # everything is fine!
@@ -91,47 +85,87 @@ class Connection:
         client = await Connection.connect(
             api_endpoint, model_uuid, username, password, cacert)
 
-        # Connect using a controller/model name
-        client = await Connection.connect_model('local.local:default')
-
-        # Connect to the currently active model
-        client = await Connection.connect_current()
-
     Note: Any connection method or constructor can accept an optional `loop`
     argument to override the default event loop from `asyncio.get_event_loop`.
     """
 
-    DEFAULT_FRAME_SIZE = 'default_frame_size'
     MAX_FRAME_SIZE = 2**22
     "Maximum size for a single frame.  Defaults to 4MB."
 
-    def __init__(
-            self, endpoint, uuid, username, password, cacert=None,
-            macaroons=None, loop=None, max_frame_size=DEFAULT_FRAME_SIZE):
-        self.endpoint = endpoint
-        self._endpoint = endpoint
+    @classmethod
+    async def connect(
+            cls,
+            endpoint,
+            uuid=None,
+            username=None,
+            password=None,
+            cacert=None,
+            bakery_client=None,
+            loop=None,
+            max_frame_size=None,
+    ):
+        """Connect to the websocket.
+
+        If uuid is None, the connection will be to the controller. Otherwise it
+        will be to the model.
+        :param str endpoint The hostname:port of the controller to connect to.
+        :param str uuid The model UUID to connect to (None for a
+            controller-only connection).
+        :param str username The username for controller-local users (or None
+            to use macaroon-based login.)
+        :param str password The password for controller-local users.
+        :param str cacert The CA certificate of the controller (PEM formatted).
+        :param httpbakery.Client bakery_client The macaroon bakery client to
+            to use when performing macaroon-based login. Macaroon tokens
+            acquired when logging will be saved to bakery_client.cookies.
+            If this is None, a default bakery_client will be used.
+        :param loop asyncio.BaseEventLoop The event loop to use for async
+            operations.
+        :param max_frame_size The maximum websocket frame size to allow.
+        """
+        self = cls()
         self.uuid = uuid
-        if macaroons:
-            self.macaroons = macaroons
-            self.username = ''
-            self.password = ''
-        else:
-            self.macaroons = []
-            self.username = username
-            self.password = password
-        self.cacert = cacert
-        self._cacert = cacert
+        if bakery_client is None:
+            bakery_client = httpbakery.Client()
+        self.bakery_client = bakery_client
+        self.usertag = username
+        if username is not None and not username.startswith('user-'):
+            # TODO this doesn't seem quite right, as 'user-foo' is
+            # actually a valid username (its tag would be user-user-foo),
+            # but don't break the existing API.
+            self.usertag = 'user-' + username
+        self.password = password
         self.loop = loop or asyncio.get_event_loop()
 
         self.__request_id__ = 0
+
+        # The following instance variables are initialized by the
+        # _connect_with_redirect method, but create them here
+        # as a reminder that they will exist.
         self.addr = None
         self.ws = None
+        self.endpoint = None
+        self.cacert = None
+        self.info = None
+
+        # Create that _Task objects but don't start the tasks yet.
+        self._pinger_task = _Task(self._pinger, self.loop)
+        self._receiver_task = _Task(self._receiver, self.loop)
+
         self.facades = {}
         self.messages = IdQueue(loop=self.loop)
         self.monitor = Monitor(connection=self)
-        if max_frame_size is self.DEFAULT_FRAME_SIZE:
+        if max_frame_size is None:
             max_frame_size = self.MAX_FRAME_SIZE
         self.max_frame_size = max_frame_size
+        await self._connect_with_redirect([(endpoint, cacert)])
+        return self
+
+    @property
+    def username(self):
+        if not self.usertag:
+            return None
+        return self.usertag[len('user-'):]
 
     @property
     def is_open(self):
@@ -141,39 +175,34 @@ class Connection:
         return ssl.create_default_context(
             purpose=ssl.Purpose.CLIENT_AUTH, cadata=cert)
 
-    async def open(self):
+    async def _open(self, endpoint, cacert):
         if self.uuid:
-            url = "wss://{}/model/{}/api".format(self.endpoint, self.uuid)
+            url = "wss://{}/model/{}/api".format(endpoint, self.uuid)
         else:
-            url = "wss://{}/api".format(self.endpoint)
+            url = "wss://{}/api".format(endpoint)
 
-        kw = dict()
-        kw['ssl'] = self._get_ssl(self.cacert)
-        kw['loop'] = self.loop
-        kw['max_size'] = self.max_frame_size
-        self.addr = url
-        self.ws = await websockets.connect(url, **kw)
-        self.loop.create_task(self.receiver())
-        self.monitor.receiver_stopped.clear()
-        log.info("Driver connected to juju %s", url)
-        self.monitor.close_called.clear()
-        return self
+        return (await websockets.connect(
+            url,
+            ssl=self._get_ssl(cacert),
+            loop=self.loop,
+            max_size=self.max_frame_size,
+        ), url)
 
     async def close(self):
         if not self.ws:
             return
         self.monitor.close_called.set()
-        await self.monitor.pinger_stopped.wait()
-        await self.monitor.receiver_stopped.wait()
+        await self._pinger_task.stopped.wait()
+        await self._receiver_task.stopped.wait()
         await self.ws.close()
         self.ws = None
 
-    async def recv(self, request_id):
+    async def _recv(self, request_id):
         if not self.is_open:
             raise websockets.exceptions.ConnectionClosed(0, 'websocket closed')
         return await self.messages.get(request_id)
 
-    async def receiver(self):
+    async def _receiver(self):
         try:
             while self.is_open:
                 result = await utils.run_with_interrupt(
@@ -200,10 +229,8 @@ class Connection:
             # make pending listeners aware of the error
             await self.messages.put_all(e)
             raise
-        finally:
-            self.monitor.receiver_stopped.set()
 
-    async def pinger(self):
+    async def _pinger(self):
         '''
         A Controller can time us out if we are silent for too long. This
         is especially true in JaaS, which has a fairly strict timeout.
@@ -219,19 +246,23 @@ class Connection:
                 pass
 
         pinger_facade = client.PingerFacade.from_connection(self)
-        try:
-            while True:
-                await utils.run_with_interrupt(
-                    _do_ping(),
-                    self.monitor.close_called,
-                    loop=self.loop)
-                if self.monitor.close_called.is_set():
-                    break
-        finally:
-            self.monitor.pinger_stopped.set()
-            return
+        while True:
+            await utils.run_with_interrupt(
+                _do_ping(),
+                self.monitor.close_called,
+                loop=self.loop)
+            if self.monitor.close_called.is_set():
+                break
 
     async def rpc(self, msg, encoder=None):
+        '''Make an RPC to the API. The message is encoded as JSON
+        using the given encoder if any.
+        :param msg: Parameters for the call (will be encoded as JSON).
+        :param encoder: Encoder to be used when encoding the message.
+        :return: The result of the call.
+        :raises JujuAPIError: When there's an error returned.
+        :raises JujuError:
+        '''
         self.__request_id__ += 1
         msg['request-id'] = self.__request_id__
         if'params' not in msg:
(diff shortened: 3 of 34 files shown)


juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
+import asyncio
+import logging
+import macaroonbakery.httpbakery as httpbakery
```
frankban: This goes to a second import block of external libraries.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+
+from juju.client.connection import Connection
+from juju.client.jujudata import JujuData
+from juju.errors import JujuConnectionError, JujuError
+
+log = logging.getLogger('connector')
+
+
+class Connector:
```
frankban: Docstring.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+        self.controller_name = None
+        self.model_name = None
+        self.jujudata = JujuData()
+
+    def is_connected(self):
+        return self._connection is not None
+
+    def connection(self):
+        if not self.is_connected():
+            raise Exception('not connected')
```
frankban: Please raise a subclass of Exception so that callers can target it specifically.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+        return self._connection
+
+    async def connect(self, *args, **kw):
+        """Connect to an arbitrary Juju model.
+
+        args and kw are passed through to Connection.connect()
+
```
frankban: Unnecessary newline.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+        self.bakery_client = bakery_client
+        self._connection = None
+        self.controller_name = None
+        self.model_name = None
+        self.jujudata = JujuData()
+
+    def is_connected(self):
+        return self._connection is not None
+
+    def connection(self):
```
frankban: Docstring.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, loop=None, max_frame_size=None, bakery_client=None):
+        self.max_frame_size = max_frame_size
+        self.loop = loop or asyncio.get_event_loop()
+        self.bakery_client = bakery_client
+        self._connection = None
+        self.controller_name = None
+        self.model_name = None
+        self.jujudata = JujuData()
+
+    def is_connected(self):
```
frankban: Dosctring.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+
+        """
+        kw.setdefault('loop', self.loop)
+        kw.setdefault('max_frame_size', self.max_frame_size)
+        kw.setdefault('bakery_client', self.bakery_client)
+        self._connection = await Connection.connect(*args, **kw)
+
+    async def disconnect(self):
+        """Shut down the watcher task and close websockets.
+        """
```
frankban: You can close this docstring in the same line. [shrug]

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+        if self._connection:
+            log.debug('Closing model connection')
+            await self._connection.close()
+            self._connection = None
+
+    async def connect_controller(self, controller_name):
+        """Connect to a controller by name. If the name is empty, it
+        connect to the current controller.
```
frankban: So make controller_name optional? Also s/controller_name/name maybe.

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
...
+            endpoint=endpoint,
+            uuid=None,
+            username=accounts.get('user'),
+            password=accounts.get('password'),
+            cacert=controller.get('ca-cert'),
+            bakery_client=self.bakery_client_for_controller(controller_name),
+        )
+        self.controller_name = controller_name
+
+    async def connect_model(self, model):
```
frankban: s/model/name/ or model_name for consistency with connect_controller.

juju/client/jujudata.py
```diff
@@ -1,9 +1,11 @@
 import os
```
frankban: Please put these in alphabetical order.

juju/client/jujudata.py
```diff
@@ -66,10 +68,10 @@ def parse_model(self, model):
         if '/' not in model_name:
             # model name doesn't include a user prefix, so add one
             # by using the current user for the controller.
-            accounts = self.accounts().get('controller_name')
```
frankban: Woah, was this an existing bug?

juju/client/jujudata.py
```diff
@@ -127,3 +129,11 @@ def _load_yaml(self, filename, key):
             data = yaml.safe_load(f)
             self._loaded[filename] = data
             return data.get(key)
+
+    def cookies_for_controller(self, controller_name):
+        f = pathlib.Path(self.path) / 'cookies' / controller_name
```
frankban: Ah, this is the first time I see the overloading of / to create a path. Not sure how I feel about that.

juju/controller.py
```diff
@@ -5,66 +5,61 @@
 from . import tag
 from . import utils
 from .client import client
```
frankban: `from .client import client, connector` and remove the line below.

juju/unit.py
```diff
@@ -2,6 +2,7 @@
 
 import pyrfc3339
 
+
```
frankban: Unnecessary new line.

setup.py
```diff
@@ -34,6 +34,7 @@
         'pyyaml',
         'theblues',
         'pyRFC3339>=1.0,<2.0',
+        'macaroonbakery<1.0',
```
frankban: It would be nice to keep this in alphabetical order.

tests/integration/test_controller.py
```diff
@@ -5,7 +5,7 @@
 from .. import base
 from juju.controller import Controller
 from juju.errors import JujuAPIError
-
+from juju.client.connection import Connection
```
frankban: Alphabetical order.

tests/integration/test_machine.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import pytest
```
frankban: pytest shoudl go in the second import group.

tests/unit/test_connection.py
```diff
@@ -42,13 +41,24 @@ def __init__(self, responses):
         {'request-id': 2},
         {'request-id': 3},
     ]
-    con._get_sll = mock.MagicMock()
+    minimal_facades = [{'name':'Pinger','versions':[1]}]
```
frankban: Are those files parsed  by lint? This dict here seems to miss the usual spaces after ":" and ",".

tests/unit/test_connection.py
```diff
@@ -42,13 +41,24 @@ def __init__(self, responses):
...
+    con = None
     try:
-        with mock.patch('websockets.connect', base.AsyncMock(return_value=ws)):
-            await con.open()
+        with \
```
frankban: I think this kind of blocks are slightly more readable if written as
```python
with (
    mock.patch('websockets.connect', base.AsyncMock(return_value=ws)),
    mock.patch(...),
):
    con = await Connection.connect('0.1.2.3:999')
```
So with avoiding back slashes and making more clear where the block actually starts.

juju/client/connection.py
```diff
@@ -92,47 +89,87 @@ class Connection:
...
+            cacert=None,
+            bakery_client=None,
+            loop=None,
+            max_frame_size=None,
+    ):
+        """Connect to the websocket.
+
+        If uuid is None, the connection will be to the controller. Otherwise it
+        will be to the model.
+        :param str endpoint The hostname:port of the controller to connect to.
```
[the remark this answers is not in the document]
rogpeppe: That's for Model and Controller, not for Connection.

juju/client/connection.py
```diff
@@ -220,26 +250,31 @@ def _get_ssl(self, cert=None):
...
+                self.monitor.close_called,
+                loop=self.loop)
+            if self.monitor.close_called.is_set():
+                break
 
     async def rpc(self, msg, encoder=None):
+        '''Make an RPC to the API. The message is encoded as JSON
+        using the given encoder if any.
+        :param msg: Parameters for the call (will be encoded as JSON).
+        :param encoder: Encoder to be used when encoding the message.
```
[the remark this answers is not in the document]
rogpeppe: Doesn't the fact that the default is None imply that it's optional?

juju/client/connection.py
```diff
@@ -220,26 +250,31 @@ def _get_ssl(self, cert=None):
...
+        :raises JujuError:
+        '''
         self.__request_id__ += 1
         msg['request-id'] = self.__request_id__
         if'params' not in msg:
             msg['params'] = {}
         if "version" not in msg:
             msg['version'] = self.facades[msg['type']]
         outgoing = json.dumps(msg, indent=2, cls=encoder)
+        log.debug('connection {} -> {}'.format(id(self), outgoing))
```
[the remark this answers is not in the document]
rogpeppe: No, I think it's useful to have debug-level messages that show incoming and outgoing messages.

juju/client/connection.py
```diff
@@ -253,7 +288,8 @@ def _get_ssl(self, cert=None):
                 # be cancelled when the pinger is cancelled by the reconnect,
                 # and we don't want the reconnect to be aborted halfway through
                 await asyncio.wait([self.reconnect()], loop=self.loop)
-        result = await self.recv(msg['request-id'])
+        result = await self._recv(msg['request-id'])
+        log.debug('connection {} <- {}'.format(id(self), result))
```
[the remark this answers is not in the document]
rogpeppe: ditto

juju/client/connection.py
```diff
@@ -324,70 +364,50 @@ def https_connection(self):
...
-            self.password,
-            self.cacert,
-            self.macaroons,
-            self.loop,
-            self.max_frame_size,
+        endpoint, kwargs = self.connect_params()
+
+        return await Connection.connect(endpoint, **kwargs)
+
+    def connect_params(self):
```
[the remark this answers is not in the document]
rogpeppe: We also use it to call other methods such as Model._connect_direct, which wouldn't work with that change.
It can also be useful for debugging - I think it's probably reasonable as is.

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
-            macaroons=None, loop=None, max_frame_size=None):
+            await self._connect_with_login([(self.endpoint, self.cacert)])
+
+    async def _connect(self, endpoints):
+        if len(endpoints) == 0:
+            raise Exception('no endpoints to connect to')
+        first_exception = None
+        # TODO try connecting concurrently.
+        for endpoint, cacert in endpoints:
+            try:
```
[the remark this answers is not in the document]
rogpeppe: We don't care too much what goes wrong where - we want to catch any exception that happens in this block and try the next endpoint rather than bombing out, I think. In a future PR, we'll change this to do all the connections concurrently.

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
+                self.ws, self.addr = await self._open(endpoint, cacert)
+                self.endpoint = endpoint
+                self.cacert = cacert
+                self._receiver_task.start()
+                log.info("Driver connected to juju %s", self.addr)
+                self.monitor.close_called.clear()
+                return
+            except Exception as e:
```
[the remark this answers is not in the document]
rogpeppe: We actually don't care what the exception types are - just that we raise whatever was raised by the first one we found.

`except Exception as e` seems pretty common in Python - given that we don't know much about it other than it's an exception, I think a short variable name makes sense. I suppose we could do `except Exception as exception` but that seems unnecessarily verbose for a scope of 3 lines.

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
+        finally:
+            if not success:
+                await self.close()
 
-        """
-        jujudata = JujuData()
+    async def _connect_with_redirect(self, endpoints):
+        try:
+            login_result = await self._connect_with_login(endpoints)
+        except JujuAPIError as e:
```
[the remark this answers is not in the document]
rogpeppe: Out of interest, I did a quick survey of all the python code in my source directory for exception variable names, and got these results, covering ~96% of all occurrences:

    3196 e
     976 exc
     937 err
     150 pe
     123 ex
     114 error
      95 why
      74 v
      60 ve
      47 x
      36 te
      23 msg
 
Based on that, I think e is probably fine.

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
+    def _build_facades(self, facades):
         self.facades.clear()
         for facade in facades:
             self.facades[facade['name']] = facade['versions'][-1]
 
     async def login(self):
-        username = self.username
-        if username and not username.startswith('user-'):
-            username = 'user-{}'.format(username)
+        params = {}
```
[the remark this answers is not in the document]
rogpeppe: Yeah, *shrug*.

juju/client/_client.py
```diff
@@ -43,7 +43,13 @@ def from_connection(cls, connection):
         @param connection: initialized Connection object.
 
         """
-        version = connection.facades[cls.__name__[:-6]]
+        facade_name = cls.__name__
+        if not facade_name.endswith('Facade'):
+           raise TypeError('Unexpected class name: {}'.format(facade_name))
```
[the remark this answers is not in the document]
johnsca: `__name__` is guaranteed to be a `str`, so this wouldn't have any effect.

juju/client/connection.py
```diff
@@ -1,21 +1,22 @@
...
-
-import asyncio
 
-from juju import tag, utils
+from juju import utils
 from juju.client import client
-from juju.client.jujudata import JujuData
-from juju.errors import JujuError, JujuAPIError, JujuConnectionError
+from juju.errors import JujuError, JujuAPIError
+
```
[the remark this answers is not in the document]
johnsca: The tests for conjure-up include [isort](https://pypi.python.org/pypi/isort) to ensure consistent, standard import ordering (with an auto-format make target to make fixing them easy).  Probably worth including in libjuju as well.

juju/client/jujudata.py
rogpeppe: Looks like it.

tests/unit/test_connection.py
rogpeppe: tox excludes the test files (deliberate choice, I think). I'll run lint on the test files too.

juju/client/connection.py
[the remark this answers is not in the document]
johnsca: Data-driven decision making?  Nice!  :p

tests/unit/test_connection.py
rogpeppe: Unfortunately that's not an option for `with` arguments (it doesn't accept a tuple in that way).

juju/client/connector.py
johnsca: +1 to optional `name`

juju/client/jujudata.py
johnsca: pathlib is part of the standard library since 3.4.  Personally, I find it much easier to read and understand than `os.path.join(foo, bar, qux, ...)`, and the other methods that the OO syntax allows makes it a clear win to me.

tests/unit/test_connection.py
johnsca: I think the tests might be excluded.  No objection to including them, though.

juju/client/_client.py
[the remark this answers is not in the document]
rogpeppe: I think it does have an effect - it quotes the string. But given that we know the name is going to be a Python identifier, I'm not sure that's worth doing.

juju/client/_client.py
```diff
@@ -43,7 +43,13 @@ def from_connection(cls, connection):
...
+        facade_name = facade_name[:-len('Facade')]
+        version = connection.facades.get(facade_name)
+        if version is None:
+            raise Exception('No facade {} in facades {}'.format(facade_name, connection.facades))
 
         c = lookup_facade(cls.__name__, version)
```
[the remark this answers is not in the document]
rogpeppe: No, that's wrong - facade_name is the name of the facade, but we need to look up the facade by its class name here.

juju/client/connection.py
```diff
@@ -1,21 +1,22 @@
+import asyncio
+
```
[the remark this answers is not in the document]
rogpeppe: done

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
-
-    @classmethod
-    async def connect_model(cls, model, loop=None, max_frame_size=None):
-        """Connect to a model by name.
-
-        :param str model: [<controller>:]<model>
+        success = False
+        try:
+            await self._connect(endpoints)
+            for i in range(0, 4):
```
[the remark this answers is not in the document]
rogpeppe: done

juju/client/connection.py
```diff
@@ -396,147 +416,88 @@ def https_connection(self):
...
+                result = await self.login()
+                macaroonJSON = result.get('discharge-required')
+                if macaroonJSON is None:
+                    self.info = result['response']
+                    success = True
+                    return result
+                # TODO implement macaroon authentication
+                raise Exception('macaroon authentication not implemented yet')
```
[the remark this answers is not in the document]
rogpeppe: done

juju/client/connection.py
```diff
@@ -552,34 +513,24 @@ def build_facades(self, facades):
...
-        for value in base64_macaroons
-    ]
+                return await(self.task())
+            finally:
+                self.stopped.set()
+        self.stopped.clear()
+        self.loop.create_task(run())
+
+
+def macaroons_for_domain(cookies, domain):
```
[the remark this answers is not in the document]
rogpeppe: done, and unexported too.

juju/client/jujudata.py
rogpeppe: done (in separate PR)

setup.py
rogpeppe: done

examples/action.py
```diff
@@ -27,7 +27,7 @@
 
 async def main():
     model = Model()
-    await model.connect_current()
+    await model.connect()
```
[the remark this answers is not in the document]
rogpeppe: I think it's probably worth it, as it's arguably simpler and having one method for connection seems nice to me. It's arguable whether we should try to preserve some backward compatibility by continuing to implement a connect_current method that just calls connect. I'm not sure about that.

juju/client/connection.py
```diff
@@ -1,21 +1,22 @@
...
-from pathlib import Path
-
-import asyncio
 
-from juju import tag, utils
+from juju import utils
 from juju.client import client
-from juju.client.jujudata import JujuData
-from juju.errors import JujuError, JujuAPIError, JujuConnectionError
+from juju.errors import JujuError, JujuAPIError
```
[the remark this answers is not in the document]
rogpeppe: done

juju/client/connector.py
```diff
@@ -0,0 +1,121 @@
+import asyncio
```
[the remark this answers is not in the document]
rogpeppe: apparently not.

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: If you want to avoid the exception, you can call is_connected. I'm not entirely convinced it's worth it, but done anyway.

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: done

juju/client/connector.py
rogpeppe: Not sure. Left as is for consistency.

juju/client/connector.py
rogpeppe: done

juju/controller.py
rogpeppe: done

juju/unit.py
rogpeppe: done

tests/integration/test_controller.py
rogpeppe: done

tests/integration/test_machine.py
rogpeppe: done

tests/unit/test_connection.py
rogpeppe: Included in #188

tests/unit/test_connection.py
johnsca: [Looks like](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement) the multiple context expression version of `with` is 3.1 and greater.

johnsca: As Roger pointed out to me on IRC, even in Python 3, the parenthetical, no-backslash syntax is in fact rejected.

johnsca:
The test failure is a spurious timeout due to load on Travis which is unrelated to this PR.  (We should figure out a better way to handle that test, or at least increase the timeout.)

PR #193, opened 2017-12-11
make libjuju work with bakery authentication

rogpeppe:
This isn't complete - we're still not writing cookies acquired
during the authentication process back to the cookie jars,
but we'll add that in another PR.

The integration tests rely on the fact that an agent user
has been added to the staging identity manager.
Its key is included in the commit, because it doesn't
grant any rights that aren't easily acquired elsewhere.

We also move the Model.grant method to Controller.grant_model
because it made two extra connections to the controller, which is
not really we want it doing behind the user's back, and we add a Model.controller
method to make it easy to get a controller connection from the Model.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -9,13 +9,22 @@ from concurrent.futures import CancelledError
 from http.client import HTTPSConnection
 
 import macaroonbakery.httpbakery as httpbakery
+import macaroonbakery.bakery as bakery
 import websockets
 from juju import utils
 from juju.client import client
 from juju.errors import JujuAPIError, JujuError
 from juju.utils import IdQueue
 
-log = logging.getLogger("websocket")
+log = logging.getLogger('juju.client.connection')
+
+
+class _RedirectError(Exception):
+    """_RedirectError is the exception raised from Monitor.login when
+    the controller is redirecting us to another controller
+    :param endpoints list((str, str)): list of (host:port, cacert) pairs"""
+    def __init__(self, endpoints):
+        self.endpoints = endpoints
 
 
 class Monitor:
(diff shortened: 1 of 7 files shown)


johnsca:
```
juju.errors.JujuError: ['could not grant controller access: user already has "login" access or greater']
```

I guess that explains the revoke, but I guess it should have read the current rights and computed the diff.

tests/integration/test_macaroon_auth.py
```diff
@@ -0,0 +1,117 @@
...
+log = logging.getLogger(__name__)
+
+
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_macaroon_auth(event_loop):
+    try:
+       auth_info, username = agent_auth_info()
+    except NoAgentInfo:
+       return
```
johnsca: It would be better for these to skip using [`pytest.skip(reason)`](https://docs.pytest.org/en/latest/skipping.html) to make it clear that they didn't actually succeed.

tests/integration/test_macaroon_auth.py
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            async with Model(
+                jujudata=NoAccountsJujuData(m._connector.jujudata),
+                bakery_client=client,
+            ):
+                pass
+        except httpbakery.BakeryException:
+            # We're expecting this because we're using the
+            # wrong key.
+            pass
```
johnsca: This test can't actually fail.  I recommend adding an `else` clause to the `try` / `except` that calls `pytest.fail(reason)`.

tests/integration/test_macaroon_auth.py
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            async with Model(
+                jujudata=NoAccountsJujuData(m._connector.jujudata),
+                bakery_client=client,
+            ) as m1:
+                pass
+        except JujuAPIError:
+            # We're expecting this because we're using the
+            # wrong user name.
+            pass
```
johnsca: Same.  Add an `else` to fail. (You could also replace the inner `pass` but in this case I think the `else` clause is a bit more descriptive.  Up to your sense of style, though.)

juju/client/connection.py
```diff
@@ -9,13 +9,22 @@
...
 from juju import utils
 from juju.client import client
 from juju.errors import JujuAPIError, JujuError
 from juju.utils import IdQueue
 
-log = logging.getLogger("websocket")
+log = logging.getLogger('juju.client.connection')
+
+
+class _RedirectError(Exception):
```
johnsca: I know this is really only for internal use, but I'd still prefer to keep all exceptions together in `juju.errors`.  Also, it's not really an error.

johnsca:
Closing in favor of #206

PR #194, opened 2017-12-13
JujuData abstract base class

rogpeppe:
This means we won't have to mock the JujuData methods
in order to be able to connect with different credentials
(for example when connecting with macaroon authentication).

It would be nice if the base class method semantics weren't
tied so closely to the underlying data format, and instead
reflected the (very few) operations that we actually need,
which would make it easier to modify its behaviour, but
this should be OK to begin with.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -95,7 +95,7 @@ class Connection:
     @classmethod
     async def connect(
             cls,
-            endpoint,
+            endpoint=None,
             uuid=None,
             username=None,
             password=None,
@@ -124,6 +124,8 @@ class Connection:
         :param max_frame_size The maximum websocket frame size to allow.
         """
         self = cls()
+        if endpoint is None:
+            raise ValueError('no endpoint provided')
         self.uuid = uuid
         if bakery_client is None:
             bakery_client = httpbakery.Client()
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -3,7 +3,7 @@ import logging
 
 import macaroonbakery.httpbakery as httpbakery
 from juju.client.connection import Connection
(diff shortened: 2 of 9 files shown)


juju/client/connection.py
```diff
@@ -95,7 +95,7 @@ class Connection:
     @classmethod
     async def connect(
             cls,
-            endpoint,
+            endpoint=None,
```
johnsca: What's the point of this change if you then add run-time validation for it below?

If the concern is about passing an expanded dict for keyword args, you should still be able to do that with some positional params, as long as those are satisfied:

```python
In [1]: def foo(a, b=None):
   ...:     print(a, b)
   ...:     

In [2]: foo(**{'a': 'a', 'b': 'b'})
a b

In [3]: foo(**{'b': 'b'})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-c1546e6b8c27> in <module>()
----> 1 foo(**{'b': 'b'})

TypeError: foo() missing 1 required positional argument: 'a'
```

tests/unit/test_controller.py
```diff
@@ -0,0 +1,38 @@
+import asynctest
+
+
+class TestControllerConnect(asynctest.TestCase):
+    @asynctest.patch('juju.client.connector.Connector.connect_controller')
+    async def test_controller_connect_no_args(self, mock_connect_controller):
+        from juju.controller import Controller
```
johnsca: There isn't really any reason to continue this pattern here, but it's not a blocker.

juju/client/connection.py
rogpeppe: Yeah, I hadn't realised you could pass positional parameters in a **argument. That's done now.

Issue #172, opened 2017-10-09
Intermittent test failure: cannot add child handler


johnsca:
Below is due to parallelized tests running in threads running afoul of https://docs.python.org/3/library/asyncio-subprocess.html#subprocess-and-threads

Only comes up intermittently, but we should see if we can prevent it.
```
___________________________________ test_scp ___________________________________
[gw1] linux -- Python 3.5.3 /home/travis/build/juju/python-libjuju/.tox/py35/bin/python3
event_loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
    @base.bootstrapped
    @pytest.mark.asyncio
    async def test_scp(event_loop):
        async with base.CleanModel() as model:
            await model.add_machine()
            await asyncio.wait_for(
                model.block_until(lambda: model.machines),
                timeout=240)
            machine = model.machines['0']
            await asyncio.wait_for(
                model.block_until(lambda: (machine.status == 'running' and
                                           machine.agent_status == 'started')),
                timeout=480)
    
            with NamedTemporaryFile() as f:
                f.write(b'testcontents')
                f.flush()
>               await machine.scp_to(f.name, 'testfile')
tests/integration/test_machine.py:59: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
juju/machine.py:145: in scp_to
    await self._scp(source, destination, scp_opts)
juju/machine.py:176: in _scp
    process = await asyncio.create_subprocess_exec(*cmd, loop=loop)
/opt/python/3.5.3/lib/python3.5/asyncio/subprocess.py:212: in create_subprocess_exec
    stderr=stderr, **kwds)
/opt/python/3.5.3/lib/python3.5/asyncio/base_events.py:1189: in subprocess_exec
    bufsize, **kwargs)
/opt/python/3.5.3/lib/python3.5/asyncio/unix_events.py:194: in _make_subprocess_transport
    self._child_watcher_callback, transp)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <asyncio.unix_events.SafeChildWatcher object at 0x7f55f58f02b0>
pid = 15616
[...]

PR #204, opened 2018-02-09
Fix macaroon auth

### LINKED ISSUE #203: Can't connect to JAAS on master
#187 apparently wasn't tested against JAAS because there are a number of issues.  The proximate error is: https://paste.ubuntu.com/26546737/

In investigating this, however, there are some other issues currently being masked by this.

johnsca:
~~Depends on https://github.com/go-macaroon-bakery/py-macaroon-bakery/pull/46~~

Doesn't actually require that, though that fixes an edge case that users might hit.

This also has a drive-by possible fix for #172.

### DIFF
--- a/examples/connect_current_model.py
+++ b/examples/connect_current_model.py
@@ -1,10 +1,6 @@
 """
-This example:
-
-1. Creates a model on the current controller
-2. Deploys a charm to it.
-3. Attempts to ssh into the charm
-
+This is a very basic example that connects to the currently selected model
+and prints the number of applications deployed to it.
 """
 import logging
 
--- a/examples/connect_current_model.py
+++ b/examples/connect_current_model.py
@@ -6,11 +6,12 @@ This example:
 3. Attempts to ssh into the charm
 
 """
+import logging
+
 from juju import loop
 from juju.controller import Model
-from logging import getLogger
 
-LOG = getLogger(__name__)
+log = logging.getLogger(__name__)
 
 
 async def main():
@@ -25,4 +26,5 @@ async def main():
 
 
 if __name__ == '__main__':
+    logging.basicConfig(level=logging.INFO)
     loop.run(main())
--- a/examples/connect_current_model.py
+++ b/examples/connect_current_model.py
@@ -1,0 +1,28 @@
+"""
+This example:
+
+1. Creates a model on the current controller
+2. Deploys a charm to it.
+3. Attempts to ssh into the charm
+
+"""
+from juju import loop
+from juju.controller import Model
+from logging import getLogger
+
+LOG = getLogger(__name__)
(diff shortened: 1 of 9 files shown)


johnsca:
So much for hopefully fixing `test_scp`.  :/

examples/connect_current_model.py
```diff
@@ -0,0 +1,30 @@
+"""
+This example:
+
+1. Creates a model on the current controller
+2. Deploys a charm to it.
+3. Attempts to ssh into the charm
```
Cynerva: This docstring isn't accurate

juju/client/connection.py
```diff
@@ -454,32 +463,44 @@ def connect_params(self):
...
+                response = result['response']
+                macaroonJSON = response.get('discharge-required')
                 if macaroonJSON is None:
-                    self.info = result['response']
+                    self.info = response
                     success = True
                     return result
                 # TODO implement macaroon authentication
-                raise NotImplementedError('macaroon authentication not implemented yet')
+                error_msg = response['discharge-required-error']
```
Cynerva: Can we safely assume a response that contains `discharge-required` will always contain `discharge-required-error` as well? I.E. we won't hit a `KeyError` here?

juju/client/connector.py
```diff
@@ -104,12 +104,19 @@ def connection(self):
...
         # instead of just the first one.
         endpoint = controller['api-endpoints'][0]
-        models = self.jujudata.models()[controller_name]
-        account = self.jujudata.accounts()[controller_name]
+        accounts = self.jujudata.accounts()
+        if controller_name not in accounts:
+            raise JujuAuthError('Account not found for controller: {}'.format(
+                controller_name))
+        account = accounts[controller_name]
+        models = self.jujudata.models().get(controller_name, {})['models']
```
Cynerva: This looks a bit odd - if `self.jujudata.models()` doesn't container `controller_name`, then it'll default to `{}`, try to get `'models'` from it, and raise a `KeyError` yeah?

juju/client/connector.py
```diff
@@ -104,12 +104,19 @@ def connection(self):
...
+        if model_name not in models:
+            raise JujuConnectionError('Model found: {}'.format(model_name))
```
Cynerva: Model not found?

Cynerva:
Commented with a couple nits but nothing that really needs to be addressed. LGTM!

I can't merge PRs on this repo so someone else will have to do it.

juju/client/connection.py
johnsca: That's correct, from what @rogpeppe told me.

johnsca:
Closing in favor of #206

PR #206, opened 2018-02-16
Full macaroon bakery support

johnsca:
This combines and supersedes #193 and #204.  This should address the review comments on both and is working with my testing on JAAS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,13 @@ before_install:
   - sudo snap install juju-wait --classic || true
 install: pip install tox-travis
 env:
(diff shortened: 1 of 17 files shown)


johnsca:
One thing I'm not sure of is how the `test_macaroon_auth` tests are supposed to pass.  They don't see like they work on either localhost or JAAS.  @rogpeppe?

johnsca:
I manually tested this branch with a conjure-up snap built from https://github.com/conjure-up/conjure-up/pull/1334 and run on a pristine xenial container and it was able to add and connect to the model with the CLI being used only for initially registering the controller.

PR #208, opened 2018-02-20
Expose a machines series

gnuoy:
It is useful to be able to assert that the application currently
deployed in the model is running the operating system series
required. This allows that to be checked by adding a series property
to a Machine.


johnsca:
CI failures are transient network issues to the charm store.

PR #217, opened 2018-03-08
Always send auth-tag even with macaroon auth

### LINKED ISSUE #216: Self-hosted controllers require auth-tag even with macaroon auth
Originally reported by @mitechie via IRC

It seems that connecting to self-hosted controllers with macaroon auth fails with the following error:

```
  File "examples/connect_current_model.py", line 26, in <module>
    loop.run(main())
  File "/home/johnsca/juju/python-libjuju/juju/loop.py", line 38, in run
    raise task.exception()
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "examples/connect_current_model.py", line 16, in main
    await model.connect()
  File "/home/johnsca/juju/python-libjuju/juju/model.py", line 507, in connect
    await self._connector.connect_model(model_name, **kwargs)
  File "/home/johnsca/juju/python-libjuju/juju/client/connector.py", line 140, in connect_model
    bakery_client=self.bakery_client_for_controller(controller_name),
  File "/home/johnsca/juju/python-libjuju/juju/client/connector.py", line 68, in connect
  File "/home/johnsca/juju/python-libjuju/juju/client/connection.py", line 168, in connect
    await self._connect_with_redirect([(endpoint, cacert)])
  File "/home/johnsca/juju/python-libjuju/juju/client/connection.py", line 511, in _connect_with_redirect
    login_result = await self._connect_with_login(endpoints)
  File "/home/johnsca/juju/python-libjuju/juju/client/connection.py", line 483, in _connect_with_login
  File "/home/johnsca/juju/python-libjuju/juju/client/connection.py", line 538, in login
    "params": params,
  File "/home/johnsca/juju/python-libjuju/juju/client/connection.py", line 316, in rpc
juju.errors.JujuAPIError: no credentials provided
```

We confirmed that the macaroon is indeed being sent, and after some additional testing I discovered that the `auth-tag` is required, even though that info is encoded into the macaroon.  As it doesn't hurt anything to send it anyway, we should just unconditionally supply it.

johnsca:
Fixes #216

### DIFF
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -38,10 +38,7 @@ Quickstart
 ----------
 Here's a simple example that shows basic usage of the library. The example
 connects to the currently active Juju model, deploys a single unit of the
-ubuntu charm, then exits.
-
-More examples can be found in the `examples/` directory of the source tree,
-and in the documentation.
+ubuntu charm, then exits:
 
 
 .. code:: python
@@ -95,3 +92,12 @@ and in the documentation.
 
   if __name__ == '__main__':
       main()
+
+
+More examples can be found in the docs, as well as in the ``examples/``
+directory of the source tree which can be run using ``tox``.  For
+example, to run ``examples/connect_current_model.py``, use:
+
+.. code:: bash
+
+  tox -e example -- examples/connect_current_model.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   - sudo lxd init --auto || true
   - sudo addgroup lxd || true
   - sudo usermod -a -G lxd $USER || true
-  - sudo -E sudo -u $USER -E bash -c 'lxc network create lxdbr0 ipv4.nat=true ipv6.address=none ipv6.nat=false'
+  - sudo -E sudo -u $USER -E bash -c 'lxc config set core.https_address "[::]"'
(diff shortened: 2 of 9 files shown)


johnsca:
To test:

```
juju bootstrap localhost
juju change-user-password
.tox/py3/bin/python examples/connect_current_model.py
```

I'd like to include this in the integration tests, but it would affect every single test, so I don't think it's a good idea.

docs/readme.rst
```diff
@@ -95,3 +92,12 @@ and in the documentation.
 
   if __name__ == '__main__':
       main()
+
+
+More examples can be found in the docs, as well as in the ``examples/``
+directory of the source tree which can be run with using ``tox``.  For
```
tvansteenburgh: s/with//

johnsca:
The tests don't do the controller bootstrapping, and depend on one already being bootstrapped, so I'm not sure how we could mark one test to be run differently.  I guess a separate tox env?

The same test passed with Juju edge on on the push run.  This is one of the inconsistent failures we sometimes see on Travis (others being `test_scp` failing / timing out, package installation failing, bootstrap failing, and store timeouts).  The error is from LXD, and it seems as if the ZFS kernel module isn't enabled sometimes on Travis; I'm not sure if we could do something to ensure that it is, but it makes me wonder if there are different images that we end up running under on Travis sometimes.  Restarting the job will probably make it pass.

johnsca:
It looks like adding the `password: ` line back to `accounts.yaml` is enough to get it back to working, but it would still affect other tests that are being run in parallel.  Is there any way to indicate that a given test must be run on its own?

PR #220, opened 2018-03-23
Explicitly close event loop

### LINKED ISSUE #219: unit.scp_to results in "TypeError: signal handler..." exception
Python 3.5.2
libjuju 0.7.3

Running unit.scp_to successfully scp's a file to the unit in question but throws the following stacktrace:
 
```
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

I think this is because loop needs an explicit close in later versions of Python. I am some what out of my depth here so I may well be wrong :-)

I can reproduce the error with:

```
#!/usr/bin/env python3

from juju import loop
from juju.model import Model

async def scp_to_unit():
    model = Model()
    await model.connect_current()
    try:
        unit = model.applications['vault'].units[0]
        await unit.scp_to('/tmp/tmb.txt', '/tmp/tmb.txt')
    finally:
        # Disconnect from the api server and cleanup.
        await model.disconnect()

loop.run(scp_to_unit())
```

gnuoy:
Explicitly close the event loop created in juju.loop. I think this
may have been optional in earlier version of Python but I don't have
any science to support that. The Python docs show examples of an
explicit close in all examples here:
https://docs.python.org/3/library/asyncio-eventloop.html#event-loop-examples
This closes #219

### DIFF
--- a/juju/loop.py
+++ b/juju/loop.py
@@ -40,3 +40,4 @@ def run(*steps):
     finally:
         if added:
             loop.remove_signal_handler(signal.SIGINT)
+        loop.close()


gnuoy:
I'm closing this pull request because I'm not sure whether closing the thread may have unintended consequences.

johnsca:
I think this is the right solution in one sense, because I think what's happening is that the handler gets garbage collected before the event loop, so explicitly closing prior to GC should prevent the error.  Unfortunately, closing it in this helper would cause it to break if called twice, or if anything else tries to use the global event loop afterward.  (It also won't help that damned `test_scp` failure, because the loop that the test uses is managed by (IIRC) the [pytest-asyncio](https://pypi.python.org/pypi/pytest-asyncio) plugin and there isn't a place to hook in to close it properly.)

Maybe the best solution would be to add a `close` wrapper to `juju.loop` and document that it should be used (and for the test, investigate the upstream project to get the loop properly closed in that).

Issue #221, opened 2018-04-03
Local charms fail on JaaS


johnsca:
When connected to a JaaS model, [uploading local charms](https://github.com/juju/python-libjuju/blob/master/juju/model.py#L578) is failing due to macaroon auth issues.  Currently, it throws a `BrokenPipeError` because the macaroon [is not included in the headers](https://github.com/juju/python-libjuju/blob/master/juju/client/connection.py#L345-L346), but even with that resolved, it gets a discharge required error with no redirect info.

juanmanuel-tirado:
This issue has expired.

PR #222, opened 2018-04-03
[wip] Fix uploading local charms to JaaS

johnsca:
In-progress fix for #221

Still getting macaroon discharge required error and I'm uncertain how to get the redirect info.  I expected the response headers to include it, but it's a 401 with only the following headers:

```
Content-Length: 1062
Content-Type: application/json
Www-Authenticate: Basic realm="juju"'
Date: Tue, 03 Apr 2018 14:03:34 GMT
```

And this response body:

```python
{'error': 'cannot upload charm: verification failed: failed to decrypt caveat 1 signature: decryption failure',
 'error-code': 'macaroon discharge required',
 'error-info': {'macaroon': { ... }}}
```


johnsca:
It may be helpful to do the deploy via the CLI with `--logging-config TRACE` to track the requests it makes.

PR #231, opened 2018-05-16
Inline jsonfile credential when sending to controller

### LINKED ISSUE #230: Invalid creds are sent to gce controllers
When adding a model on a GCE controller, the `UploadCredentials` call sends the name of the file rather than the contents, which results in a parse error on the subsequent `AddModel` call.

johnsca:
Fixes #230

### DIFF
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from pathlib import Path
 
 from . import errors, tag, utils
 from .client import client, connector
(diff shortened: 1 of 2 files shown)


johnsca:
Test failures are unrelated to this PR.  The failure on stable is fixed in #217 and the failure on edge is an issue with Juju edge bootstrapping on Travis that I'm not sure about yet.

I'm going to merge this anyway.

PR #233, opened 2018-05-17
Surface errors from bundle plan

### LINKED ISSUE #232: Errors from GetBundleChanges are ignored
An error in the placement directives in a bundle led to the following response:

```python
{
  'request-id': 7,
  'response': {
    'errors': ['too many units specified in unit placement for application "kubernetes-worker"']
  }
}
```

This didn't raise a `JujuAPIError` and wasn't handled by either `BundleHandler.fetch_plan` or `BundleHandler.execute_plan`, and thus was ignored, being treated as an empty bundle plan.

johnsca:
Fixes #232

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1774,6 +1774,9 @@ class BundleHandler:
         self.plan = await self.client_facade.GetBundleChanges(
             yaml.dump(self.bundle))
 
+        if self.plan.errors:
+            raise JujuAPIError(self.plan.errors)
+
     async def execute_plan(self):
         for step in self.plan.changes:
             method = getattr(self, step.method)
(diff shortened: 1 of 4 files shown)


juju/model.py
```diff
@@ -1774,6 +1774,9 @@ def __init__(self, model):
         self.plan = await self.client_facade.GetBundleChanges(
             yaml.dump(self.bundle))
 
+        if self.plan.errors:
+            raise JujuError(self.plan.errors)
+
     async def execute_plan(self):
```
kwmonroe: I don't see any check for `self.plan.errors` here in `execute_plan`.  For now, are you confident that this is not an issue because `deploy` will always fetch (and error) before we execute?  Or would it be better to check for plan errors during both fetch and execute?

johnsca: Yes, the only way that the plan is set is via `fetch_plan`, so there's no need to additionally check for errors on execute.

PR #234, opened 2018-06-01
Add support for Application get_actions

### LINKED ISSUE #226: Implement juju.application.Application.get_actions
Pretty much as the issue description. It would be nice to be able to use get_actions.

gnuoy:
Currently juju.application.get_actions raises an NotImplementedError
exception. This pr implements it. Oddly, ApplicationsCharmsActions
takes a list of application tags so it returns a list containing
one element. get_actions double checks that the tags match before
returning the actions. If a charm has no actions
ApplicationsCharmsActions still returns an
ApplicationsCharmsActionsResults object but the actions attribute
is None. 

The test_application functional tests have also been switched to
use cs:xenial/percona-cluster rather than cs:trusty/mysql as mysql
has no actions defined.

This PR also contains two small unrelated fixes.

* Remove unused Model import in tests/integration/test_controller.py
* gitignore pytestdebug.log

closes #226

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -232,19 +232,20 @@ class Application(model.ModelEntity):
         """Get actions defined for this application.
 
         :param bool schema: Return the full action schema
-
+        :return dict: The charms actions, empty dict if none are defined.
         """
-        app_tag = "application-{}".format(self.name)
-        entity = [{"tag": app_tag}]
+        actions = {}
+        entity = [{"tag": self.tag}]
         action_facade = client.ActionFacade.from_connection(self.connection)
         results = (
             await action_facade.ApplicationsCharmsActions(entity)).results
-        # There should only be one element in results so this is probably OTT
         for result in results:
-            if result.application_tag == app_tag:
-                return result.actions
-        else:
-            return None
+            if result.application_tag == self.tag and result.actions:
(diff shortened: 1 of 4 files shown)


johnsca:
CI failures are the spurious zfs issue that I'm at a loss as how to fix.  I already tried to make the default storage pool for lxd be dir instead of zfs, but I guess I'm not doing it correctly.  If you have any ideas on how to fix that, it would be awesome, but those failures shouldn't block this PR.

juju/application.py
```diff
@@ -228,13 +228,23 @@ def collect_metrics(self):
...
-        raise NotImplementedError()
+        app_tag = "application-{}".format(self.name)
+        entity = [{"tag": app_tag}]
+        action_facade = client.ActionFacade.from_connection(self.connection)
+        results = (
+            await action_facade.ApplicationsCharmsActions(entity)).results
+        # There should only be one element in results so this is probably OTT
+        for result in results:
+            if result.application_tag == app_tag:
+                return result.actions
```
johnsca: It's probably worth normalizing the return type here.  I'd like to see the object layer results be at least somewhat insulated from changes in the Go code, though this is pretty unlikely to change.  At the very least, the doc string should document the return structure.

Issue #106, opened 2017-04-10
Manual provisioning of existing machines does not work


ThomasBri:
Trying to add an existing machine to a model via ssh using the model.add_machine method like this:

`await model.add_machine(spec='ssh:ubuntu@{}'.format(ip_address))`

raises the following error:

ValueError: ('Error adding machine: %s', 'invalid model name "ssh"')

tvansteenburgh:
Here's the code that was used:
```
async def add_machine(ip, model):
    ssh_ubuntu__format = 'ssh:ubuntu@{}'.format(ip)
    m = await model.add_machine(spec=ssh_ubuntu__format)
    return m
 
 
async def test():
    model = Model()
    await model.connect('xx.xx.xx.xx:17070', 'd06c23f7-aacf-44c8-8819-166b9baf3860', 'xxxx', 'xxxxx')
 
    print('Start adding machine')
    print(await add_machine('xx.xx.xx.xx', model))
 
    print('Finished')
    await model.disconnect()
```

And here's the log:
```
DEBUG:websockets.protocol:client >> Frame(fin=True, opcode=1, data=b'{\n  "request-id": 4,\n  "request": "AddMachines",\n  "params": {\n    "params": [\n      {\n        "placement": {\n          "scope": "ssh",\n          "directive": "ubuntu@192.168.77.242"\n        },\n        "parent-id": null,\n        "addresses": [],\n        "disks": [],\n        "jobs": [\n          "JobHostUnits"\n        ],\n        "hardware-characteristics": null,\n        "nonce": null,\n        "container-type": null,\n        "constraints": null,\n        "series": null,\n        "instance-id": null\n      }\n    ]\n  },\n  "version": 1,\n  "type": "Client"\n}')
WARNING:asyncio:Executing <Task pending coro=<test() running at /opt/openbaton/jujunew/test/new.py:178> wait_for=<Future pending cb=[Task._wakeup()] created at /usr/lib/python3.5/asyncio/base_events.py:252> cb=[_run_until_complete_cb() at /usr/lib/python3.5/asyncio/base_events.py:164] created at /usr/lib/python3.5/asyncio/base_events.py:367> took 3.741 seconds
DEBUG:asyncio:poll took 0.051 ms: 1 events
[...]

PR #240, opened 2018-06-06
Add support for adding a manual (ssh) machine

AdamIsrael:
As discussed in issue #106, the model.add_machine method does not work when trying to add a manually instantiated machine with ssh credentials.

The reason for this is that the code that sets up and provisions the Juju agent on the target machine executes from the machine where the controller is bootstrapped, via the Juju CLI, and not the API. Part of this approach is because the ssh keys that the Juju controller uses to perform these operations only exist, by default, on the machine -- in `~/.local/share/juju/ssh` -- and not on the controller itself.

This PR re-implements a Python version of the process taken by the Juju code in `https://github.com/juju/juju/blob/develop/environs/manual/sshprovisioner/sshprovisioner.go`.

By necessity, I've changed the parameter required to add a manual machine, as it needs to know where the ssh keys are located. In cases where libjuju is being run from a machine different than the bootstrap host, the Juju keypair would need to be copied to it, or a new keypair made available and injected into the manually provisioned machine to enlist in Juju.


juju/provisioner.py
```diff
@@ -0,0 +1,364 @@
+from .client import client
+
+import paramiko
```
tvansteenburgh: Need to add paramiko as a dep?

AdamIsrael: Added!

tvansteenburgh:
It would be great to add a test for this, using a lxc as the machine to add.

juju/provisioner.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+
+        if isinstance(cmd, str):
+            cmd = shlex.split(cmd)
+
+        if type(cmd) is not list:
+            cmd = [cmd]
+
+        cmds = ' '.join(cmd)
+        stdin, stdout, stderr = ssh.exec_command(cmds, get_pty=pty)
```
johnsca: Not worth holding this PR over, but worth looking in to [asyncssh](https://github.com/ronf/asyncssh#asyncssh-asynchronous-ssh-for-python) to do the SSH work asynchronously.

PR #242, opened 2018-06-08
Implement model.get_action_output

gnuoy:


juju/model.py
```diff
@@ -1536,14 +1536,25 @@ def set_constraints(self, constraints):
...
+        :raises: :class:`JujuError` if invalid action_uuid
         """
-        raise NotImplementedError()
+        action_facade = client.ActionFacade.from_connection(
+            self.connection()
+        )
+        entity = [{'tag': tag.action(action_uuid)}]
+        action_output = await action_facade.Actions(entity)
+        # ActionResult.output is None if the action produced no output
+        if action_output.results[0].output:
```
ajkavanagh: If it is likely to be None, then the check should be `is None` and not just false-y.

johnsca:
CI is failing with a "no output in last 10 minutes" timeout.  This could possibly be due to the lack of a timeout in the wait for action.  It should be fairly easy to implement the `wait` param using [`asyncio.wait_for`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) which would at least let the test fail with a proper timeout error if that's the case.  I'll also try re-running the jobs to see if it was just a load issue on Travis.

johnsca:
The no output timeout appears to be consistent, so I think it's pretty likely due to the unbounded wait (and, of course, whatever underlying issue is causing the wait to not return).

PR #243, opened 2018-06-08
Fix JSON serialization error for bundle with lxd to unit placement

johnsca:
When a bundle contains lxd placement to a unit name, the full unit object model instance was ending up getting passed directly to the API. This change handles that case by resolving it to the proper machine
number instead.

This also has a fix for the workaround version check sometimes failing if triggered before Model.get_status() had been called.

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -14,7 +14,17 @@ log = logging.getLogger(__name__)
 class Machine(model.ModelEntity):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        if self.model.info.agent_version < client.Number.from_json('2.2.3'):
+        self.model.loop.create_task(self._queue_workarounds())
+
+    async def _queue_workarounds(self):
+        model = self.model
+        if not model.info:
+            await utils.run_with_interrupt(model.get_info(),
+                                           model._watch_stopping,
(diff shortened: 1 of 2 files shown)


johnsca:
It also contains a couple of trivial lint fixes.

johnsca:
Test failure is unrelated to this change:

```
  File "/home/travis/build/juju/python-libjuju/tests/unit/test_model.py", line 103, in test_get_series
    model = Model()
  File "/home/travis/build/juju/python-libjuju/juju/model.py", line 419, in __init__
    jujudata=jujudata,
  File "/home/travis/build/juju/python-libjuju/juju/client/connector.py", line 33, in __init__
    self.loop = loop or asyncio.get_event_loop()
  File "/opt/python/3.6.3/lib/python3.6/asyncio/events.py", line 676, in get_event_loop
    return get_event_loop_policy().get_event_loop()
  File "/opt/python/3.6.3/lib/python3.6/asyncio/events.py", line 584, in get_event_loop
    % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'Dummy-1'.
```

That's caused by the asyncio code being run in another thread, but I'm not sure how to prevent it.  There shouldn't be any threads running, and the event loop is managed by the test harness.  Suggestions welcome.

PR #253, opened 2018-07-12
Implement set/get model constraints

gnuoy:
Add methods for setting and getting model constraints.


gnuoy:
I think the ci failures are unrelated to this change

johnsca:
Yeah, I think the failures were due to an upstream issue with Juju edge which is resolved now.  Re-running just to be sure, but +1 and will go ahead and merge.

PR #260, opened 2018-08-14
Reorder scp parameters (#259)

### LINKED ISSUE #259: `scp_from` raises JujuError("command failed: %s" % cmd)
When using the `scp_from` function with `scp_opts` like in the zaza framework:
```python
await unit.scp_from(source, destination, user=user, proxy=proxy,  scp_opts=scp_opts)
```
Raises a `JujuError` exception:
```
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/zaza/charm_tests/ceph/tests.py", line 296, in test_ceph_encryption
    scp_opts='-r -p')
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 44, in _wrapper
    return run(_run_it())
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 32, in run
    return task.result()
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 43, in _run_it
    return await f(*args, **kwargs)
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/zaza/model.py", line 226, in async_scp_from_unit
    scp_opts=scp_opts)
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/juju/unit.py", line 205, in scp_from
    proxy=proxy, scp_opts=scp_opts)
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/juju/machine.py", line 173, in scp_from
    await self._scp(source, destination, scp_opts)
  File "/home/ubuntu/ceph-osd/.tox/func-smoke/lib/python3.6/site-packages/juju/machine.py", line 192, in _scp
    raise JujuError("command failed: %s" % cmd)
juju.errors.JujuError: command failed: ['scp', '-i', '/home/ubuntu/.local/share/juju/ssh/juju_id_rsa', '-o', 'StrictHostKeyChecking=no', '-q', '-B', 'ubuntu@10.5.0.11:/etc/ceph/dmcrypt-keys', '/tmp/tmp965enr1k', '-r', '-p']
```
The error (while using the scp command directly) is:
```
scp -i /home/ubuntu/.local/share/juju/ssh/juju_id_rsa -o StrictHostKeyChecking=no -q -B ubuntu@10.5.0.11:/etc/ceph/dmcrypt-keys /tmp/tmp965enr1k -r -p
-p: No such file or directory
```
It seems that scp doesn't like to have the extra parameters where it expects the `source` and `destination`
https://github.com/juju/python-libjuju/blob/26c86c81f763acae352478ff77bdc97cd5469299/juju/machine.py#L175-L192
Could be modified into something like:
```python
cmd = ['scp']
cmd += scp_opts.split()
cmd.extend([
            '-i', os.path.expanduser('~/.local/share/juju/ssh/juju_id_rsa'),
            '-o', 'StrictHostKeyChecking=no',
            '-q',
            '-B',
            source, destination
        ])
```

n-pochet:
Fixes #259

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -146,7 +146,8 @@ class Machine(model.ModelEntity):
         :param str destination: Remote destination of transferred files
         :param str user: Remote username
         :param bool proxy: Proxy through the Juju API server
-        :param str scp_opts: Additional options to the `scp` command
+        :param scp_opts: Additional options to the `scp` command
+        :type scp_opts: str or list
         """
         if proxy:
             raise NotImplementedError('proxy option is not implemented')
@@ -163,7 +164,8 @@ class Machine(model.ModelEntity):
         :param str destination: Local destination of transferred files
         :param str user: Remote username
         :param bool proxy: Proxy through the Juju API server
-        :param str scp_opts: Additional options to the `scp` command
+        :param scp_opts: Additional options to the `scp` command
+        :type scp_opts: str or list
         """
         if proxy:
             raise NotImplementedError('proxy option is not implemented')
@@ -176,15 +178,15 @@ class Machine(model.ModelEntity):
         """ Execute an scp command. Requires a fully qualified source and
         destination.
         """
-        cmd = ['scp']
-        cmd += scp_opts.split()
-        cmd.extend([
+        cmd = [
+            'scp',
             '-i', os.path.expanduser('~/.local/share/juju/ssh/juju_id_rsa'),
             '-o', 'StrictHostKeyChecking=no',
             '-q',
-            '-B',
-            source, destination
-        ])
+            '-B'
+        ]
+        cmd.extend(scp_opts.split() if isinstance(scp_opts, str) else scp_opts)
+        cmd.extend([source, destination])
         loop = self.model.loop
         process = await asyncio.create_subprocess_exec(*cmd, loop=loop)
         await process.wait()
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -176,15 +176,15 @@ class Machine(model.ModelEntity):
         """ Execute an scp command. Requires a fully qualified source and
         destination.
         """
-        cmd = [
-            'scp',
+        cmd = ['scp']
+        cmd += scp_opts.split()
+        cmd.extend([
             '-i', os.path.expanduser('~/.local/share/juju/ssh/juju_id_rsa'),
             '-o', 'StrictHostKeyChecking=no',
             '-q',
             '-B',
             source, destination
-        ]
-        cmd += scp_opts.split()
+        ])
         loop = self.model.loop
         process = await asyncio.create_subprocess_exec(*cmd, loop=loop)
         await process.wait()
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ dist/
 dev/
 .pytest_cache
 pytestdebug.log
+.vscode/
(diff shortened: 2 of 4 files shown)


juju/machine.py
```diff
@@ -176,15 +176,15 @@ def run(self, command, timeout=None):
         """ Execute an scp command. Requires a fully qualified source and
         destination.
         """
-        cmd = [
-            'scp',
+        cmd = ['scp']
+        cmd += scp_opts.split()
```
n-pochet: @ryan-beisner 
In this specific case, `+=` is not used with strings directly but with a list of strings `scp_opts.split()`
So, it would result in something like:
```python
>>> cmd = ['scp']
>>> scp_opts = '-p -r'
>>> cmd += scp_opts.split()
>>> cmd
['scp', '-p', '-r']
```
Which is what is expected.
Nevertheless, I could use `cmd.extend(scp_opts.split())` it you prefer

johnsca:
Test failure is:

```
juju.errors.JujuAPIError: cannot retrieve charm "cs:ubuntu-0": cannot get archive: Get https://api.jujucharms.com/charmstore/v5/ubuntu-0/archive?channel=stable: dial tcp: lookup api.jujucharms.com on 10.0.125.1:53: read udp 10.0.125.81:55961->10.0.125.1:53: read: connection refused
```

This only happened on one out of four runs and seems like a transient network issue, so I restarted that run.  Regardless, it's unrelated to this change.

juju/machine.py
johnsca: I really don't like the existing interface (`scp_opts` as a `str`), since it means you can't pass in any option values that actually contain a space.  Do you think you could wrap the split in an `if isinstance(scp_opts, str)` and update the docs to indicate that it supports a list or string?

juju/machine.py
```diff
@@ -176,15 +176,15 @@ def run(self, command, timeout=None):
...
+        cmd.extend([
             '-i', os.path.expanduser('~/.local/share/juju/ssh/juju_id_rsa'),
             '-o', 'StrictHostKeyChecking=no',
             '-q',
             '-B',
             source, destination
-        ]
-        cmd += scp_opts.split()
+        ])
```
johnsca: I think the original intention of add the `scp_opts` at the end was to allow user-supplied options to override the defaults.  I'm not entirely certain if that would work as expected, but maybe it would be best to inject the user options between the defaults and `source` / `destination` instead of either at the beginning or end.  Something like:

```python
        cmd = [
            '-i', os.path.expanduser('~/.local/share/juju/ssh/juju_id_rsa'),
            '-o', 'StrictHostKeyChecking=no',
            '-q',
            '-B',
        ]
        cmd.extend(scp_opts.split() if isinstance(scp_opts, str) else scp_opts)
        cmd.extend([source, destination])
```

n-pochet:
@johnsca 
Many tests seem to have failed because of a slowness/timeout of https://api.jujucharms.com
Could you please restart the tests?

n-pochet:
@johnsca 
Thanks for restarting the tests.
This time, it seems to be a DNS problem for python3.7-dev/STABLE.
For python3.6/EDGE, I don't know but it is not impacting the scp tests.
Do I need to adapt something else?

johnsca:
The tests are passing save for transient network connectivity errors, so I'm just going to go ahead and merge.

n-pochet:
Thanks!

Issue #261, opened 2018-09-12
Increase charm store timeout from 5s


eggyal:
**python-libjuju** currently sets a 5s timeout for calls to the charm store:

https://github.com/juju/python-libjuju/blob/514e47986c5b7875e9143ff4c46a212afacebd60/juju/model.py#L2165

Of course, failed attempts to access the store are retried after a 1s delay (up to three attempts in total):

https://github.com/juju/python-libjuju/blob/514e47986c5b7875e9143ff4c46a212afacebd60/juju/model.py#L2179-L2185

Unfortunately, if the store is responding slowly, all three attempts can easily fail (causing a deployment to error and abort).  A longer timeout would reduce the risk of such an annoying happenstance: I suggest 15s?

johnsca:
Fixed in #262

PR #262, opened 2018-09-17
Increase the timeout for charmstore connections

gnuoy:
The charm store often does not respond in 5s. Increase the timeout
to 20s. Charm store issue being tracked in https://github.com/CanonicalLtd/jujucharms.com/issues/551
. Closes issue #261


hatched:
While increasing the timeout should resolve the immediate issue. Going forward it would be better if the timeout was configurable.

PR #274, opened 2018-12-10
Updates for new Juju version

### LINKED ISSUE #255: GetBundleChanges is deprecated; should use BundleFacade.GetChanges
The newer facade is [already available in `_client1.py`](https://github.com/juju/python-libjuju/blob/master/juju/client/_client1.py#L460-L470), so I'm not sure why we aren't using it.

### LINKED ISSUE #272: Bundle deploy fails with Juju 2.5-beta
Originally reported at https://github.com/conjure-up/conjure-up/issues/1537

Apparently, this is due to the addition of support for bundles for Kubernetes models.

Some additional info from @wallyworld:

https://discourse.jujucharms.com/t/bundles-now-supported/365/2

There are 2 new AddApplication GUI args appended to the end of the current list:

- scale/num_units (int)
- placement (string)

These are used with the ApplicationDeploy params. For IAAS bundles we always
send numUnits=0 when making the deploy api call. For k8s, we need to pass in
that number.

The placement string is used to construct a single placement directive struct:

https://github.com/juju/juju/blob/develop/cmd/juju/application/bundle.go#L622

There's also a new ScaleApplication change event created when the user
re-deploys a k8s bundle with a scale change. It calls the ScaleApplication() API.

https://github.com/juju/juju/blob/develop/cmd/juju/application/bundle.go#L668

johnsca:
Update schema and bundle handling to account for changes in latest (2.5-rc1) version of Juju, as well as fixing some lint errors caused by a new release of flake8.

Fixes #255 
Fixes #272


.travis.yml
```diff
@@ -34,19 +33,17 @@ matrix:
...
+  - sudo lxd waitready --timeout 30
+  - sudo chmod 666 /var/snap/lxd/common/lxd/unix.socket
+  - lxd init --auto --network-address='[::]' --network-port=8443 --storage-backend=dir
 
-  - sudo addgroup lxd || true
-  - sudo usermod -a -G lxd $USER || true
-  # Trigger generation of the client certificates
-  - sg lxd -c "echo y|lxc remote add "$USER" https://0.0.0.0:8443 --accept-certificate --password='${LXD_PASSWORD}'"
+  # Horrible workaround to LP Bug #1738614
+  - sudo mkdir /var/snap/lxd/common/lxd/storage-pools/juju-zfs
```
kwmonroe: Need a `-p` here, or are we sure `./lxd/storage-pools` exists?

.travis.yml
```diff
@@ -34,19 +33,17 @@ matrix:
...
+  - lxc storage create juju-zfs dir source=/var/snap/lxd/common/lxd/storage-pools/juju-zfs
 
 script:
-  - sudo -E sudo -u $USER -E juju bootstrap localhost test --config 'identity-url=https://api.staging.jujucharms.com/identity' --config 'allow-model-access=true'
+  # dual sudos are to ensure lxd group is available
```
kwmonroe: invalid comment?  there are no sudos here anymore.  are we sure the user is in the lxd group now?

juju/provisioner.py
```diff
@@ -12,12 +12,12 @@
...
-    [re.compile("(arm$)|(armv.*)"), "armhf"],
-    [re.compile("aarch64"), "arm64"],
-    [re.compile("ppc64|ppc64el|ppc64le"), "ppc64el"],
-    [re.compile("ppc64|ppc64el|ppc64le"), "s390x"],
+    [re.compile(r"amd64|x86_64"), "amd64"],
+    [re.compile(r"i?[3-9]86"), "i386"],
+    [re.compile(r"(arm$)|(armv.*)"), "armhf"],
+    [re.compile(r"aarch64"), "arm64"],
+    [re.compile(r"ppc64|ppc64el|ppc64le"), "ppc64el"],
+    [re.compile(r"ppc64|ppc64el|ppc64le"), "s390x"],
```
kwmonroe: I know this was in the code previously (sans 'r'), but why s390 == ppc?  Surely the regex should have been `'s390|s390x'`.

johnsca: That is a good question.  I have no idea.  I'm not sure what the correct values should be.

.travis.yml
johnsca: Yeah, it's created by the snap, and specifically the `lxd init` on the previous line.

PR #278, opened 2018-12-14
Implement App.get_resources and pinned resources in bundles

### LINKED ISSUE #273: when deploying a bundle, resource revisions are ignored
Deploy a bundle that includes something like this:

```
services:
  flannel:
    charm: cs:~containers/flannel-339
    resources:
      flannel-amd64: 3
      flannel-arm64: 1
      flannel-s390x: 3
```

Instead of getting the specified bundle revisions, you'll get whichever revisions the charm store gives you (in this case, the latest resources from the edge channel). In contrast, if you deploy via Juju CLI, you will get the specified revisions.

The issue occurs [here](https://github.com/juju/python-libjuju/blob/87d9388191edfcfea804a563125e5634bba53db9/juju/model.py#L2053), where the `resources` that are passed in _do_ include the revision numbers, but that parameter is ignored.

johnsca:
Fixes #273

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -272,14 +272,24 @@ class Application(model.ModelEntity):
             actions = {k: v['description'] for k, v in actions.items()}
         return actions
 
-    def get_resources(self, details=False):
+    async def get_resources(self):
         """Return resources for this application.
 
-        :param bool details: Include detailed info about resources used by each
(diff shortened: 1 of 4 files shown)


Cynerva:
LGTM :+1:

Issue #280, opened 2019-01-14
Juju 2.5-rc1 doesn't work as CloudFacade has no attribute 'UpdateCredentials'


ajkavanagh:
It looks like (according to my small investigation) that UpdateCredentials has been renamed/changed to `UpdateCredentialsCheckModels`

```
$ juju version
2.5-rc1-bionic-amd64

$ functest-run-suite --keep-model -b trusty-icehouse 
Traceback (most recent call last):
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/bin/functest-run-suite", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 129, in main
    bundle=args.bundle)
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 55, in func_test_runner
    prepare.prepare(model_name)
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/charm_lifecycle/prepare.py", line 93, in prepare
    zaza.controller.add_model(model_name, config=get_model_settings())
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 44, in _wrapper
    return run(_run_it())
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 32, in run
    return task.result()
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/__init__.py", line 43, in _run_it
    return await f(*args, **kwargs)
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/zaza/controller.py", line 34, in async_add_model
    model = await controller.add_model(model_name, config=config)
  File "/home/ubuntu/charms/non-reactive/tox-ceph-osd/func-smoke/lib/python3.6/site-packages/juju/controller.py", line 256, in add_model
    owner=owner)
[...]

PR #281, opened 2019-01-16
call related update credential cloud facade methods based on facade version

ycliuhw:
call related update credential cloud facade methods based on facade version.

fix issue: https://github.com/juju/python-libjuju/issues/280

### DIFF
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -217,7 +217,12 @@ class Controller:
 
         log.debug('Uploading credential %s', name)
         cloud_facade = client.CloudFacade.from_connection(self.connection())
-        await cloud_facade.UpdateCredentials([
+        if cloud_facade.version >= 3:
+            # UpdateCredentials was renamed to UpdateCredentialsCheckModels in facade version 3.
+            update_credentials_func = cloud_facade.UpdateCredentialsCheckModels
+        else:
+            update_credentials_func = cloud_facade.UpdateCredentials
+        await update_credentials_func([
             client.UpdateCloudCredential(
                 tag=tag.credential(cloud, tag.untag('user-', owner), name),
                 credential=credential,
--- a/tests/unit/test_controller.py
+++ b/tests/unit/test_controller.py
@@ -86,7 +86,7 @@ class TestControllerConnect(asynctest.TestCase):
                                         max_frame_size='max_frame_size')
 
     @asynctest.patch('juju.client.client.CloudFacade')
-    async def test_file_cred(self, mock_cf):
+    async def test_file_cred_v2(self, mock_cf):
         with NamedTemporaryFile() as tempfile:
             tempfile.close()
             temppath = Path(tempfile.name)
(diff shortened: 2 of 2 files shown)


juju/controller.py
```diff
@@ -182,6 +182,9 @@ def controller_name(self):
             Defaults to the same cloud as the controller.
         :param str owner: Username that will own the credential. Defaults to
             the current user.
+        :param bool force: Force indicates whether the update should be forced.
+            It's only supported facade 3 or later.
```
anastasiamac: "supported ***for*** facade v3 or..."

ycliuhw: thx

jameinel:
I think this fixes the immediate issue, but I don't think it changes libjuju so that it defaults to being compatible.
If you look at how github.com/juju/juju/api is done, it tracks a list of versions for each facade that it understands.
And it negotiates/sends requests in the version that it understands.
And then we have a separate test that says "does api/ use the same version as apiserver".
Because you *should* be able to talk to a 2.5 controller and make a V2 request to the UpdateCredential method.

Anyway, I wouldn't block this landing, but I do think it is part of the fix we need, but missing the key part that means people have to scramble whenever we make a change.

jameinel:
To clarify my point as there was some confusion.
This is the correct fix to support version 3 of the API. We should definitely land it, I didn't convey that correctly.

However, there is fundamental issue with libjuju as it shouldn't break when we introduce a *new* version. That is a bigger fix we should look to be making

ycliuhw:
hi @jameinel @wallyworld ,
yes, I agreed with the better approach @jameinel mentioned which is more like a fundamental/general **feature**.
Are you happy to land this **fix** for the specific issue https://github.com/juju/python-libjuju/issues/280 for now then we could discuss with maintainers of this repo to implement the client facade best version feature later?
Thanks.

jameinel:
yes, please land this

John
=:->

ycliuhw:
$$merge$$
:shipit:

johnsca:
@jameinel FWIW, the library does [handle](https://github.com/juju/python-libjuju/blob/master/juju/client/facade.py#L46-L60) downgrading to the highest facade version that it and the controller both support.  The facade versions considered to be supported by the library is based on what was been synced into `juju.client.*` via [schemagen](https://github.com/juju/schemagen).

The code that failed here is in the higher-level wrapper around the facades (what we generally call the object layer) which does need to be manually updated after a schemagen sync to handle any changes like this to the facade interfaces.

I assume that this new facade version came in via #274 and there was obviously already a test case covering the code that failed, so I really don't understand how this wasn't caught in #274 since the Travis tests run against both the stable Juju snap as well as edge.  The only thing I can think of is that, since the  sync happened on an unfinalized version of Juju, perhaps the facade version was updated prior to the sync for a different change and then this rename happened after the sync?  A couple of ways we could prevent this in the future would be to: only sync on finalized versions of Juju (far from ideal), or get the libjuju tests that are currently only run on PRs for the libjuju repo to also be run against PRs on the juju repo (which of course would need to ensure that it used the Juju code from the PR and not just the edge channel).  Since we already have the Travis config for running those tests, perhaps we can work together to get a modified version set up in the Juju repo as well.

johnsca:
I should also note that I don't think having a manually maintained list of facade versions supported in the object layer would have helped here, since I almost certainly would have updated it when I synced the new facades and the tests passed.

PR #284, opened 2019-01-17
Add Twine for dist upload on release

### LINKED ISSUE #283: release process needs updating to using twine
To do a release on the new pypi we need to update to using twine to upload:

    make release
    .tox/py3/bin/pip3 install --upgrade Twine
    .tox/py3/bin/twine upload --repository-url https://upload.pypi.org/legacy/ dist/*

mitechie:
Fixes #283 

- Add make lint target
- Update deps to be in order

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,10 @@ endif
 test:
 	tox
 
+.PHONY: lint
+lint: 
+	tox lint --notest
+
 docs: .tox
 	$(PIP) install -r docs/requirements.txt
 	rm -rf docs/_build/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,6 +22,10 @@ endif
 test:
 	tox
 
+.PHONY: lint
+lint: 
+	tox lint --notest
```
johnsca: This should instead be: `tox -e lint`

Makefile
```diff
@@ -30,7 +34,8 @@ docs: .tox
 
 release:
 	git fetch --tags
-	$(PY) setup.py sdist upload
+	$(PY) setup.py sdist
+	$(BIN)/twine upload --repository-url https://upload.pypi.org/legacy/ dist/*
```
johnsca: Do we need to clear old build artifacts from `dist/` so they don't get picked up here, or does twine handle that automatically?

mitechie:
@johnsca PTAL

PR #288, opened 2019-02-01
Add retry for connection if all endpoints fail

johnsca:
Sometimes, particularly during reconnects, the first attempt to connect will fail on all endpoints due to network delays or controller issues.

This adds a retry (default 3 attempts) to try all of the endpoints again.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -104,6 +104,8 @@ class Connection:
             bakery_client=None,
             loop=None,
             max_frame_size=None,
+            retries=3,
+            retry_backoff=10,
     ):
         """Connect to the websocket.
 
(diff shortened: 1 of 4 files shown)


juju/client/connection.py
```diff
@@ -446,16 +456,30 @@ def connect_params(self):
...
+                except ConnectionError:
+                    continue  # ignore; try another endpoint
+            else:
+                _endpoints_str = ', '.join([endpoint
+                                            for endpoint, cacert in endpoints])
+                if attempt < self._retries:
+                    log.debug('Retrying connection to endpoints: {}; '
+                              'attempt {} of {}'.format(_endpoints_str,
+                                                        attempt + 1,
+                                                        self._retries + 1))
```
kwmonroe: got a little crazy with the `x + 1` here didn't ya?  this is gonna say `attempt [1|2|3] of 4` when `retries = 3`.

johnsca: One initial attempt, plus 3 retries equals 4 total attempts.  *shrug*  I guess it depends on whether you think the `retries` param should be "total number of times to try" vs "number of times to try again if it fails".

kwmonroe: aaaah, ok.  i retract my objection, but know that nobody but you thinks this way.

PR #290, opened 2019-02-15
Additional work with annotations.

howbazaar:
Support setting and getting annotations from models, machines, applications, and units.

There is still the Annotations type at the top level that I don't think we need.

Actually setting machine annotations didn't work as it was passing the machine id not the tag. This is also fixed here.

My juju-annotations plugin tests the behaviour of this, but not sure what other tests there should be.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -312,6 +312,21 @@ class Application(model.ModelEntity):
             [],
         )
 
+    async def get_annotations(self):
+        """Get annotations on this application.
+
+        :return dict: The annotations for this application
+        """
+
+        facade = client.AnnotationsFacade.from_connection(
+            self.connection)
+
+        result = (await facade.Get([{"tag": self.tag}])).results[0]
+        if result.error is not None:
+            raise errors.JujuError(result.error)
+
+        return result.annotations
+
     async def set_annotations(self, annotations):
         """Set annotations on this application.
 
@@ -319,16 +334,18 @@ class Application(model.ModelEntity):
             pairs.
 
         """
+        # TODO: ensure annotations is dict with only string keys
+        # and values.
         log.debug('Updating annotations on application %s', self.name)
 
-        self.ann_facade = client.AnnotationsFacade.from_connection(
+        facade = client.AnnotationsFacade.from_connection(
             self.connection)
 
         ann = client.EntityAnnotations(
             entity=self.tag,
             annotations=annotations,
         )
-        return await self.ann_facade.Set([ann])
+        return await facade.Set([ann])
 
     async def set_config(self, config):
         """Set configuration options for this application.
--- a/juju/application.py
+++ b/juju/application.py
@@ -15,7 +15,8 @@
 import asyncio
 import logging
 
-from . import annotation, model, tag
+from . import model, tag
+from .annotationhelper import _get_annotations, _set_annotations
 from .client import client
 from .errors import JujuError
 from .placement import parse as parse_placement
@@ -317,7 +318,7 @@ class Application(model.ModelEntity):
 
         :return dict: The annotations for this application
         """
-        return await annotation._get_annotations(self.tag, self.connection)
+        return await _get_annotations(self.tag, self.connection)
 
     async def set_annotations(self, annotations):
         """Set annotations on this application.
@@ -326,8 +327,7 @@ class Application(model.ModelEntity):
             pairs.
 
         """
-        return await annotation._set_annotations(
-            self.tag, annotations, self.connection)
+        return await _set_annotations(self.tag, annotations, self.connection)
 
     async def set_config(self, config):
         """Set configuration options for this application.
--- a/juju/application.py
+++ b/juju/application.py
@@ -15,7 +15,7 @@
 import asyncio
 import logging
 
-from . import model, tag
+from . import annotation, model, tag
 from .client import client
 from .errors import JujuError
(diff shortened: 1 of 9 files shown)


johnsca:
This looks great, and the test failures are connection issues with the lxd controller, unrelated to this change.

I do think that tests for this behavior should be added to `tests/integration/test_{model,application,unit,machine}.py`.  Just a simple test that sets then retrieves them.

juju/application.py
```diff
@@ -312,23 +312,40 @@ def collect_metrics(self):
             [],
         )
 
+    async def get_annotations(self):
```
johnsca: To reduce the repetition of these methods, maybe we should factor them up to a `AnnotatableEntity` mixin class?  I was thinking it could just be moved up to the `ModelEntity` base, but `Relation` is also a `ModelEntity` and maybe isn't annotatable (but should it be)?

The `Annotation` class is also a `ModelEntity` but I think we should just remove that entirely, since it is unused and overkill.  Would you mind doing that in this PR?

juju/application.py
```diff
@@ -312,23 +312,40 @@ def collect_metrics(self):
...
-        self.ann_facade = client.AnnotationsFacade.from_connection(
+        facade = client.AnnotationsFacade.from_connection(
             self.connection)
 
         ann = client.EntityAnnotations(
             entity=self.tag,
             annotations=annotations,
         )
-        return await self.ann_facade.Set([ann])
+        return await facade.Set([ann])
```
johnsca: Do we actually want to return anything here?  And do we need to check the result object for any errors?

juju/model.py
```diff
@@ -543,6 +543,7 @@ def connection(self):
                 raise ValueError('Authentication parameters are required '
                                  'if model_name not given')
             await self._connector.connect(**kwargs)
+        self.uuid = self.connection().connect_params()['uuid']
```
johnsca: This line is causing the unit test failures:

```
  File "/home/travis/build/juju/python-libjuju/tests/unit/test_model.py", line 212, in test_with_endpoint_and_uuid_with_bakery
    bakery_client='bakery')
  File "/home/travis/build/juju/python-libjuju/juju/model.py", line 546, in connect
    self.uuid = self.connection().connect_params()['uuid']
  File "/home/travis/build/juju/python-libjuju/juju/model.py", line 444, in connection
    return self._connector.connection()
  File "/home/travis/build/juju/python-libjuju/juju/client/connector.py", line 49, in connection
    raise NoConnectionException('not connected')
juju.client.connector.NoConnectionException: not connected
```

The `self._connector.connect()` call gets mocked out to a no-op, but this call isn't intercepted.

howbazaar:
@johnsca I think I'm going to need a hand getting these tests sorted.

I've had a stab at mocking out the connection params to return a uuid. Take a look and see if they are ok.

Secondly, how do you run the integration tests?

Was looking at adding something like this:

```go
@base.bootstrapped
@pytest.mark.asyncio
async def test_model_annotations(event_loop):

        annotations = await model.get_annotations()
        assert len(annotations) == 0

        expected = {"foo": "bar", "another": "value"}
        await model.set_annotations(expected)

        annotations = await model.get_annotations()
        assert annotations == expected
```
How does that look?

juju/application.py
howbazaar: I just used some stand alone functions. The model had connection as a function whereas machine, unit, and application had a property.

I also wasn't clear that the mixin could reach down into attributes of the derived class in order to get acces to tag and connection.

juju/application.py
howbazaar: Hmm... probably. Looked for examples in the code and all the examples I found for set methods were using this pattern.  Something to queue up for later maybe?

johnsca: +1 to remaining consistent.  The return value seems unnecessary and a bit leaky, but isn't really an issue, so maybe refactor at some point down the line, yeah.

juju/application.py
johnsca: A mixin could use other attributes on the derived class as long as it documented that as part of the required interface, but I think the helper functions are just as good, just a different style.

johnsca: Also, I dislike the inconsistency between the various `connection` properties / methods, but that was an oversight in the macaroon bakery work.  Not sure if we could reasonably clean it up at this point.

howbazaar:
Added the Annotation class back in.

johnsca:
```
juju/annotation.py:10: in <module>
    class Annotation(model.ModelEntity):
E   AttributeError: module 'juju.model' has no attribute 'ModelEntity'
```

I have no idea what's going on there.  The class is definitely in that module.

howbazaar:
When I run `tox -e lint` it is all good. :confused:

howbazaar:
Ah... I think this is a circular reference problem.

howbazaar:
@johnsca the integration test failures have nothing to do with the changes I have proposed. What is their rate of failure normally?

johnsca:
@howbazaar The tests are usually reliable for the most part, except for the occasional intermittent connection failures that seem mostly to do with load on the Travis servers.  Usually restarting the job will result in a passing run.  I would like to see the tests improved to have better handling of these types of failures, but I'm not entirely clear on how to do so.  The SSH test in particular does fail somewhat often, but I'm really not sure how we could improve it.

johnsca:
This is released in 0.11.4

Issue #292, opened 2019-03-12
ValueError: not enough values to unpack (expected 4, got 3)


johnsca:
Originally reported here: https://github.com/conjure-up/conjure-up/issues/1591

libjuju started failing to deploy bundles with Juju 2.5.2

```python-traceback
[gw0] linux -- Python 3.6.7 /home/johnsca/juju/python-libjuju/.tox/py3/bin/python
Traceback (most recent call last):
  File "/home/johnsca/juju/python-libjuju/tests/integration/test_model.py", line 88, in test_deploy_bundle
    await model.deploy('bundle/wiki-simple')
  File "/home/johnsca/juju/python-libjuju/juju/model.py", line 1283, in deploy
    await handler.execute_plan()
  File "/home/johnsca/juju/python-libjuju/juju/model.py", line 1966, in execute_plan
    result = await method(*step.args)
  File "/home/johnsca/juju/python-libjuju/juju/model.py", line 2127, in deploy
    devices, resources, num_units, placement = args
ValueError: not enough values to unpack (expected 4, got 3)
==================================================================================================== short test summary info ====================================================================================================
FAILED tests/integration/test_model.py::test_deploy_bundle
```

This succeeds on 2.5.1

```
[gw2] PASSED tests/integration/test_model.py::test_deploy_bundle 
```

johnsca:
I think this may have been caused by https://github.com/juju/bundlechanges/pull/48

johnsca:
@wallyworld @ycliuhw  Can one of you guys take a look at this?

PR #293, opened 2019-03-13
k8s bundles no longer have application placement

wallyworld:
k8s bundles no longer support application placement, which is now done using constraints.
Fixes issue #292

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2108,11 +2108,6 @@ class BundleHandler:
             will be 0 and separate AddUnitChanges will be used.  For Kubernetes
             models, this will be used to scale the application.
             (Only given on Juju 2.5+)
-
-        :param placement string:
-            Placement holds the placement directive for units of this
-            application.  Only applicable for Kubernetes models.
-            (Only given on Juju 2.5+)
         """
         # resolve indirect references
         charm = self.resolve(charm)
@@ -2120,11 +2115,11 @@ class BundleHandler:
         if len(args) == 1:
             # Juju 2.4 and below only sends the resources
             resources = args[0]
-            devices, num_units, placement = None, None, None
+            devices, num_units = None, None
         else:
             # Juju 2.5+ sends devices before resources, as well as num_units
-            # and placement
-            devices, resources, num_units, placement = args
+            # There might be placement but we need to ignore that.
+            devices, resources, num_units = args[:3]
 
         if not charm.startswith('local:'):
             resources = await self.model._add_store_resources(
@@ -2139,7 +2134,6 @@ class BundleHandler:
             resources=resources,
             storage=storage,
             devices=devices,
-            placement=placement,
             num_units=num_units,
         )
         return application


juju/model.py
```diff
@@ -2108,23 +2108,17 @@ def resolve(self, reference):
...
         if len(args) == 1:
             # Juju 2.4 and below only sends the resources
             resources = args[0]
-            devices, num_units, placement = None, None, None
+            devices, num_units = None, None, None
         else:
             # Juju 2.5+ sends devices before resources, as well as num_units
-            # and placement
-            devices, resources, num_units, placement = args
+            devices, resources, num_units = args
```
howbazaar: Hasn't this just broken deploys on 2.5.0 and 2.5.1?

wallyworld: Not anymore :-)

juju/model.py
```diff
@@ -2108,23 +2108,18 @@ def resolve(self, reference):
...
             # Juju 2.4 and below only sends the resources
             resources = args[0]
-            devices, num_units, placement = None, None, None
+            devices, num_units = None, None
         else:
             # Juju 2.5+ sends devices before resources, as well as num_units
-            # and placement
-            devices, resources, num_units, placement = args
+            # There might be placement but we need to ignore that.
+            devices, resources, num_units = args[:3]
```
jameinel: Model.deploy is using:
    async def deploy(
            self, entity_url, application_name=None, bind=None, budget=None,
            channel=None, config=None, constraints=None, force=False,
            num_units=1, plan=None, resources=None, series=None, storage=None,
            to=None, devices=None):

BundleHandler.deploy isn't. Is it plausible to change bundlehandler to use the named argument form instead of strictly positional?

I don't see where in the code it is calling \.deploy, so I'm guessing it is something at a different level?

I see places in examples, etc, but nothing under juju/* that is calling deploy() directly.

*args is always going to be a problem as you have to pass exactly only 1 thing in the 5th place, and we should never ever change what that thing is. And if you are removing placement in this patch, and we ever need to *add* something in the future, its going to break because then both 'placement' and 'newthing' end up in position arg[3].

johnsca: `BundleHandler.deploy()` is dispatched dynamically based on the `BundleChangeResult` objects returned by `BundleFacade.GetChanges()`:

https://github.com/juju/python-libjuju/blob/2acbdc400a98e97882b902d922383bd64e4fb101/juju/model.py#L1963-L1967

It might make it more clear to explicitly enumerate the mapping between the `step.method` value and the actual method being called, but Python doesn't have a switch / select statement, so it would have to be done with a large `if` / `elif` block or a `dict` lookup table.

Unfortunately, the positional arg usage comes from the `BundleChangeResult` structure and I think is ultimately dictated by the GUI.  And since JavaScript doesn't support named arguments, I don't know that we can realistically change it.  But if that is an option, I'm definitely for it.

jameinel:
https://github.com/juju/juju/blob/8192adf7820be9f9a4a94e70d7109af80103d81e/apiserver/params/params.go#L871

ugh. Why is this our interface to the system.
That seems rife for breakage.
I'm guessing the issue was that at some point in the past that was the way
they did it in the GUI and so we preserved their syntax. But our methods
take formatted JSON not arbitrary args. We could just as easily have Args
be a map[string]interface{} if we wanted to.

The other question is why we are using GetBundleChanges at all. AFAICT it
isn't used inside the Juju codebase, and we deploy bundles just fine.

@howbazaar any thoughts on how we could move the libjuju to doing bundles
the way you updated the CLI to do them? Or are you using GetBundleChanges
there as well, and I'm just not finding it?

johnsca:
Tests seem to be consistently failing with one of two errors:

```python-traceback
  File "/home/travis/build/juju/python-libjuju/tests/integration/test_application.py", line 72, in test_upgrade_charm
    await app.upgrade_charm()
  File "/home/travis/build/juju/python-libjuju/juju/application.py", line 520, in upgrade_charm
    storage_constraints=None
  File "/home/travis/build/juju/python-libjuju/juju/client/facade.py", line 420, in wrapper
  File "/home/travis/build/juju/python-libjuju/juju/client/_client8.py", line 1075, in SetCharm
    reply = await self.rpc(msg)
  File "/home/travis/build/juju/python-libjuju/juju/client/facade.py", line 545, in rpc
  File "/home/travis/build/juju/python-libjuju/juju/client/connection.py", line 326, in rpc
juju.errors.JujuAPIError: unable to set charm profile: unit "ubuntu/0" is not assigned to a machine
```

And:

```python-traceback
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/pylxd/client.py", line 299, in __init__
    response = self.api.get()
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/pylxd/client.py", line 149, in get
    response = self.session.get(self._api_endpoint, *args, **kwargs)
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/requests/sessions.py", line 546, in get
    return self.request('GET', url, **kwargs)
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/home/travis/build/juju/python-libjuju/.tox/py3/lib/python3.6/site-packages/requests/adapters.py", line 498, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', PermissionError(13, 'Permission denied'))
```

Neither of those seem related to this change and I can't reproduce them locally.  I'm rather confused on what might cause either one.  Any thoughts?

johnsca:
@jameinel I believe it started with the GUI and we just copied what the GUI was doing in libjuju, but I do feel that it's valuable to limit the number of places where we have to (re)implement the parsing and translation of a bundle into model actions.  That said, I think what we currently have is probably just as complicated as reimplementing the bundle logic anyway.

I suspect that the only reason why we don't just hand over the YAML and let Juju do all of the work is because the GUI wanted to represent the bundle deployment as a set of pending actions that you could then tweak before committing, but we definitely don't need that in libjuju so my preference would be to just give the bundle to Juju and be done with it.

PR #296, opened 2019-04-01
Check server cert.

howbazaar:
The client connection code was not checking the server cert. The purpose of the cert needs to be set to SERVER_AUTH not CLIENT_AUTH for this to work.

### DIFF
--- a/tests/integration/test_connection.py
+++ b/tests/integration/test_connection.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 from juju.client import client
 from juju.client.connection import Connection
+from juju.client.jujudata import FileJujuData
 from juju.controller import Controller
 from juju.utils import run_with_interrupt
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_connection.py
```diff
@@ -234,3 +235,46 @@ def _find_free_port(self):
...
+
+    try:
+        connection = await Connection.connect(
+            endpoint=endpoint,
+            username=account['user'],
+            password=account['password'],
+            cacert=my_random_selfsigned_cert,
+        )
+        await connection.close()
+        raise RuntimeError("connection shouldn't be allowed with self signed cert.")
```
johnsca: This will work fine as-is, but the convention for explicitly failing a test using pytest is [`pytest.fail(msg)`](https://docs.pytest.org/en/latest/reference.html#pytest-fail)

johnsca:
This was manually merged in https://github.com/juju/python-libjuju/commit/068161f41c4011cdf1d92959d833c7346f90ac2a but due to the squash commit, GH isn't detecting it, so I'm just going to close this PR.

johnsca:
This is released in 0.11.4

tests/integration/test_connection.py
howbazaar: Yeah... I wasn't sure exactly what the convention was.

PR #297, opened 2019-04-01
Fix the fix for the upgrade charm tests

johnsca:
We have to actually wait until the unit is assigned to a machine, not just until the unit is available.


johnsca:
This PR isn't getting properly picked up by Travis, which is the whole point.  So I'm going to close and re-open it in a new branch name.

PR #298, opened 2019-04-01
Fix the fix for the upgrade charm tests

johnsca:
We have to actually wait until the unit is assigned to a machine, not just until the unit is available.


johnsca:
Travis is refusing to pick up the branch for this PR, even after renaming it, and it won't even allow me to trigger a  manual build for this branch.  I'm just going to merge it and run the tests on master.  They pass for me locally, so at worst, it is status quo.

Issue #295, opened 2019-03-28
Model._after_connect hangs on controller created with juju edge


knkski:
I spun up a controller with Juju edge (`2.6-beta1+develop-ba5b72c`) and tried to run this test: https://github.com/juju-solutions/kubernetes-jenkins/blob/master/jobs/integration/test_kubeflow.py

The result was a hung test, with no normal output. Stable juju works fine. Here's the verbose output:

```
env PY_IGNORE_IMPORTMISMATCH=1 CONTROLLER=microk8s-controller MODEL=microk8s-model pytest -v -s integration/test_kubeflow.py::test_validate --fulltrace
=================================================================================== test session starts ====================================================================================
platform linux -- Python 3.6.7, pytest-4.0.2, py-1.7.0, pluggy-0.8.1 -- VIRTUALENV/bin/python3
cachedir: .pytest_cache
rootdir: /projects/charms/kubernetes-jenkins/jobs, inifile:
plugins: xdist-1.26.0, forked-1.0.1, asyncio-0.10.0
collected 1 item                                                                                                                                                                           

integration/test_kubeflow.py::test_validate ^C

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

config = <_pytest.config.Config object at 0x7f17e4949b70>, doit = <function _main at 0x7f17e4e5b730>

    def wrap_session(config, doit):
        """Skeleton command line program"""
        session = Session(config)
        session.exitstatus = EXIT_OK
        initstate = 0
        try:
            try:
                config._do_configure()
                initstate = 1
                config.hook.pytest_sessionstart(session=session)
                initstate = 2
>               session.exitstatus = doit(config, session) or 0

VIRTUALENV/lib/python3.6/site-packages/_pytest/main.py:185: 
[...]

PR #299, opened 2019-04-01
Handle deltas of unknown types.

howbazaar:
Juju 2.6 introduces a Delta type for 'charm'. This is because Juju is using the multiwatcher to populate an internal cache. There are likely to be new additions as well.

This branch makes libjuju more robust to unknown types. It will log at debug when it sees one, but otherwise ignores them.

Against a 2.6-beta1 controller and master `tox -e integration -- -k test_full_status` blocks forever.

With this patch it passes.

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -851,9 +851,12 @@ class Model:
                             pass  # can't stop on a closed conn
                         break
                     for delta in results.deltas:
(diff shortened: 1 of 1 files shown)


howbazaar:
Addresses #295.

PR #305, opened 2019-05-22
Disable hostname checking on controller connection

### LINKED ISSUE #302: Certificate Error while connecting to application
I've got an Ambassador instance that I've exposed like this:

    juju config kubeflow-ambassador juju-external-hostname=localhost
    juju expose kubeflow-ambassador

Previously that worked for me, however after refreshing on the edge channel of the juju snap, I get this error when attempting to connect to the instance:

```python
CertificateError("hostname '10.152.183.243' doesn't match either of 'anything', 'juju-apiserver', 'juju-mongodb', 'localhost'",)
```

I tried exposing Ambassador with an external hostname of the IP address and got the same error. @johnsca seemed to think it was related to https://github.com/juju/python-libjuju/pull/296, cc @howbazaar

johnsca:
The hostnames listed in the controller cert don't include the IP address which is typically what the endpoint address uses for self-bootstrapped controllers.  This means the hostname check will always fail for that
case.  However, since we pre-share and trust the cert from the connection info, along with the endpoint address, it is safe to skip that check during the connection.

Fixes #302

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -189,8 +189,17 @@ class Connection:
         return self.monitor.status == Monitor.CONNECTED
 
     def _get_ssl(self, cert=None):
-        return ssl.create_default_context(
+        context = ssl.create_default_context(
             purpose=ssl.Purpose.SERVER_AUTH, cadata=cert)
+        if cert:
+            # Disable hostname checking if and only if we have an explicit cert
+            # to validate against, because the cert doesn't contain the IP addr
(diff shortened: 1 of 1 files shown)


johnsca:
Per https://github.com/juju/juju/blob/develop/api/apiclient.go#L1129-L1138 this in effect matches what the Juju CLI client does (forcing the host name to a known value is functionally equivalent to disabling the check).

johnsca:
Released in 0.11.6

PR #306, opened 2019-05-31
Skip intermittent test failures

SimonRichardson:
The following commit skips intermittent test failures whilst we work
on fixing each one at a time. This is so we can see green for juju
CI tests and then improve as we go. The following are baby steps as
we integrate libjuju into the juju team.

### DIFF
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -27,6 +27,7 @@ SSH_KEY = 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCsYMJGNGG74HAJha3n2CFmWYsOOaORn
 @base.bootstrapped
 @pytest.mark.asyncio
 async def test_deploy_local_bundle_dir(event_loop):
+    pytest.skip('test_deploy_local_bundle_dir intermittent test failure')
     tests_dir = Path(__file__).absolute().parent.parent
     bundle_path = tests_dir / 'bundle'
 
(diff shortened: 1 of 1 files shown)


SimonRichardson:
@mitechie done

SimonRichardson:
$$merge$$

johnsca:
The issue is that the specific tests that intermittently fail change frequently.  The underlying issues are that we're doing too much work on overloaded Travis workers (running both the tests and controllers there) and so hitting timeouts within the tests, as well as timeouts and network connectivity issues with tests that require store interactions.

One approach that was suggested was to use Travis secrets and bootstrap the controller on, say, AWS instead of lxd on the Travis worker.  That might make it more reliable by separating the work done by the tests from the work done by the controller & units.

Another, arguably better option, is to a better job of having a smaller number of integration tests that cover the cases we need, rather than spinning up so many models with tons of overlap in what they're deploying.  That will be a larger refactor that needs to be done, though.  We could start this by picking a smaller subset of tests that cover most of the cases and mark those to be the only ones run on Travis, then gradually move the checks from the rest of the tests into those.

All that said, I don't see much benefit in merging this PR, since as you can see from the current Travis results, it's not resolving the issue.

SimonRichardson:
@johnsca this is locally against 2.5 and 2.6, also with Jenkins CI on the juju side.

SimonRichardson:
$$merge$$

johnsca:
Ah, I see.  +1

mitechie:
$$merge$$

mitechie:
$$merge$$

PR #313, opened 2019-06-06
Trusted bundles

SimonRichardson:
The following is the start of trusted bundles for the client. Work
also needs to be done to test that charms can be deployed along with
trust.

### Success

```console 
$ juju bootstrap lxd test --no-gui

$ tox -e example -- examples/deploy_bundle_with_trust.py

$ juju config ubuntu | grep -A 6 application-config
application-config:
  trust:
    default: false
    description: Does this application have access to trusted credentials
    source: user
    type: bool
    value: true    <- all set!
```

### Error:

Apply changes:

```diff
diff --git a/examples/deploy_bundle_with_trust.py b/examples/deploy_bundle_with_trust.py
index 90a0f44..56173dc 100644
--- a/examples/deploy_bundle_with_trust.py
+++ b/examples/deploy_bundle_with_trust.py
@@ -22,7 +22,7 @@ async def main():
             'cs:~juju-qa/bundle/basic-trusted-1',
             series='bionic',
             channel='beta',
-            trust=True,
+            trust=False,
         )
 
         print('Waiting for active')
```

```console
$ tox -e example -- examples/deploy_bundle_with_trust.py

juju.errors.JujuError: Bundle cannot be deployed without trusting applications with your cloud credentials.
Please repeat the deploy command with the --trust argument if you consent to trust the following application
 - ubuntu
```

### Force

Apply changes:

```diff
diff --git a/examples/deploy_bundle_with_trust.py b/examples/deploy_bundle_with_trust.py
index 90a0f44..ffab5b4 100644
--- a/examples/deploy_bundle_with_trust.py
+++ b/examples/deploy_bundle_with_trust.py
@@ -22,7 +22,8 @@ async def main():
             'cs:~juju-qa/bundle/basic-trusted-1',
             series='bionic',
             channel='beta',
-            trust=True,
+            trust=False,
+            force=True,
         )
 
         print('Waiting for active')
```

```
$ tox -e example -- examples/deploy_bundle_with_trust.py

$ juju config ubuntu | grep -A 6 application-config
application-config:
  trust:
    default: false
    description: Does this application have access to trusted credentials
    source: user
    type: bool
    value: false
```

### DIFF
--- a/examples/deploy_bundle.py
+++ b/examples/deploy_bundle.py
@@ -15,14 +15,13 @@ from juju.model import Model
 async def main():
     model = Model()
     print('Connecting to model')
-    # connect to current model with current user, per Juju CLI
+    # Connect to current model with current user, per Juju CLI
     await model.connect()
 
     try:
         print('Deploying bundle')
         applications = await model.deploy(
             'cs:~juju-qa/bundle/basic-0',
-            series='xenial',
             channel='beta',
         )
 
@@ -31,13 +30,14 @@ async def main():
             lambda: all(unit.workload_status == 'active'
                         for application in applications
                             for unit in application.units))
-
+        print("Successfully deployed!")
         print('Removing bundle')
         for application in applications:
             await application.remove()
     finally:
         print('Disconnecting from model')
         await model.disconnect()
+        print("Success")
 
 
 if __name__ == '__main__':
--- a/examples/deploy_bundle.py
+++ b/examples/deploy_bundle.py
@@ -1,0 +1,44 @@
+"""
+This example:
+
+1. Connects to the current model
+2. Deploy a bundle and waits until it reports itself active
+3. Destroys the units and applications
+
+"""
+import asyncio
+
+from juju import loop
+from juju.model import Model
+
+
+async def main():
+    model = Model()
+    print('Connecting to model')
+    # connect to current model with current user, per Juju CLI
+    await model.connect()
+
+    try:
+        print('Deploying bundle')
+        applications = await model.deploy(
+            'cs:~juju-qa/bundle/basic-0',
+            series='xenial',
+            channel='beta',
+        )
+
+        print('Waiting for active')
+        await model.block_until(
+            lambda: all(unit.workload_status == 'active'
+                        for application in applications
+                            for unit in application.units))
+
+        print('Removing bundle')
+        for application in applications:
+            await application.remove()
+    finally:
+        print('Disconnecting from model')
+        await model.disconnect()
+
+
+if __name__ == '__main__':
+    loop.run(main())
--- a/examples/deploy_bundle_with_trust.py
+++ b/examples/deploy_bundle_with_trust.py
@@ -3,7 +3,7 @@ This example:
 
 1. Connects to the current model
 2. Deploy a bundle with trust and waits until it reports itself active
-3. Destroys the unit and application
+3. Destroys the units and applications
 
 """
 from juju import loop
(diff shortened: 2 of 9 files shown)


examples/deploy_bundle_with_trust.py
```diff
@@ -0,0 +1,43 @@
...
+            channel='beta',
+            trust=True,
+        )
+
+        print('Waiting for active')
+        await model.block_until(
+            lambda: all(unit.workload_status == 'active'
+                        for application in applications
+                            for unit in application.units))
+
```
mitechie: Same with success message.

juju/model.py
```diff
@@ -1266,7 +1264,9 @@ def _get_series(self, entity_url, entity):
             '24/lxd/3' - place in container 3 on machine 24
 
             If None, a new machine is provisioned.
-
+        :param bool trust: Trust signifies that the charm should be deployed 
+            with access to trusted credentials. Hooks run by the charm can access
+            cloud credentials and other trusted access credentials.
```
mitechie: worth it to drop a link to the docs here for it?

juju/model.py
```diff
@@ -1303,7 +1303,7 @@ def _get_series(self, entity_url, entity):
                      (not is_local and 'bundle/' in entity_id))
 
         if is_bundle:
-            handler = BundleHandler(self)
+            handler = BundleHandler(self, trusted=trust, forced=force)
```
mitechie: where does the forced value come from?

tests/integration/test_model.py
```diff
@@ -93,6 +91,16 @@
...
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_deploy_trusted_bundle(event_loop):
+    async with base.CleanModel() as model:
+        await model.deploy('cs:~juju-qa/bundle/basic-trusted-1', trust=True)
+
+        for app in ('ubuntu', 'ubuntu-lite'):
+            assert app in model.applications
```
mitechie: should we just check the config/trust value of the expected charm?

examples/deploy_bundle.py
```diff
@@ -0,0 +1,44 @@
...
+    print('Connecting to model')
+    # connect to current model with current user, per Juju CLI
+    await model.connect()
+
+    try:
+        print('Deploying bundle')
+        applications = await model.deploy(
+            'cs:~juju-qa/bundle/basic-0',
+            series='xenial',
+            channel='beta',
```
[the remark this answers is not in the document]
SimonRichardson: It's because the bundle is in the `beta` channel, which the pylibjuju seems to be concerned about.

juju/model.py
SimonRichardson: I'm unsure, as we can't link to a heading in this atm https://docs.jujucharms.com/deploying-advanced-applications

juju/model.py
SimonRichardson: From the deploy method arguments - https://github.com/juju/python-libjuju/pull/313/files#diff-934c79f928f581e5f1d8903f275de68fR1239

juju/model.py
mitechie: Ah, missed it up there ty

mitechie:
Sorry, my link and share earlier was half-ass'd a bit. 

What I was going for was using @property and the getter/setter steps on that so that you can use code like



    application.trusted = True
    print(application.trusted)
    > True

You can do this using the example here 
https://www.datacamp.com/community/tutorials/property-getters-setters

check out the final example with the private getter/setter methods defined as the property.

SimonRichardson:
@mitechie we can't have a `@trusted.setter` for the old `set_trusted` method as it's not allowed, unless you know some other way to do it?

SimonRichardson:
$$merge$$

PR #315, opened 2019-06-17
Schemagen is no longer required

SimonRichardson:
Schemagen is no longer required because Juju itself has incorporated
the schema generation into the workflow of Juju. This should make it
a lot easier to build upon pylibjuju if it's always up to date with
Juju.

Work to do this has already been completed, see:

 1. https://github.com/juju/juju/pull/10336
 2. https://github.com/juju/juju/pull/10338

This should support 2.6 and onwards. Older releases can be done
the old way, but it requires you to have schemagen installed in the
old positions. The error messages might be more cryptic because of
that.

### DIFF
--- a/docs/upstream-updates/index.rst
+++ b/docs/upstream-updates/index.rst
@@ -13,24 +13,30 @@ Rarely, you may also have to add or update an override.
 Creating a Schema File
 ----------------------
 
-First, you will need to fetch SchemaGen_ and a copy of the Juju source.
-Once your copy of the Juju source is at the version you want to update to
-(probably the `develop` branch, or a release tag) and you have updated
-and reinstalled SchemaGen to reflect those changes, you just need to send
-the output into a file in the libjuju repository:
+Firstly, you'll need to checkout the Juju_ project and correctly have it
+setup according to it's documentation. Juju_ has the ability to generate
+the schema and stores it in version control (git). Copying that file from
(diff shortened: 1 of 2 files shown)


docs/upstream-updates/index.rst
```diff
@@ -13,24 +13,30 @@ Rarely, you may also have to add or update an override.
 Creating a Schema File
 ----------------------
 
-First, you will need to fetch SchemaGen_ and a copy of the Juju source.
-Once your copy of the Juju source is at the version you want to update to
-(probably the `develop` branch, or a release tag) and you have updated
-and reinstalled SchemaGen to reflect those changes, you just need to send
-the output into a file in the libjuju repository:
+Firstly, you'll need to checkout the Juju_ project and correctly have it
```
manadart: I tend to avoid abbreviations like "you'll", "won't", "shouldn't" etc in docs and formal communication, because it is analogous to slurring. Pedantry on my part does not block landing though :)

docs/upstream-updates/index.rst
```diff
@@ -13,24 +13,30 @@ Rarely, you may also have to add or update an override.
...
+setup according to it's documentation. Juju_ has the ability to generate
+the schema and stores it in version control (git). Copying that file from
```
manadart: "Copy" ?

SimonRichardson:
$$merge$$

PR #316, opened 2019-06-18
Update pylibjuju to 2.6.4 release to mirror upstream Juju

SimonRichardson:
This mirrors the Juju release 2.6.4.

Once this lands, I'll cherry-pick the changes to develop.


SimonRichardson:
> Do we have a changelog that should go with this?
That's a good point, I have no idea, we would have to extract them from Juju.

> I know we updated a example previously but want to make sure what's done is "done done".
OK let me look into this.

mitechie:
I'm fine if the changelog is pylibjuju specific. If we just updated the facades then call out what version of juju it's generated from and maybe link to the changelog of that juju version. If we add bits to pylibjuju like helpers/etc we note that as well.

SimonRichardson:
@mitechie I've updated the changelog in https://github.com/juju/python-libjuju/pull/316/commits/30686556f8bb5cb5039ba8b656d2af0822d8451f

docs/changelog.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+0.11.7
+^^^^^^
+Wednesday April 19 2019
+
+* Update facade methods for Juju 2.6
```
mitechie: 2.6.3? 2.6.4?

I think we should be explicit here. Not sure if we should go so far as to a sha it was built against or not though...

SimonRichardson:
$$merge$$

PR #318, opened 2019-06-27
Accept root-disk-source constraint

xiwenc:
When the client sets machine constraint as follow:

```
constraints: instance-type=n1-standard-1 root-disk-source=volume root-disk=100G
```

We get the following error response from the Juju controller running 2.6.4 which according to https://github.com/juju/juju/pull/10296, this constraint is a supported use case:

```
File "/usr/local/lib/python3.7/site-packages/juju/model.py", line 2073, in addMachines
raise ValueError("Error adding machine: %s" % error.message)
ValueError: Error adding machine: cannot add a new machine: constraint root-disk cannot be specified with instance-type unless constraint root-disk-source=volume
```
The error message indicates the controller did not receive the actual constraints we are passing.

After some digging it seems the attribute `root-disk-source` ends up in the `unknown_fields`. To fix this we have add support for `root-disk-source` in `class Value(Type)` in the file `juju/client/_definitions.py`. Then this attribute is passed down correctly to the controller.

The client bump happened in 2c945ef3

### DIFF
--- a/juju/client/_definitions.py
+++ b/juju/client/_definitions.py
@@ -11675,9 +11675,9 @@ class UserModelList(Type):
 
 
 class Value(Type):
-    _toSchema = {'arch': 'arch', 'container': 'container', 'cores': 'cores', 'cpu_power': 'cpu-power', 'instance_type': 'instance-type', 'mem': 'mem', 'root_disk': 'root-disk', 'spaces': 'spaces', 'tags': 'tags', 'virt_type': 'virt-type'}
-    _toPy = {'arch': 'arch', 'container': 'container', 'cores': 'cores', 'cpu-power': 'cpu_power', 'instance-type': 'instance_type', 'mem': 'mem', 'root-disk': 'root_disk', 'spaces': 'spaces', 'tags': 'tags', 'virt-type': 'virt_type'}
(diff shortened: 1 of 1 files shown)


mitechie:
!!build!!

mitechie:
Thanks, we'll try to get this landed and part of the next update.

xiwenc:
Thanks for the quick feedback!

mitechie:
I think that this patch will be auto done when we regenerate with the new API available. Let me double check with @SimonRichardson around if this file gets updated and see if this PR is necessary or we just need to run our update code.

SimonRichardson:
So this PR won't work, mainly because of a known issue on how we generate the definitions and clients from the schema. The problem is that optional api arguments don't currently translate to optional arguments of pylibjuju. In the previous release I moved them to `unknown_fields` for the previous release with the intention of finding a permanent solution long term.

With this in mind, your changes will be always over ridden when we do a `make client` (see https://discourse.jujucharms.com/t/python-libjuju/1553)... I'll spend sometime ASAP to remedy this issue.

SimonRichardson:
@xiwenc closing this in favour of the latest 2.6 changes https://github.com/juju/python-libjuju/blob/2.6/juju/client/_definitions.py#L11912 - I'll be doing a release real soon and I'll ping back here when done.

PR #321, opened 2019-07-02
Write all definitions

SimonRichardson:
The following writes all definitions to the definitions.py file.
This is a WIP as we're doing less looping and recursion, but it
makes more sense to do it this way.

 - Remove unwanted methods


juju/machine.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, *args, **kwargs):
                                            loop=model.loop)
         if model._watch_stopping.is_set():
             return
-        if model.info.agent_version < client.Number.from_json('2.2.3'):
+        if 'agent_version' in model.info and model.info.agent_version < client.Number.from_json('2.2.3'):
```
SimonRichardson: in newer versions of `ModelInfo` the `agent_version` is gone, so we need to check that it exists before checking.

SimonRichardson: Further investigation, this wasn't the case, we just wasn't exporting the right ModelInfo. I've fixed this so that we do indeed do that correctly.

SimonRichardson:
https://trello.com/c/NB3ycKIK/1611-pylibjuju-write-definitions-out-for-latest-version-of-the-schema

SimonRichardson:
@jameinel I updated the code to be backwards compatible with older api servers. Note that all integration tests don't work with older api servers as the integration tests actually use new features which might not be available with old api servers. An example being "trusted" against juju 2.3.

This commit message explains the changes https://github.com/juju/python-libjuju/pull/321/commits/24c770717bfabd47c36dc1eb87d4d08db6ee4abf

SimonRichardson:
The rest of the PR will attempt to ensure we use named (positional) arguments for all facade calls.

jameinel:
Is it possible to have the test recognize that fact/do you at least get a
decent error about it not being supported?

Certainly we don't expect support where it doesn't exist. It would be good
to have that clearly marked and tracked so we don't have to think about it
every time.

John
=:->

SimonRichardson:
@jameinel I don't know, but I can sure check!

juju/application.py
```diff
@@ -241,30 +243,35 @@ def collect_metrics(self):
...
+
+        app_config = config.application_config
+        return app_config['trust']['value'] is True
 
     async def set_trusted(self, trust):
         """Set the trusted configuration of the application.
 
         :param bool trust: Trust the application or not
         """
+        if self.model.info.agent_version < client.Number.from_json('2.4.0'):
```
jameinel: Shouldn't this be done with Facade versions? Is agent version really the only way we can detect this?

juju/client/_client1.py
```diff
@@ -43,7 +43,7 @@ class ActionPrunerFacade(Type):
     
 
     @ReturnMapping(ModelConfigResult)
-    async def ModelConfig(self):
```
jameinel: kwargs in responses makes a little bit of sense, but kwargs in requests feels a bit wrong. Either you know what to send or you don't. You don't happen to know just what to put in kwargs. If you *did* then you should be using a newer client lib that supports it, shouldn't you?
(kwargs in responses is more about not causing a runtime error because extra data was returned, but by the same token, it doesn't really matter because you can't know what to do with things that you don't know about yet.)
I guess all of these ignore anything passed in **kwargs, so it is just dropped?

juju/client/_client1.py
```diff
@@ -7697,8 +7701,8 @@ class ModelConfigFacade(Type):
                    request='SetSLALevel',
                    version=1,
                    params=_params)
+        _params['ModelSLAInfo'] = modelslainfo
```
jameinel: is this to make the current client backwards compatible with the old API ?

juju/client/_client10.py
```diff
@@ -658,9 +658,10 @@ class ApplicationFacade(Type):
     
 
     @ReturnMapping(AddRelationResults)
-    async def AddRelation(self, endpoints):
+    async def AddRelation(self, endpoints=[], via_cidrs=[], **kwargs):
```
jameinel: are these changes just meant to correctly match facade version 10 ?

juju/unit.py
```diff
@@ -138,11 +138,11 @@ def resolved(self, retry=False):
...
-            [],
-            command,
-            [],
-            timeout,
-            [self.name],
+            applications=[],
+            commands=command,
+            machines=[],
+            timeout=timeout,
+            units=[self.name],
```
jameinel: I like passing in the named parameters a lot. I think its much easier to understand the request than lots of '[], command, [], ' (what is the 3rd argument mean)

juju/model.py
```diff
@@ -1055,7 +1055,7 @@ def predicate(delta):
 
         # Submit the request.
         client_facade = client.ClientFacade.from_connection(self.connection())
-        results = await client_facade.AddMachines([params])
+        results = await client_facade.AddMachines(params=[params])
```
mitechie: why are we list-ifying params the object?

juju/model.py
```diff
@@ -1130,7 +1130,7 @@ def add_space(self, name, *cidrs):
 
         """
         key_facade = client.KeyManagerFacade.from_connection(self.connection())
-        return await key_facade.AddKeys([key], user)
+        return await key_facade.AddKeys(ssh_keys=[key], user=user)
```
mitechie: another where turning things to a list on passing in seems odd

jameinel:
you could set it to () if you want an iterable. [] just has the problem
that if you mutate the list it will stay mutated in the next call to the
function.

John
=:->

jameinel:
I'm pretty sure this is because the Facade needs a list of keys to add but
the client api is just adding one key.

John
=:->

juju/client/_client1.py
SimonRichardson: I've addressed this in the later commits, by dropping kwargs on requests and also pinning to a facade version. Moving to a later facade requires understanding the changes and things don't come for free and requires thought and coding to move to a later facade.

juju/client/connection.py
```diff
@@ -8,15 +8,106 @@
...
+import macaroonbakery.httpbakery as httpbakery
 import websockets
 from juju import errors, tag, utils
 from juju.client import client
 from juju.utils import IdQueue
 
 log = logging.getLogger('juju.client.connection')
 
+client_facades = {
+    'Action': { 'version': 2 },
```
jameinel: don't these need to be lists of supported versions (even if all the current versions are only length 1) ?

juju/client/connection.py
```diff
@@ -649,14 +649,12 @@ def _build_facades(self, facades):
...
             if name in client_facades:
-                self.facades[name] = client_facades[name]['version']
-            else:
-                # if a facade is passed back that we don't know about, just use
-                # the latest version, so we don't break against new controller
-                # versions.
-                self.facades[name] = facade['versions'][-1]
+                known = client_facades[name]['versions']
+                discovered = facade['versions']
+                version = max(set(known).intersection(set(discovered)))
```
jameinel: we should probably handle:
```
>>> x = [1, 2]
>>> y = [3, 4]
>>> max(set(x).intersection(y))
ValueError: max() arg is an empty sequence
```

```
try:
  version = ...
except ValueError:
  # facade does not have a common version, so it isn't a facade we can support
else:
  self.facades[name] = version
```

SimonRichardson: Good shout, didn't think about that one

SimonRichardson: @jameinel i've updated this in the subsequent commit https://github.com/juju/python-libjuju/pull/321/commits/7582e5208263c274b615ceceb642f9e769a24011

juju/client/connection.py
```diff
@@ -553,7 +645,23 @@ def connect_params(self):
...
+            # so in order to be compatible forwards and backwards we speak a
+            # common facade versions.
+            if name in client_facades:
+                try:
+                    known = client_facades[name]['versions']
+                    discovered = facade['versions']
+                    version = max(set(known).intersection(set(discovered)))
+                except ValueError:
+                    # this can occur if known is [1, 2] and discovered is [3, 4]
+                    # there is just know way to know how to communicate with the
```
jameinel: 'just no way'

juju/client/connection.py
```diff
@@ -553,7 +645,23 @@ def connect_params(self):
...
+                    # facades we're trying to call.
+                    raise errors.JujuConnectionError('unknown common facade version')
```
jameinel: My concern with raising an exception here is that all of these facades are being looked up *before* we actually try to use them. It seems a shame to prevent being able to operate against a server *at all* if *one* facade is not supported. I would be happy with it if we gave this sort of error when a shim calls "GetFacade('name')", but not in an 'iterate everything first.
I think its probably better to 
a) log a warning?
b) just treat the facade as not present.

juju/model.py
jameinel: params here is the template definition for 1 machine. AddMachines is a bulk API that would let you supply multiple templates at the same time. The shim is doing "add_machine" to add a single machine.

juju/model.py
jameinel: its another case where the shim is implementing "add_key()" singular, but the API is "add_keys()" plural

juju/client/connection.py
SimonRichardson: log a warning, feels better.

SimonRichardson:
$$merge$$

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #332, opened 2019-08-05
Consume bundle

SimonRichardson:
Adds the ability to consume a bundle with CMR offerings. Changes include add
missing facades in the client connection and the ability to override a facade so
that clients of pylibjuju can state that their controller uses a much higher version
of the facade than the pinned facade.

This adds a topographical sorting of the bundle changes so we can ensure that
we can process any large bundle file without running into issues.

Note: this moves the bundle logic to a new file to make it easier to grok.

## QA

```
tox -e example -- examples/crossmodel_bundle.py
tox -e integration -- tests/integration/test_crossmodel.py::test_add_bundle
```


SimonRichardson:
@mitechie this is now ready for a review.

examples/crossmodel.py
```diff
@@ -48,6 +49,10 @@
         print('Consuming offer')
         await consuming_model.consume("admin/test-cmr-1.mysql")
 
+        print('Exporting bundle')
```
mitechie: might move this log into the with statement so you can log where the file/path is expected to be in case folks are using this log to chase anything down? I guess it's just an example...never mind

examples/crossmodel_bundle.py
```diff
@@ -0,0 +1,84 @@
...
+from juju.client.connection import facade_versions
+from juju.controller import Controller
+
+log = getLogger(__name__)
+
+
+async def main():
+    controller = Controller()
+    print("Connecting to controller")
+    await controller.connect(specified_facades=facade_versions("Bundle", [1, 2, 3]))
```
mitechie: ? does the end user need to specify these facade versions?

examples/crossmodel_bundle.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        print('Creating models')
+        offering_model = await controller.add_model('test-cmr-1')
+        consuming_model = await controller.add_model('test-cmr-2')
+
+        print('Deploying mysql')
+        application_1 = await offering_model.deploy(
+            'cs:mysql-58',
+            application_name='mysql',
+            series='xenial',
```
mitechie: is this available on bionic?

SimonRichardson: good question.

SimonRichardson: Nope, unfortunately not :(

examples/crossmodel_bundle.py
SimonRichardson: Nope, I'll remove it.

SimonRichardson:
$$merge$$

Issue #335, opened 2019-08-12
Provisioning ssh without ubuntu user fails


AdamIsrael:
This is an outstanding issue with the SSH provisioner, which expects there to be an ubuntu user on the target machine being added to the Juju model.

For example, this would fail always fail, even though the root user had the private key added to its authorized keys:

```python
machine1 = await model.add_machine(spec='ssh:{}@{}:{}'.format(
    "root",
    host_address,
    private_key_path,
))
```
The provisioner always tries to connect as the 'ubuntu' user, ignoring the username passed to add_machine.

AdamIsrael:
Fix landed in master.

PR #336, opened 2019-08-12
Fix issue #335: Allow provisioning with non-"ubuntu" user

AdamIsrael:
This patch fixes an issue with the SSH provisioner, which previously
always assumed there was an "ubuntu" user with sudo access available on
the target machine.

The provisioner will now honor the username passed to
`model.add_machine`.

Added an additional integration test for testing the root user, as well
as modifying the existing integration test to use a randomized username
w/sudo access.

### DIFF
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -16,7 +16,6 @@ from juju.model import Model, ModelObserver
 from juju.utils import block_until, run_with_interrupt
 import random
 import string
-import sys
 from .. import base
 
 MB = 1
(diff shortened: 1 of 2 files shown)


mitechie:
@SimonRichardson @achilleasa can you two take a peek at this please and help progress it.

mitechie:
!!build!!

tests/integration/test_model.py
```diff
@@ -185,16 +190,28 @@
...
+        cloud_init = """
+        #cloud-config
+        users:
+        - name: {}
+          ssh_pwauth: False
+          ssh_authorized_keys:
+            - {}
+          sudo: ["ALL=(ALL) NOPASSWD:ALL"]
+          groups: adm, sudoers
+        """.format(test_user, public_key)
```
SimonRichardson: :+1:

mitechie:
$$merge$$

PR #326, opened 2019-07-25
Cross model relations: Offers

SimonRichardson:
Adding cross model offer add offer, remove offer and list offer.
The following also includes integration test around add and
removing offers.

## QA

You need to prepare a juju before doing the following:

```console
juju bootstrap lxd test --no-gui
```

Ensure that the example works by doing:

```console
tox -e example -- examples/crossmodel.py
```

Ensure that integration also works:

```console
tox -e integration -- tests/integration/test_crossmodel.py
```

PR #338, opened 2019-08-14
2.6 merge

SimonRichardson:
Merge 2.6 branch which includes the following:

 - #316
 - #317
 - #321
 - #326  
 - #327
 - #332 
 - #334

It's worth re-iterating what the 2.6 branch did as an overview, it added the ability 
to release more complex bundles, by sorting the bundle changes we get back 
from the change list we can topologically sort those changes to prevent dependency
conflicts.

Additionally with the sorting changes we can then handle cross-model relations (CMR)
bundles so that deploying a SAAS application can add relations to an application.

Finally I've bumped `theblues` dependency to `0.5.1` to prevent issues whereby the
charmstore would return a forbidden if the metadata `revision-info` is also included
in the api calls (see https://github.com/juju/theblues/commit/31c645b9e89e6c9dcc662f848f2b58ee6d8d1397).


SimonRichardson:
$$merge$$

PR #340, opened 2019-08-15
Add support for channels in bundles

achilleasa:
This PR enables the python library to handle per-charm channel specifications when deploying bundles. 

Support for bundles with per-charm channels has landed on juju develop branch via:  https://github.com/juju/juju/pull/10480


achilleasa:
$$merge$$

Issue #341, opened 2019-08-16
ambiguous constraints: "cores" overlaps with "instance-type"'


dannf:
As an attempt at working around #333 I switched to the juju edge channel (2.7-beta1+develop-6c8272e). While I can deploy charms from the command line, it is now failing via python-juju (2.6.0):
```
$ python3
Python 3.6.8 (default, Jan 14 2019, 11:02:34) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from juju.controller import Controller
>>> import asyncio
>>> async def main():
...     controller = Controller()
...     await controller.connect()
...     model = await controller.add_model('pyjuju')
...     app = await model.deploy('ubuntu', constraints={'arch': 'amd64'})
... 
>>> loop = asyncio.get_event_loop()
>>> loop.run_until_complete(main())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File "<stdin>", line 5, in main
  File "/usr/local/lib/python3.6/dist-packages/juju/model.py", line 1431, in deploy
    devices=devices,
  File "/usr/local/lib/python3.6/dist-packages/juju/model.py", line 1508, in _deploy
    result = await app_facade.Deploy(applications=[app])
  File "/usr/local/lib/python3.6/dist-packages/juju/client/facade.py", line 423, in wrapper
    reply = await f(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/juju/client/_client8.py", line 751, in Deploy
    reply = await self.rpc(msg)
  File "/usr/local/lib/python3.6/dist-packages/juju/client/facade.py", line 548, in rpc
    result = await self.connection.rpc(msg, encoder=TypeEncoder)
  File "/usr/local/lib/python3.6/dist-packages/juju/client/connection.py", line 472, in rpc
    raise errors.JujuError(err_results)
juju.errors.JujuError: ['cannot add application "ubuntu": ambiguous constraints: "cores" overlaps with "instance-type"']
```

mitechie:
Hmm, cores is intended to be cloud agnostic while instance types aren't. While it's true that they can overlap, I believe we rely on instance-type being more specific and winning. Can I ask what you attempted to do and what you expected to have happen? Maybe we can do better at managing the expectations there?

[...]

PR #342, opened 2019-08-19
Hide method default values behind a flag

SimonRichardson:
The following hides the new default values behind a feature flag,
so that experimental features don't impact existing setups.
Unfortunately default values introduce a problem where default
values are useful except when they're not. In this case they're not
valuable when setting default values for instance types. Moving to
the None type elevates this, but it means that we loose some of the
type goodness. We should instead look at where default types are
useful and when they're not and come up with a mechanism for
generating schema that knows when to use them and when not.
Potentially a white and black list?

Fixes https://github.com/juju/python-libjuju/issues/341


SimonRichardson:
Actual fix is here https://github.com/juju/python-libjuju/pull/342/files#diff-2a349e9e46915822f3ee8b7121ea72daR208

juju/client/facade.py
```diff
@@ -286,6 +294,20 @@ def as_kwargs(self):
             return ', '.join(parts)
         return ''
 
+    def as_validation(self):
+        if self:
```
mitechie: why the if self? is this called as a static method somewhere?

juju/client/facade.py
```diff
@@ -286,6 +294,20 @@ def as_kwargs(self):
...
+            parts = []
+            for item in self:
+                ''' if name is not None and not isinstance(name, type):
+                        raise JujuError("Expected {} to be of type {}".format(name, type))
```
mitechie: Expected {} to be a {}  

?

juju/client/facade.py
```diff
@@ -296,6 +318,17 @@ def get_doc(self):
         return self._get_arg_str(True, "\n")
 
 
+def buildValidation(name, instance_type, instance_sub_type, ident=None):
+    INDENT = ident or "    "
```
mitechie: hmm, this is because we're generating the code vs doing the work? 

Normally I'd think that we'd have a function call and we'd pop calls onto a stack based on what we found instead of building if blocks, but I guess we're writing the code that does the validation vs actually doing it at all here right?

juju/client/facade.py
SimonRichardson: that I don't know, this is the same code as `as_kwargs` and I had assumed it was done for a reason, if that's not the case, I can change that.

juju/client/facade.py
SimonRichardson: Sure, easy change

juju/client/facade.py
SimonRichardson: Exactly, this is actually build a if for building a validation, rather than build a validation block. I would also rather have a stack based system, where we pass a list through. It makes it much easier to reason about. I might do that in another round of refactoring at a later date.

juju/client/facade.py
mitechie: self is the instance of the class so if this were to be used like a static method I could see self not being defined. It's odd to see something that could be used in this way though.

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

Issue #344, opened 2019-08-21
libjuju 2.6.1 breaks juju.model.Model.get_config() against juju 2.6.6 model


afreiberger:
The following code works against juju==2.6.0, but just started failing today against juju==2.6.1.

It appears that the code that validates ConfigValues is now expecting value to be either a dict or a set, but does not allow for values to be strings.

I've reverted my testing environment to use juju==2.6.0 as a workaround.

Here is my testing traceback:

```
tests/functional/test_deploy.py:51: in get_http_proxy
    config = await model.get_config()
.tox/functional/lib/python3.6/site-packages/juju/model.py:1575: in get_config
    config[key] = ConfigValue.from_json(value)
.tox/functional/lib/python3.6/site-packages/juju/client/facade.py:614: in from_json
    return cls(**d)
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

```
self = <[AttributeError("'ConfigValue' object has no attribute 'source'") raised in repr()] ConfigValue object at 0x7f43eefd5f28>, source = 'default', value = '', unknown_fields = {}
source_ = 'default', value_ = ''

    def __init__(self, source=None, value=None, **unknown_fields):
        '''
        source : str
        value : typing.Mapping[str, typing.Any]
        '''
        source_ = source
        value_ = value
    
        # Validate arguments against known Juju API types.
        if source_ is not None and not isinstance(source_, (bytes, str)):
            raise Exception("Expected source_ to be a str, received: {}".format(type(source_)))
    
        if value_ is not None and not isinstance(value_, (dict, set)):
>           raise Exception("Expected value_ to be a Mapping, received: {}".format(type(value_)))
E           Exception: Expected value_ to be a Mapping, received: <class 'str'>

.tox/functional/lib/python3.6/site-packages/juju/client/_definitions.py:5194: Exception
```

```
$ juju status -m controller
Model       Controller  Cloud/Region         Version  SLA          Timestamp
controller  lxd         localhost/localhost  2.6.6    unsupported  19:04:21Z

[...]

PR #345, opened 2019-08-22
Handle interface{} params from api server

SimonRichardson:
The following commit changes how we handle `interface{}` from the
api server. Previously any `interface{}` was handled as a
`map[string]interface{}`, but that's not actually the case. In
reality the value returned could be anything. To support this change
we should not validate the value and just ignore it for the user to
handle correctly.


mitechie:
This ties in with #344

SimonRichardson:
$$merge$$

PR #350, opened 2019-08-27
Bundle changes

SimonRichardson:
The following adds a set of bundle changes classes as a way to
abstract the possible issue where breakages upstream cause chaos
downstream.

This is an idea and the real fix is to make sure that juju itself
correctly handles those changes.

The benefit of doing it this way is that we have similar set of
abstractions to the golang source, making it more familiar to all.

### DIFF
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -8,7 +8,6 @@ import yaml
 from toposort import toposort_flatten
 
 from .client import client
-from .constraints import normalize_key
 from .constraints import parse as parse_constraints
 from .errors import JujuError
 
@@ -51,8 +50,7 @@ class BundleHandler:
                            ConsumeOfferChange,
                            ExposeChange,
                            ScaleChange,
-                           SetAnnotationsChange,
-                          ]
+                           SetAnnotationsChange]
         self.change_types = {}
         for change_cls in change_type_cls:
             self.change_types[change_cls.method()] = change_cls
@@ -444,16 +442,16 @@ class AddApplicationChange(ChangeInfo):
                 self.num_units = params[9]
 
         elif isinstance(params, dict):
-            self.charm = params.charm
-            self.series = params.series
-            self.application = params.application
-            self.options = params.options
-            self.constraints = params.contstraints
-            self.storage = params.storage
-            self.devices = params.devices
-            self.endpoint_bindings = params.get("endpoint-bindings")
-            self.resources = params.resources
-            self.num_units = params.get("num-units")
+            self.charm = params["charm"]
+            self.series = params["series"]
+            self.application = params["application"]
+            self.options = params["options"]
+            self.constraints = params["constraints"]
+            self.storage = params["storage"]
+            self.devices = params["devices"]
+            self.endpoint_bindings = params["endpoint-bindings"]
+            self.resources = params["resources"]
+            self.num_units = params["num-units"]
         else:
             raise Exception("unexpected params type")
 
@@ -473,9 +471,10 @@ class AddApplicationChange(ChangeInfo):
             units_info = " with {num_units} unit{plural}".format(num_units=self.num_units,
                                                                  plural=plural)
         return "deploy application {application}{units_info}{series} using {charm}".format(application=self.application,
-                                                                                          units_info=units_info,
-                                                                                          series=series,
-                                                                                          charm=self.charm)
+                                                                                           units_info=units_info,
+                                                                                           series=series,
+                                                                                           charm=self.charm)
+
 
 class AddCharmChange(ChangeInfo):
     """
@@ -497,10 +496,10 @@ class AddCharmChange(ChangeInfo):
             else:
                 self.channel = None
         elif isinstance(params, dict):
-            self.charm = params.charm
-            self.series = params.series
-            if "channel" in params and params.channel != "":
-                self.channel = params.channel
+            self.charm = params["charm"]
+            self.series = params["series"]
+            if "channel" in params and params["channel"] != "":
+                self.channel = params["channel"]
             else:
                 self.channel = None
         else:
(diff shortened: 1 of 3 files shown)


SimonRichardson:
https://trello.com/c/uWtEG93I/1785-pylibjuju-bundle-changes-26x-can-run-against-27x

juju/bundle.py
```diff
@@ -464,3 +388,408 @@ def sorted(self):
...
+             'options': 'options',
+             'constraints': 'constraints',
+             'storage': 'storage',
+             'devices': 'devices',
+             'endpoint-bindings': 'endpoint_bindings',
+             'resources': 'resources',
+             'num-units': 'num_units'}
+
+    """
+    AddCharmChange holds a change for deploying a Juju application.
```
SimonRichardson:
```suggestion
    AddApplicationChange holds a change for deploying a Juju application.
```

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
                 reference = ref
         return reference
 
-    async def addCharm(self, change_id, charm, series):
+
+def get_charm_series(path):
```
mitechie: is this something that's bundle specific? Should it be in charm or a util section?

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+    """Inspects the charm directory at ``path`` and returns a default
+    series from its metadata.yaml (the first item in the 'series' list).
+
+    Returns None if no series can be determined.
+    """
+    md = Path(path) / "metadata.yaml"
+    if not md.exists():
+        return None
+    data = yaml.load(md.open())
```
mitechie: do we want to try to catch any errors here in the yaml?

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+    _toPy = {'charm': 'charm',
+             'series': 'series',
+             'application': 'application',
+             'options': 'options',
+             'constraints': 'constraints',
+             'storage': 'storage',
+             'devices': 'devices',
+             'endpoint-bindings': 'endpoint_bindings',
+             'resources': 'resources',
+             'num-units': 'num_units'}
```
mitechie: is this just here to map - to _ ? Can we just do that transition automatically in place? e.g. xxx.replace('-', '_') or xxx.standardize() or something?

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+
+    """
+    AddApplicationChange holds a change for deploying a Juju application.
+
+    :charm: holds the URL of the charm to be used to deploy this application.
```
mitechie: let's drop the "holds the" it's repeat noise imo

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+    @staticmethod
+    def method():
+        """
+        method returns an associated ID for the Juju API call.
+        """
+        return "deploy"
+
+    async def run(self, context):
+        """
+        Run executes a AddApplicationChange using the returned parameters from
```
mitechie: I'd drop "Run" and just leave it as "executes..." since the function name sets the context.

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+
+class AddCharmChange(ChangeInfo):
+    _toPy = {'charm': 'charm',
+             'series': 'series',
+             'channel': 'channel'}
+
+    """
+    AddCharmChange holds a change for adding a charm to the environment.
+
+    :charm: holds the URL of the charm to be added.
```
mitechie: change_id, requires, params

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
-                machines.
+    """
+    AddMachineChange holds a change for adding a machine or container.
 
-            parent_id: string holding a placeholder pointing to another
-                machine change or to a unit change. This value is only
-                specified in the case this machine is a container, in
-                which case also ContainerType is set.
+    :series: holds the optional machine OS series.
+    :constraints: holds the optional machine constraints.
```
mitechie: is this the list of optional params? or just commands in the method not true any more after changes?

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
+        elif isinstance(params, dict):
+            AddMachineChange.from_dict(self, params)
+        else:
+            raise Exception("unexpected params type")
 
+    @staticmethod
+    def method():
         """
-        params = params or {}
+        method returns an associated ID for the Juju API call.
```
mitechie: Convention is that the docstring starts after the """ vs on a newline:

https://www.python.org/dev/peps/pep-0257/#id17

juju/bundle.py
```diff
@@ -156,311 +171,663 @@ def resolve(self, reference):
...
-        return await self.model.add_relation(ep1, ep2)
 
-    async def deploy(self, change_id, charm, series, application, options, constraints,
-                     storage, endpoint_bindings, *args):
-        """
-        :param charm string:
-            Charm holds the URL of the charm to be used to deploy this
-            application.
+class AddRelationChange(ChangeInfo):
+    _toPy = {'endpoint1': 'endpoint1',
```
mitechie: why do we need _toPy here as well?

juju/bundle.py
```diff
@@ -129,12 +142,14 @@ def __init__(self, model, trusted=False, forced=False):
...
         for step in changes.sorted():
-            method = getattr(self, step.method)
-            result = await method(step.id_, *step.args)
-            self.references[step.id_] = result
+            change_cls = self.change_types.get(step.method)
+            if change_cls is None:
+                raise NotImplementedError("unknown change type: {}".format(step.method))
+            change = change_cls(step.id_, step.requires, step.args)
+            log.info("Applying change: {}".format(change))
+            self.references[step.id_] = await change.run(self)
```
mitechie: ooh, that reads nice

juju/bundle.py
SimonRichardson: I think this is just a bundle handler thing, was already existing in the codebase.

juju/bundle.py
SimonRichardson: Before closing out this PR, I wanted to address this comment. So the reason this exists is for the following reasons (although they might not be good enough reasons?):

 1. Conversion between what juju outputs and what pylib expects is somewhat different and the meshing of the different languages to types is somewhat problematic. Luckily it's a well trodden path in pylib and this is the same convention that's in the client facade code. So if you're hacking away at pylib it should be rather familiar to use this.
 2. We want an exact match of expected values, you don't get anything for free, so you are forced to add to the values. This has previously bitten us before, where we get some new values, unsure how to deal with them, so we either drop them or pass them on directly, either way ends up being wrong. So with this, we're saying what we want to pluck from the api response, anything else is assigned None.
 3. All values that aren't found in the api response are assigned None directly, without having to check if it exists or not. It removes a lot of if, else code.

I'm not sold on the name `_toPy` and we could change it to make more sense, but that's currently in the rest of the code base, so reuse and familiarity might be worth something?

SimonRichardson:
$$merge$$

PR #351, opened 2019-08-29
Bump pyyaml dependency version

battlemidget:
There is a CVE against pyyaml <5.1 and since pyyaml 4 was never actually released (only prereleases) anyone including this library would bring in the vulnerable pyyaml. This patch bumps it to the current blessed release and subsequently addresses the CVE https://nvd.nist.gov/vuln/detail/CVE-2017-18342


mitechie:
TY, going with that tests pass with the change and running with the change.

mitechie:
$$merge$$

PR #354, opened 2019-09-04
Bundle facade api call

SimonRichardson:
The following commit ensures that we use the latest bundle api facade
changes call. This cleans up the problem of what to use in terms of
bundle changes, shame pylib juju is now ahead of golang cli :D


SimonRichardson:
https://trello.com/c/NRsXYWNy/1800-update-pylibjuju-27-with-new-facade-changes

PR #355, opened 2019-09-05
Handle missing data from the API

SimonRichardson:
The following ensures that we handle missing data returned back
from the new API. Although the new API isn't turned on in this
branch, we should at least handle the case where it could be.

### DIFF
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -683,10 +683,10 @@ class CreateOfferChange(ChangeInfo):
     def __str__(self):
         endpoints = ""
         if self.endpoints is not None:
-            endpoints = self.endpoints.join(",")
-        return "create offer {offer_name} using {application}:{endpoints}".format(offer_name=self.offer_name,
-                                                                                  application=self.application,
-                                                                                  endpoints=endpoints)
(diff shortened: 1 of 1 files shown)


juju/bundle.py
```diff
@@ -681,9 +681,12 @@ def method():
             await context.model.create_offer(ep, offer_name=self.offer_name, application_name=application)
 
     def __str__(self):
+        endpoints = ""
+        if self.endpoints is not None:
+            endpoints = self.endpoints.join(",")
         return "create offer {offer_name} using {application}:{endpoints}".format(offer_name=self.offer_name,
                                                                                   application=self.application,
-                                                                                  endpoints=self.endpoints.join(","))
+                                                                                  endpoints=endpoints)
```
achilleasa: Small nitpick here. If endpoints is empty the returned string will end with a colon. How about if you changed L686 to read `":" + self.endpoints.join(",")` and remove the colon from L687?

SimonRichardson: Updated and fixed, good spot.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

mitechie:
$$merge$$

PR #360, opened 2019-10-17
Add cloud() and clouds() commands to controller.py

davigar15:
The cloud() and clouds() commands provide information about the clouds managed by a controller.

The cloud() method returns a ~CloudResult type object.
The clouds() method returns a ~CloudsResult type object.

### DIFF
--- a/examples/cloud.py
+++ b/examples/cloud.py
@@ -1,0 +1,28 @@
+"""
+This example:
+
+1. Connects to current controller.
+2. Gets the cloud
+3. Disconnects from the controller
+
+"""
+import logging
+
+from juju import loop
+from juju.controller import Controller
+
+
+async def main():
+    controller = Controller()
+    await controller.connect()
+
+    await controller.cloud()
+
+    await controller.disconnect()
+
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG)
+    ws_logger = logging.getLogger('websockets.protocol')
+    ws_logger.setLevel(logging.INFO)
+    loop.run(main())
--- a/examples/clouds.py
+++ b/examples/clouds.py
@@ -1,0 +1,28 @@
+"""
+This example:
+
+1. Connects to current controller.
+2. Gets all the clouds from a controller
+3. Disconnects from the controller
+
+"""
+import logging
+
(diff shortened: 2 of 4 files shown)


examples/cloud.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+async def main():
+    controller = Controller()
+    await controller.connect()
+
+    cloud = await controller.cloud()
+    print(cloud)
+
+    pdb.set_trace()
```
SimonRichardson: This should be removed, as the example should just output the cloud directly.

examples/clouds.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+async def main():
+    controller = Controller()
+    await controller.connect()
+
+    info = await controller.clouds()
+    print(info)
+
+    pdb.set_trace()
```
SimonRichardson: Remove (same as above).

examples/get_cloud.py
```diff
@@ -0,0 +1,29 @@
...
+from juju.controller import Controller
+
+
+async def main():
+    controller = Controller()
+    # connect to current controller with current user, per Juju CLI
+    await controller.connect()
+    cloud_name = await controller.get_cloud()
+    print(cloud_name)
+    import pdb; pdb.set_trace()
```
SimonRichardson: Remove (same as above)

juju/controller.py
```diff
@@ -437,6 +437,28 @@ def kill(self):
...
+        :param str name: Cloud name. If not specified, the cloud where
+                         the controller lives on is returned.
+        :returns: -> ~CloudResult
+        """
+        if name is None:
+            name = await self.get_cloud()
+        entity = client.Entity(tag.cloud(name))
+        cloud_facade = client.CloudFacade.from_connection(self.connection())
+        cloud = await cloud_facade.Cloud(entities=[entity])
+        return cloud.results[0]
```
SimonRichardson: We should check that the results length is valid, before trying to access the results. That way we can give better error messages to the user.

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #373, opened 2019-12-11
Fix all watcher example

SimonRichardson:
The example for the all watcher was out of date, so it would help if we
brought that up to speed, so others can see how to correctly use it.


SimonRichardson:
$$merge$$

Issue #377, opened 2020-01-07
Generating facades from schema fails with python 3.7


SimonRichardson:
Unfortunately python3.7 broke compatibility between python3.6 and python3.7 when introspecting types for generating the facades.

Attempting to updating it requires some significant changes to pylibjuju.

See:

```
➜ make client
.tox/py3/bin/python3.7 -m juju.client.facade -s "juju/client/schemas*" -o juju/client/
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/simon/Documents/python-libjuju/juju/client/facade.py", line 895, in <module>
    main()
  File "/home/simon/Documents/python-libjuju/juju/client/facade.py", line 885, in main
    definitions = generate_definitions(schemas)
  File "/home/simon/Documents/python-libjuju/juju/client/facade.py", line 824, in generate_definitions
    buildTypes(schema, definitions)
  File "/home/simon/Documents/python-libjuju/juju/client/facade.py", line 387, in buildTypes
    elif issubclass(arg_type, typing.Sequence):
  File "/usr/lib/python3.7/typing.py", line 719, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
TypeError: Subscripted generics cannot be used with class and instance checks
make: *** [Makefile:16: client] Error 1
```

SimonRichardson:
This was fixed by @hpidcock for #412

PR #378, opened 2020-01-07
Facade 2.7.0 version bumping

SimonRichardson:
The following commit changes the version of pylibjuju to focus on 2.7.0
as the tracking branch. In doing so, it requires bringing in the 2.7.0
schemas and updating the library to Juju 2.7.0.

Unfortunately we can't use python 3.7.x as there is an issue in
atttempting to introspect the types. It seems like there is a breakage
between 3.6.x and 3.7.x and some significant changes are required in
juju/client/facade (ones I don't fully understand atm) before we can use
that.

The client libraries where generated using a lxc container and
installing 3.6.x and using that as a way to bypass the error.

I've created a tracking issue[1] to resolve it.

 1. https://github.com/juju/python-libjuju/issues/377


SimonRichardson:
> I note that there are changes to standing versions of the facades that we would generally not anticipate based on our versioning, such as removal of methods from V2 of the HA facade. This is a factor of practice rather than logic, but we should be vigilant with this in order not to break backwards-compatibility.

The problem is that people are willing to change parameter names and not fully understanding the consequences downstream. Luckily the changes seem to not have any exposure in pylibjuju directly, but when they do (eventually) land in the library the people will have to be vigilant.

SimonRichardson:
$$merge$$

SimonRichardson:
!!build!!

Just seeing why we failed to pass the merge.

SimonRichardson:
$$merge$$

PR #379, opened 2020-01-08
Model connection failure

SimonRichardson:
The following fixes the issue where a new facade was introduced in 2.7,
but wasn't added to the library. Pylibjuju expects all new facades to
registered so it can negotiate how to speak to Juju. If a new facade is
introduced in to Juju and pylibjuju isn't aware of it, then
unfortunately it can't understand the whole picture.

The changes ensure that we're using the correct facade versions and that
we re-enable a the skipped test. The intermittent test is better to be
intermittent than to release a library release that's broken.

Unfortunately not all integration tests are run and there could be other
issues like that.

---

Additionally I noticed that a model delta type from the all watcher
isn't being handled, we should fix this in the future. For now it seems
safe enough to ignore it.

---

Along with the changes I've updated warnings about deprecated code.


juju/delta.py
```diff
@@ -108,5 +108,7 @@ def get_entity_class(self):
     'relation': RelationDelta,
     'remoteApplication': RemoteApplicationDelta,
     'charm': CharmDelta,
-    'applicationOffer': ApplicationOfferDelta,
+    'applicationOffer': ApplicationOfferDelta
+    # TODO (stickupkid): Currently ModelDelta is missing from the all watcher,
```
mitechie: Let's create an issue for this to reference please. I'm a bit -1 on TODO without a bug/issue to go with it in general.

SimonRichardson: https://github.com/juju/python-libjuju/issues/380

SimonRichardson:
$$merge$$

babbageclunk:
Hi guys - really sorry about the breakage. I didn't realise that pylibjuju needed to be updated for any new facade, even if it was an internal one. I'll make sure to check that next time.

ChrisMacNaughton:
Thanks for the super quick turnaround!

SimonRichardson:
@babbageclunk Turns out you found a bug - well played sir!

I've made a PR to fix this #382

PR #381, opened 2020-01-09
2.7.1 Release

SimonRichardson:
The is the version bumping for the 2.7.1 release.

It includes the new facade types that where missing when attempting to
connect to a model. Failure to have that facade type means that
pylibjuju doesn't know how to work with Juju.


SimonRichardson:
$$merge$$

PR #382, opened 2020-01-09
Unknown facade

SimonRichardson:
The following changes how we handle future compatibility. In an effort
to improve compatibility with Juju and it's clients. Previously the
client would attempt to negotiate every facade that Juju would have ever
released, which allows backwards compatibility but not forward.

With the following changes we're essentially unrestricting forward
compatibility. Allowing the client to speak to future Juju's even if the
facade version isn't known.

We warn the user as much as possible about the missing facade version,
before ignoring it and allowing accessing to be done.

In reality this means if you try and access a facade that doesn't exist
in the library, but does in Juju, it won't work.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -690,8 +690,10 @@ class Connection:
                 # client can define the non-conservitive facade client pinning.
                 if name in self.specified_facades:
                     known = self.specified_facades[name]['versions']
-                else:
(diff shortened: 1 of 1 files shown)


SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #383, opened 2020-01-10
SCP Example

SimonRichardson:
The following shows how to use scp from pylibjuju. This should help
others understand how scp works for machines. I'm unsure scp'ing from
containers will work, so investigation is required.


SimonRichardson:
$$merge$$

PR #390, opened 2020-02-24
Model summary watcher

howbazaar:
The biggest part of this diff is the update of the schemas to include a beta1 for Juju 2.8. This should be replaced by a final 2.8 as we release the code.

The watch_model_summaries method in juju/controller.py is the main entry point. The new model summary watcher facade is also added.

An example is provided to show it working.


howbazaar:
$$merge$$

PR #392, opened 2020-03-03
Fix bug 391: Archive symlinks to folders

davigar15:
Basically I'm checking if the dirnames are links or not. If they are, then I write them to the zip.

### DIFF
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -95,6 +95,25 @@ async def test_deploy_bundle(event_loop):
 
 @base.bootstrapped
 @pytest.mark.asyncio
+async def test_deploy_local_charm_folder_symlink(event_loop):
+    from pathlib import Path
+    tests_dir = Path(__file__).absolute().parent.parent
+    charm_path = tests_dir / 'charm-folder-symlink'
+
+    async with base.CleanModel() as model:
+        simple = await model.deploy(str(charm_path))
+        assert 'simple' in model.applications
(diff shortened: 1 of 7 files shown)


SimonRichardson:
@davigar15 can I get some tests for this, so we don't regress in the future.

tests/integration/test_model.py
```diff
@@ -93,6 +93,25 @@
             assert app in model.applications
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_deploy_local_charm_operator_framework(event_loop):
+    from pathlib import Path
```
SimonRichardson: If you remove this to the top of the file, we can drop the E402 changes to flake8?

davigar15:
That was meant for the arc/charm.py in the charm example.

We need to append the lib in order to import the objects in ops.

Since we do sys.append(‘lib’) before  actually importing the stuff in ops,
there’s no way we can avoid that addition in flake8.

David

SimonRichardson:
@davigar15 having more thought on this is, can we replicate this without the operator framework. Unfortunately, I think adding the operator framework is a blocker on this.

Additionally, we've got some 2.8 code in master that we don't want to release until later on in the cycle, so a release of pylibjuju isn't going to happen until then.

SimonRichardson:
!!build!!

davigar15:
You’re welcome 😇

SimonRichardson:
$$merge$$

PR #393, opened 2020-03-04
Add SSH support for units and machines

### LINKED ISSUE #79: Implement ssh
For purposes like Matrix's reboot and kill_juju_agent actions, "ssh" is probably less likely to error than "run".

ChrisMacNaughton:
Fixes #79

### DIFF
--- a/tests/integration/test_unit.py
+++ b/tests/integration/test_unit.py
@@ -101,6 +101,33 @@ async def test_scp(event_loop):
 
 @base.bootstrapped
 @pytest.mark.asyncio
+async def test_ssh(event_loop):
+    # ensure that asyncio.subprocess will work;
+    try:
+        asyncio.get_child_watcher().attach_loop(event_loop)
+    except RuntimeError:
+        pytest.skip('test_ssh will always fail outside of MainThread')
+    async with base.CleanModel() as model:
+        app = await model.deploy('ubuntu')
+
+        await asyncio.wait_for(
+            model.block_until(lambda: app.units),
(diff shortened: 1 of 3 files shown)


tests/integration/test_unit.py
```diff
@@ -99,6 +99,33 @@
...
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_ssh(event_loop):
+    # ensure that asyncio.subprocess will work;
+    try:
+        asyncio.get_child_watcher().attach_loop(event_loop)
+    except RuntimeError:
+        pytest.skip('test_scp will always fail outside of MainThread')
```
SimonRichardson:
```suggestion
        pytest.skip('test_ssh will always fail outside of MainThread')
```

johnsca:
[`provisioner.py`](https://github.com/juju/python-libjuju/blob/master/juju/provisioner.py) already contains some SSH logic using the Paramiko library.  I'm not entirely sure why that library was used instead of using an async subprocess like this PR does.  I wonder if that dependency could be removed by refactoring that code to use this approach?

ChrisMacNaughton:
@SimonRichardson `scp` already does shell out?

SimonRichardson:
$$merge$$

SimonRichardson:
@ChrisMacNaughton can you rebase to the latest master, our merge bot won't let this land.

SimonRichardson:
$$merge$$

hpidcock:
#422 fixes unit ssh async

PR #396, opened 2020-04-03
Machine manually provisioned

SimonRichardson:
To check if a machine is manually provisioned is a bit cryptic. To
better expose this, we can use the same method that juju does.


juju/machine.py
```diff
@@ -304,3 +304,9 @@ def series(self):
     @property
     def tag(self):
         return tag.machine(self.id)
+
+    @property
+    def is_manual(self):
+        if 'instance-id' in self.safe_data:
+            return self.safe_data['instance-id'].startswith('manual:')
+        return False
```
davigar15, line 312: Add one empty line below if you can :)

mitechie:
I wanted to check on why we were treating a manual machine differently in any way. Ideally you'd add the machine and then treat it like others. I'd like to chat on the use case and if there might be a cleaner path forward vs adding this in.

davigar15:
any updates?¿

petevg (approved):
Nice and simple. +1 :-)

Issue #394, opened 2020-03-06
Remove or double-down on Paramiko dependency for SSH


johnsca:
The [manual provisioning code](https://github.com/juju/python-libjuju/blob/master/juju/provisioner.py) uses the Paramiko library to invoke commands on the new machines via SSH, but #393 shows that it may be simpler to simply use async subprocesses to the command-line SSH client.  I think the original rationale was to enable support on systems that didn't have an ssh command-line client.  If that's still a concern, then we should convert the model code to use Paramiko instead, but if it's not, it would be great to be able to remove that dependency.

johnsca:
@davigar15 I believe you can shed some light on whether we need to be concerned about the potential lack of an ssh CLI client?

AdamIsrael:
@johnsca The only reason for Paramiko, that I can recall, was because that's what I was familiar with at the time. I remember that we talked briefly about using `asyncssh` at one point.

I think it's safe to assume -- or require -- the machine running the provisioner to have an ssh client installed, though.

juanmanuel-tirado:
Close. Follow comments on #399

PR #399, opened 2020-04-08
Replace paramiko with asyncssh

davigar15:
Related to issue #394


SimonRichardson:
The one issue I do have about this change is that this changes the API. @ChrisMacNaughton brought up a very good point that adding/removing async type of a def, changes the method that is being called.

I'm not in any way opposed to landing this (less dependencies + consistency though ssh), but probably not without bumping a major version to do so.

Any thoughts on this@johnsca, @AdamIsrael and @mitechie ?

juju/asyncprovisioner.py
```diff
@@ -0,0 +1,260 @@
+import os
+import re
+import shlex
+import tempfile
+import uuid
+from subprocess import CalledProcessError
+
+import asyncio
+import paramiko
```
johnsca, line 9: You're still importing this even though you removed it from the dependency list.

davigar15: @johnsca And if for simplicity I just replace the current one with the new one?

johnsca: I would much prefer to just drop the sync version entirely but it was @SimonRichardson's concern that it breaks an API that was potentially being used, though I strongly suspect that it was only ever used via the model method anyway, rather than being used directly. I lean toward just saying that it was an internal implementation detail and not worrying about the major version bump, but I do understand the concerns.

Thoughts, @SimonRichardson @ChrisMacNaughton?

ChrisMacNaughton: Honestly, I'd be tempted to minimally wrap the Async version into a Sync version that maintains the current API and logs a warning about the deprecation now, and then remove it entirely soonish

SimonRichardson: I agree with @ChrisMacNaughton on this one and by soonish, another major version of juju 2.9.

johnsca: Unfortunately, calling async code from sync code can only be done in asyncio if no other async code is running (that is, there is no existing event loop).  It seems likely that that if the existing sync implementation class has been used directly, it was likely from async code as in the current implementation of `add_machine`.  That means if we don't want to break it, we have to wrap in the other direction using `run_in_executor` and a thread, refactor the existing class to also use subprocess, or just leave the existing one alone and keep the paramiko dependency until we remove the deprecated version.

PR #401, opened 2020-04-12
remove travis since its been broken for ages

### LINKED ISSUE #400: travis ci broken?
every single commit in this repo for many moons and every single pr on this repo are failing status checks, which conveys a connotation of a project thats not really working. if your not using travis then remove the .travis.yml config from this repo or fix.

kapilt:
closes #400


SimonRichardson:
@kapilt so the real fix here is to send something to stdout/stderr to prevent travis thinking the job is dead. I think I want to try that before removing travis.

kapilt:
sure it could be fixed, but umm. whats the value of a ci thats been broken for months if no one is paying attention to it? if its not being used/looked at, just remove it. the other juju projects seemed to have moved on to jenkins or github actions as well. it also may not be apparent but at least for github users its pretty common to use status checks are a basic quality check on the project.

SimonRichardson:
@kapilt because I still find the Travis tests useful. Unfortunately I don't have the time to migrate the Travis test to Jenkins or fix them at the moment.

Close for now, reopen if you feel differently

kapilt:
given the travis tests don't get past install the few i looked at, the utility of them is pretty obscure ;-) like i said i think its helpful if you want users, to actually not have build failures labeled on every commit and pr in github, but to each their own. as the original author of this project, i did care once upon time, but moving on.

SimonRichardson:
@kapilt I'll happily take a pr that tries to move us forward, just they found an issue that wasn't covered by Jenkins. Compilation against different python versions, if we can cover that I'll happily drop Travis

kapilt:
@SimonRichardson perhaps your unaware, but the current Jenkins setup is antithetical to opensource imo, since the build results on contributions/prs are not available (even logged into ubuntu one), also given none of the build configuration assets are in the repo, its unclear how you expect contribution against them. I'd suggest just moving to GitHub Actions since travis has been shedding. The primary issue with travis seems to be using ppa's that aren't working re getting past install, using the travis versions of those would resolve. Alternatively GitHub Actions let you use arbitrary docker containers.

SimonRichardson:
> perhaps you're unaware, but the current Jenkins setup is antithetical...

I am very aware, but my hands are tied.

Issue #395, opened 2020-03-13
libjuju not prepared for HA usage


AdamIsrael:
In reviewing libjuju for it's readiness for use with Juju configured in high availability mode, it looks like there are some gaps.

https://github.com/juju/python-libjuju/blob/9e31ce1e79c9954ef226cb3076de66ddaafcb3ee/juju/client/connector.py#L91
The above implies that while libjuju receives a list of API endpoints it only uses the first endpoint. There's a TODO above this code that `Connection` needs to be modified to accept a list of endpoints, rather than a single endpoint.

Expected behavior:

We would expect libjuju to consume the list of API endpoints and failover if the selected address fails to connect.

PR #402, opened 2020-04-15
High availability support

davigar15:
This patch aims to add HA support in libjuju, and fix bug #395.

There are a few things that have to be implemented:
- Property in the controller to get the api_endpoints
- When the controller machine that libjuju is connected to goes down, reconnect internally to the other endpoints if available
- Notify the libjuju client when juju has scaled.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -230,7 +230,7 @@ class Connection:
         If uuid is None, the connection will be to the controller. Otherwise it
         will be to the model.
 
-        :param str endpoint: The hostname:port of the controller to connect to.
+        :param str endpoint: The hostname:port of the controller to connect to (or list of strings).
         :param str uuid: The model UUID to connect to (None for a
             controller-only connection).
         :param str username: The username for controller-local users (or None
@@ -256,6 +256,8 @@ class Connection:
         self = cls()
         if endpoint is None:
             raise ValueError('no endpoint provided')
+        if not isinstance(endpoint, str) and not isinstance(endpoint, list):
+            raise TypeError("Endpoint should be either str or list")
         self.uuid = uuid
         if bakery_client is None:
             bakery_client = httpbakery.Client()
@@ -279,6 +281,7 @@ class Connection:
         self.addr = None
         self.ws = None
         self.endpoint = None
+        self.endpoints = None
         self.cacert = None
         self.info = None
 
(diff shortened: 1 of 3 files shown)


davigar15:
This is work in progress.

davigar15:
Added api_endpoints property to controller

juju/client/connector.py
```diff
@@ -67,6 +67,13 @@ async def connect(self, **kwargs):
             for macaroon in kwargs.pop('macaroons'):
                 jar.set_cookie(go_to_py_cookie(macaroon))
         self._connection = await Connection.connect(**kwargs)
+        controller = self.jujudata.controllers()[
+            self.jujudata.current_controller()
+        ]
+        self._connection.endpoints = [
+            (e, controller["ca-cert"])
+            for e in controller["api-endpoints"]
+        ]
```
davigar15, line 76: @SimonRichardson I would like to set the endpoints within the Connection, but the problem is that if I get the controller from the connection, it enters into a loop.

If you come up with a better solution, let me know :-)

juju/client/connection.py
```diff
@@ -279,6 +281,7 @@ async def connect(
         self.addr = None
         self.ws = None
         self.endpoint = None
+        self.endpoints = None
```
SimonRichardson, line 284: I'm wondering if we can get rid of `self.endpoint` altogether and then say `self.endpoints = [endpoint]`. That way we reduce the number of `if isinstance(endpoint, str)` etc.

davigar15: I can do this: self.endpoints = [endpoint]

But removing the self.endpoint is more complicated, because that variable is used to by the class to know to which endpoint in particular it is connected to

davigar15: mmmmm but if I do [endpoint], and endpoint is a list, then self.endpoints will be [["endpoint1", ...]]

SimonRichardson: @davigar15 then we need to update `self.endpoint` to always be the one it's connected to. That way we're always in sync.

davigar15: @SimonRichardson That is updated in line 629, in the _connect function.

Every time libjuju reconnects, it executes the function _connect, and then this is executed:
         self.endpoint = result[2]

SimonRichardson: Excellent.

SimonRichardson:
$$merge$$

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #405, opened 2020-04-20
Fix Travis builds

johnsca:
Travis builds have been failing on the PPAs, and it's not clear that they're actually needed any longer.


johnsca:
Closing in favor of #415

PR #430, opened 2020-07-14
Derived application status

SimonRichardson:
The following updates pylibjuju to the same logic as 2.8.1, whereby the
derived application status that's in juju is ported across to libjuju.

It uses the same logic as juju[1], which can be found here. The code is
really simple and should work as intended to find out what the correct
application status is, if the application status has been set to
"unset".

1. https://github.com/juju/juju/commit/8bc76070c55c5991c0863e8f8c42df4780acdcbe#diff-7f5882040ea646fab639d5c6351b19f2R129

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -84,7 +84,7 @@ class Application(model.ModelEntity):
                 unit_status.append(unit.workload_status)
             return derive_status(unit_status)
 
-        return self.safe_data['status']['current']
+        return status
 
     @property
     def status_message(self):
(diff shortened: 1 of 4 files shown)


juju/application.py
```diff
@@ -76,6 +77,13 @@ def status(self):
...
 
         """
+        status = self.safe_data['status']['current']
+        if status == 'unset':
+            unit_status=[]
+            for unit in self.units:
+                unit_status.append(unit.workload_status)
+            return derive_status(unit_status)
+
         return self.safe_data['status']['current']
```
manadart, line 87: `return status`?

SimonRichardson:
$$merge$$

PR #432, opened 2020-07-14
Update to the latest juju 2.8.1 schema.

SimonRichardson:
The following updates the schema as described by[1]. Additionally with
the new schema changes, I've also wired up schema method description to
the output of the client methods. The methods should be now more user
friendly towards developers.

Additional work around bumping version numbers and releases.

1. https://discourse.juju.is/t/python-libjuju/1553


SimonRichardson:
$$merge$$

Issue #441, opened 2020-09-15
Application status not getting properly updated


davigar15:
Juju controller version: 2.8.2
Libjuju version: 2.8.3 (In 2.8.2 happens too)

Description of the issue: we get from the app.status a "waiting" when we should get blocked

Python script to reproduce the bug:
https://pastebin.canonical.com/p/gZKZmBn62b/

Output: https://pastebin.canonical.com/p/v4bk2TB58D/ 
^ We expect to retrieve a “blocked” string instead of “waiting”

SimonRichardson:
Unfortunately, the workaround until I can fully test all scenarios (see pastebin). The real problem is that I'm unsure how to handle application status for older jujus. Is it safe to always read the unit status and then derive that for even older jujus or do we have to do some facade checking? The problem with this is that I believe the changes didn't actually bump a facade, so it will be very difficult to identify when the change happened.

https://pastebin.canonical.com/p/kJ7jKmSRP8/

davigar15:
It's happening with this charm: cs:~davigar15/simple-native-0

PR #442, opened 2020-09-16
Derive the application status

SimonRichardson:
With changes to juju, the application status was changed for the CLI to
use the model cache to derive the status of the application from the
unit status. Unfortunately, the logic didn't reach the all watcher and in
doing so always reports the last application status.

Unfortunately, the facade version wasn't bumped, which makes facade
checking for old and new logic almost impossible. Instead of the solution
here is to always query the unit status, BUT add the application status
in to the derive_status() call so that we always bubble the most relevant
severity to the top.

This should happily work in a backward compatibility way, so if the
application is blocked and any units are waiting, then the application
status will always win. Similarly, if the units are blocked and the
application status is waiting, then blocked will be shown.

The one caveat is, that if the older juju didn't bubble up the worst
severity to the application status, then you could end up with a
mismatch. For example; the unit sets to error, but the application status
is waiting. Then juju CLI would show waiting, but pylibjuju would show
an error.

I'm unsure if that is actually a fundamental problem, as I don't believe
it should happen, but in theory, it should be at least known.

Fixes: #441

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -74,17 +74,16 @@ class Application(model.ModelEntity):
 
     @property
     def status(self):
-        """Get the application status, as set by the charm's leader.
+        """Get the application status.
 
+        If the application is unknown it will attempt to derive the unit
+        workload status and highlight the most relevant (severity).
         """
         status = self.safe_data['status']['current']
-        if status == 'unset':
-            unit_status = []
-            for unit in self.units:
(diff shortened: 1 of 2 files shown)


manadart (approved):
We should be able to cook a test for this right?

SimonRichardson:
$$merge$$

PR #444, opened 2020-09-17
Handle application set status

SimonRichardson:
Because pylibjuju has an internal cache, very similar to how juju model
cache works, we need to correctly handle when the application status is
set by the charm author, vs how the application is dervied via the unit
status.


manadart (approved):
Great.

SimonRichardson:
$$merge$$

PR #449, opened 2020-10-07
Support expose_endpoints parameter for applications

achilleasa:
This PR adds support for granular, per-endpoint expose parameters to the expose/unexpose methods of the Application model as well as an example for using the new feature. For more details on using this feature please refer to https://discourse.juju.is/t/granular-control-of-application-expose-parameters-in-the-upcoming-2-9-juju-release/3597

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -34,6 +34,9 @@ class Application(model.ModelEntity):
     def _facade(self):
         return client.ApplicationFacade.from_connection(self.connection)
 
+    def _facade_version(self):
+        return client.ApplicationFacade.best_facade_version(self.connection)
+
     def on_unit_add(self, callable_):
         """Add a "unit added" observer to this entity, which will be called
         whenever a unit is added to this application.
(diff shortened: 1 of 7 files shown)


juju/application.py
```diff
@@ -31,6 +31,12 @@ class Application(model.ModelEntity):
     def _unit_match_pattern(self):
         return r'^{}.*$'.format(self.entity_id)
 
+    def _facade(self):
+        return client.ApplicationFacade.from_connection(self.connection)
```
SimonRichardson, line 35: Yes! Don't know why this was never done previously?

:100:

juju/application.py
```diff
@@ -222,30 +228,129 @@ async def destroy(self):
...
+                raise ValueError("endpoints must be a dictionary with ExposedEndpoint values")
+
+            # The bundle changes code will pass in raw dicts with the exposed
+            # endpoint data. We need to convert those into ExposedEndpoints
+            for k, v in exposed_endpoints.items():
+                if not isinstance(v, ExposedEndpoint):
+                    exposed_endpoints[k] = ExposedEndpoint.from_dict(v)
+
+            # Check if the specified exposed_endpoints would cause security
+            # issues when applied to a pre 2.9 controller.
```
SimonRichardson, line 270: Is this done client-side in Juju CLI as well?

juju/client/connection.py
```diff
@@ -23,7 +23,7 @@
     'Agent': {'versions': [2]},
     'AgentTools': {'versions': [1]},
     'Annotations': {'versions': [2]},
-    'Application': {'versions': [1, 2, 3, 4, 5, 6, 7, 8]},
+    'Application': {'versions': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]},
```
SimonRichardson, line 26: Can't see anything going wrong here :face_with_head_bandage: :thinking:

SimonRichardson (changes requested):
Can we also add an integration test for expose/unexpose...?

juju/application.py
achilleasa: Yes, both for [expose](https://github.com/juju/juju/blob/develop/api/application/client.go#L865) and [unexpose](https://github.com/juju/juju/blob/develop/api/application/client.go#L902). The contents of the error messages also match so we are consistent.

achilleasa:
@SimonRichardson Also note that while I have implemented support for endpoints in the relevant expose change operation (for bundles), this feature will not work as the python bits do not handle multi-doc yaml bunles. We should probably fix that at some point...

SimonRichardson (approved):

achilleasa:
$$merge$$

PR #451, opened 2020-10-23
Handle strict mode

SimonRichardson:
Strict mode is a new feature in Juju 2.9. The mode allows more things to
fail early if the mode is switched to strict. If no mode is detected
then it will fallback to a more passive mode, where it will fallback to
other solutions and try to carry on.

The purpose of strict mode is to enable people developing with
Juju/pylibjuju is highlight errors more early on and to prevent less
magic in Juju.

The PR for implementing strict mode is simple, EXCEPT when I turned it
one, it found the first error! ModelInfo delta was wrongly implemented
and wasn't gaining all the history correctly from the deltas over time.

This requires[1] juju/juju#12183 to land before this PR is really
useful.

To enable it at bootstrap time:

```sh
juju bootstrap lxd test --config mode=strict
```

 1. https://github.com/juju/juju/pull/12183

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -301,6 +301,12 @@ class ModelEntity:
         'application' or 'unit', etc.
 
         """
+        # Allow the overriding of entity names from the type instead of from
+        # the class name. Useful because Model and ModelInfo clash and we really
+        # want ModelInfo to be called Model.
+        if hasattr(self.__class__, "entity_name") and callable(self.__class__.entity_name):
+            return self.__class__.entity_name()
+
         def first_lower(s):
             if len(s) == 0:
                 return s
@@ -449,6 +455,7 @@ class Model:
         self._observers = weakref.WeakValueDictionary()
         self.state = ModelState(self)
         self._info = None
+        self._mode = None
         self._watch_stopping = asyncio.Event(loop=self._connector.loop)
         self._watch_stopped = asyncio.Event(loop=self._connector.loop)
         self._watch_received = asyncio.Event(loop=self._connector.loop)
@@ -797,6 +804,10 @@ class Model:
         """
         return self._info
 
+    @property
+    def strict_mode(self):
+        return self._mode is not None and "strict" in self._mode
+
     def add_observer(
             self, callable_, entity_type=None, action=None, entity_id=None,
             predicate=None):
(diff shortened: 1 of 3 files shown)


juju/model.py
```diff
@@ -301,6 +301,12 @@ def entity_type(self):
         'application' or 'unit', etc.
 
         """
+        # Allow the overriding of entity names from the type instead of from
+        # the class name. Useful because Model and ModelInfo clash and we really
+        # want ModelInfo to be called Model.
+        if hasattr(self.__class__, "entity_name") and callable(self.__class__.entity_name):
+            return self.__class__.entity_name()
+
```
petevg, line 309: I'd rather see this called "type_override" or something similar. I like the principle, but the name is confusing.

juju/model.py
```diff
@@ -893,16 +918,15 @@ async def _all_watcher():
...
-                            log.warning("unknown delta type: %s", e.args[0])
+                            entity = get_entity_delta(delta)
+                            old_obj, new_obj = self.state.apply_delta(entity)
+                            await self._notify_observers(entity, old_obj, new_obj)
+                            # Post step ensure that we can handle any settings
+                            # that need to be correctly set as a post step.
+                            _post_step(new_obj)
+                        except KeyError:
+                            if self.strict_mode:
+                                raise JujuError("unknown delta type '{}'".format(delta.entity))
```
petevg, line 929: Which call does the KeyError come from?

An anti-pattern in Python is to take a low level error like KeyError thrown by a called function, and interpret it in the calling function. There might be other causes for KeyError, and it would be better to catch the error closer to the source, where we're more certain of the cause.

(This isn't a hill that I'm going to die on -- this might be a place where the pattern is not risky -- it just jumped out to me as I was reading the code.)

petevg (changes requested):
One change request, and one mild code smell that may or may not need to be addressed.

juju/model.py
SimonRichardson: Sure, I don't mind that.

juju/model.py
SimonRichardson: I didn't write this originally, so I'm more than happy to try and fix this.

petevg (approved):
LGTM.

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #456, opened 2021-01-04
Update to the latest 2.9 schema

SimonRichardson:
The following updates the 2.9 branch to the latest schema. The generated
python is the fallout from this.


juju/client/_client11.py
```diff
@@ -3697,27 +3692,6 @@ async def SetSupportedContainers(self, params=None):
 
 
 
-    @ReturnMapping(StringsResult)
-    async def StateAddresses(self):
```
SimonRichardson, line 3701: This is a bit scary...

achilleasa: Why is that?

achilleasa (approved):

juju/client/_client11.py
SimonRichardson: As we shouldn't be removing methods from the API server i.e. it's always additive. 

I believe this is ok, as it's a change @wallyworld did https://github.com/juju/juju/pull/12184

SimonRichardson:
$$merge$$

PR #457, opened 2021-01-04
Ensure all charm-store URLs have prefix

SimonRichardson:
The following PR ensures that all the URLs have the correct prefix, that
way when we come to implement the charm-hub deployment workflow we do
that without breaking the old tests.

Eventually the cs: (charm-store) will be removed, but whilst we're in
transition, we should still support it and the tests associated with
them.

The code is very mechanical and should show that a cs: prefix is
mandatory if the user wants to contact the charm-store.


achilleasa (approved):
Looking forward to the PR for deploying with charmhub

SimonRichardson:
$$merge$$

PR #460, opened 2021-01-05
CharmHub - Info

SimonRichardson:
The following introduces queries to the charmhub store API. Now that all
requests go through a juju controller it makes it very trivial for other
clients like pylibjuju to gain the same powers as the go CLI.

The code is rather simple, it farms off a new type to encapsulate the
charmhub related calls (info and find which will follow) as the model
type is large and unwieldly, once there we just create a new facade and
perform the query.

Offically it would be nice to have to channel validation, but for now
we'll let the controller handle that.

### DIFF
--- a/examples/charmhub.py
+++ b/examples/charmhub.py
@@ -1,0 +1,29 @@
+"""
+Example to show how to connect to the current model and query the charm-hub
+store for information about a given charm.
+"""
+import logging
+
+from juju import loop
+from juju.model import Model
+
+log = logging.getLogger(__name__)
+
+
+async def main():
+    model = Model()
+    try:
+        # connect to the current model with the current user, per the Juju CLI
+        await model.connect()
+
+        charm = await model.charmhub.info("mattermost")
+        print(charm)
+    finally:
+        if model.is_connected():
+            print('Disconnecting from model')
+            await model.disconnect()
+
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.INFO)
+    loop.run(main())
--- a/juju/charmhub.py
+++ b/juju/charmhub.py
@@ -1,0 +1,24 @@
+from .client import client
+from .errors import JujuError
+
+
+class CharmHub:
+    def __init__(self, model):
+        self.model = model
+
+    async def info(self, name, channel=None):
+        """info displays detailed information about a CharmHub charm. The charm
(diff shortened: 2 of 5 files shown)


juju/charmhub.py
```diff
@@ -0,0 +1,24 @@
...
+from .errors import JujuError
+
+
+class CharmHub:
+    def __init__(self, model):
+        self.model = model
+
+    async def info(self, name, channel=None):
+        """info displays detailed information about a CharmHub charm. The charm
+        can be specified by name or by path.
```
hmlanigan, line 11: What does path mean in this context? The first thing I think of is a local charm

juju/charmhub.py
```diff
@@ -0,0 +1,24 @@
...
+
+        """
+        if not name:
+            raise JujuError("name expected")
+
+        if channel is None:
+            channel = "latest"
```
hmlanigan, line 18: A channel cannot be a track only.  It'll fail at some point on the controller.  I'd prefer leaving channel blank if unspecified.  The caller won't know which channels are available.

juju/model.py
```diff
@@ -776,6 +777,14 @@ def relations(self):
         """
         return list(self.state.relations.values())
 
+    @property
+    def charmhub(self):
+        """Return a charmhub store for requesting charm information using
```
hmlanigan, line 782: nit: use of the word "store".  It's overused within juju land.  Repository?

examples/charmhub.py
```diff
@@ -0,0 +1,29 @@
+"""
+Example to show how to connect to the current model and query the charm-hub
+store for information about a given charm.
```
hmlanigan, line 3: nit: use of the word "store". It's overused within juju land. Repository?

hmlanigan (commented):
Reviewed from a charmhub perspective.

juju/charmhub.py
SimonRichardson: Yes it can, channels are normalised from `latest/stable` to `latest`...

juju/charmhub.py
SimonRichardson: I robbed it from Juju, I'll fix

petevg (approved):
Nice and simple. +1, LGTM.

SimonRichardson:
$$merge$$

PR #462, opened 2021-01-06
CharmHub - Find

SimonRichardson:
This requires #460 to land first.

The following adds the ability to query the charmhub repository. It uses
the controller to query the charmhub repository so is really easy for
clients to also use the same logic/code path as the go CLI.

PR #463, opened 2021-01-06
CharmHub - Origin

SimonRichardson:
Requires #460 and #462 to land first

The following starts to implement charm origin. These are required for
the deploy commands where we're talking about origin, channels and
platforms. If we make types of these then we can pass them around
knowing exactly the correct types at hand.

The code is pretty much a copy and pasted from juju itself with some
modifications to channels that don't take into account branches.

### DIFF
--- a/juju/origin.py
+++ b/juju/origin.py
@@ -1,0 +1,195 @@
+from enum import Enum
+from .errors import JujuError
+
+
+class Source(Enum):
+    """Source defines a origin source. Providing a hint to the controller about
+    what the charm identity is from the URL and origin source.
+
+    """
+    LOCAL = "local"
+    CHARM_STORE = "charm-store"
+    CHARM_HUB = "charm-hub"
+
+    def __str__(self):
+        return self.value
+
+
+class Origin:
+    def __init__(self, source, channel, platform):
+        self.source = source
+        self.channel = channel
+        self.platform = platform
+
+    def __str__(self):
+        return "origin using source {} for channel {} and platform {}".format(str(self.source), self.channel, self.platform)
+
+
+class Risk(Enum):
+    STABLE = "stable"
+    CANDIDATE = "candidate"
+    BETA = "beta"
+    EDGE = "edge"
+
+    def __str__(self):
+        return self.value
+
+    @staticmethod
+    def valid(potential):
+        for risk in [Risk.STABLE, Risk.CANDIDATE, Risk.BETA, Risk.EDGE]:
+            if str(risk) == potential:
+                return True
+        return False
+
+
+class Channel:
+    """Channel identifies and describes completely a store channel.
+
+     A channel consists of, and is subdivided by, tracks, risk-levels and
+      - Tracks enable snap developers to publish multiple supported releases of
+        their application under the same snap name.
+      - Risk-levels represent a progressive potential trade-off between stability
+        and new features.
+
+     The complete channel name can be structured as three distinct parts separated
+     by slashes:
+
+        <track>/<risk>
+
+    """
+    def __init__(self, track=None, risk=None):
+        if not Risk.valid(risk):
+            raise JujuError("unexpected risk {}".format(risk))
+
+        if track is None:
+            track = ""
+        self.track = track
+        self.risk = risk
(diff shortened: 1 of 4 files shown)


juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+
+     The complete channel name can be structured as three distinct parts separated
+     by slashes:
+
+        <track>/<risk>
+
+    """
+    def __init__(self, track=None, risk=None):
+        if not Risk.valid(risk):
+            raise JujuError("unexpected risk {}".format(risk))
```
petevg, line 62: I think that it's probably be more conventional here to raise the error in the validation method. You're going to raise it anyway, and this leaves room to write an @valid_risk wrapper for functions.

None of that is absolutely necessary. But there is a more compact way to do things.

juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if track is None:
+            track = ""
+        self.track = track
```
petevg, line 66: You can shorten the above to:

```
self.track = track or ""
```

There's nothing wrong or really less efficient about the code as written, though. I'm just back seat golfing :-)

juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+        """parse a channel from a given string.
+        Parse does not take into account branches.
+
+        """
+        if not s:
+            raise JujuError("channel cannot be empty")
+
+        p = s.split("/")
+
+        print(len(p))
```
petevg, line 80: This should be a logging statement or be removed (looks like leftover from debugging.)

juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+        track = None
+        if len(p) == 1:
+            if Risk.valid(p[0]):
+                risk = p[0]
+            else:
+                track = p[0]
+                risk = str(Risk.STABLE)
+        elif len(p) == 2:
+            track = p[0]
+            risk = p[1]
```
petevg, line 92: No harm in validating the risk here, as long as you have your validator. :-)

juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+
+        p = s.split("/")
+
+        print(len(p))
+
+        risk = None
+        track = None
+        if len(p) == 1:
+            if Risk.valid(p[0]):
+                risk = p[0]
```
petevg, line 86: Aha. Now I understand why you don't raise the error in the risk. Disregard my above comment. It would be a lot messier to do a try here -- the validator expects to get invalid risks without complaining.

juju/origin.py
```diff
@@ -0,0 +1,195 @@
...
+            series = p[1]
+        elif len(p) == 3:
+            arch = p[0]
+            os = p[1]
+            series = p[2]
+        else:
+            raise JujuError("platform is malformed and has too many components {}".format(s))
+
+        if arch is None or arch == "":
+            raise JujuError("architecture in platform {} is not valid".format(s))
```
petevg, line 167: I'd simplify this to:

```
if not arch:  # always required
    raise ...
```

That clearly calls the logic out as different from the two checks below, and makes it a little bit easier to parse for me.

petevg (changes requested):
Requesting some minor changes. Overall, this looks good, and I don't see any large flaws in the logic or structure.

juju/origin.py
SimonRichardson: This would only work for `kwargs` though right, as there is no way to know via `args`? So in order to use the wrapper you'd have to ensure that?

```py
def valid_risk(func):
    def wrap(*args, **kwargs):
        if "risk" in kwargs and not Risk.valid(kwargs["risk"]):
            raise "unexpected risk {}".format(kwargs["risk"])
        return func(*args, **kwargs)
    return wrap
```

petevg: You could do it with args, which are always in the same order.

Regardless, this is the way it is for other reasons, so I'm good with it :-)

petevg (approved):
Thanks for addressing the comments! This is good to merge, I think.

SimonRichardson:
$$merge$$

PR #464, opened 2021-01-08
CharmHub - Charm URL

SimonRichardson:
Requires #463 to land, then rebase this on top.

The following introduces a charm URL, this will help when trying to
ensure that what we get from the API is indeed a valid URL. The code is
pretty much a lift from the juju/charm package, minus the FQDN charm
URLs, which should never have been added to the charm package.

The code is rather procedural and doesn't validate the name or series,
but the controller can do that once we know the pattern is correct.

Essentially we want the charm URL for the schema and the name, so rather
than have methods that sort of do the job, we should replicate the exact
same parsing layout so we don't get it wrong.

### DIFF
--- a/juju/url.py
+++ b/juju/url.py
@@ -1,0 +1,136 @@
+from enum import Enum
+from .errors import JujuError
+from urllib.parse import urlparse
+
+
+class Schema(Enum):
+    LOCAL = "local"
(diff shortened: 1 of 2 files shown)


petevg (approved):
LGTM.

SimonRichardson:
$$merge$$

PR #470, opened 2021-02-04
Fix bundles referencing .charm files and add wait_for_bundle

johnsca:
Bundles which referenced local `.charm` files were failing.

This also adds a `Model.wait_for_bundle()` helper, which waits for all of the units deployed by the bundle to come up and for all hooks to settle. It is not as powerful as `juju-wait`, but it at least provides an easy way to block until a deployed bundle has settled into a steady state.


tests/integration/test_model.py
```diff
@@ -60,6 +60,22 @@ async def test_deploy_local_bundle_file(event_loop):
                           timeout=60 * 4)
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_deploy_bundle_local_charms(event_loop):
```
wallyworld, line 65: I can't see a test for the .charm or .zip scenario?

juju/model.py
```diff
@@ -2144,6 +2146,91 @@ async def _get_source_api(self, url, controller_name=None):
         await controller.connect(controller_name=controller_name)
         return controller
 
+    async def wait_for_bundle(self, bundle_path, raise_on_error=True,
```
wallyworld, line 2149: This seems reasonable to help tests be better written. I wonder, given its narrow focus - just on a bundle, not a replacement for juju wait etc - whether it should be put in a test helper package initially?

SimonRichardson: I have to agree with @wallyworld here... I think we should iterate this internally in tests for now so that becomes more generic.

Interestingly using the model cache that is backed by the `AllWatcher` will be far superior to how `juju wait` frequently calls the status of juju. Juju status currently requires a database trip to the database for all the data (there are a few exceptions, which hit the internal model cache) and that puts pressure on the database especially when you don't want to put pressure on the database; during a deployment. The `AllWatcher` returns smaller deltas of changes which prevents the mining of superfluous information.

Juju is also [gaining powers](https://github.com/juju/juju/tree/develop/cmd/plugins/juju-wait-for) to wait for certain conditions to be resolved.

tests/integration/test_model.py
johnsca: The file is [here in the diff](https://github.com/juju/python-libjuju/pull/470/files#diff-5104c1c0a71e77dc6ad34c63eb0723ae5c4e272c9c5f75aa656eec2292b9e837) and it's referenced [here in the bundle](https://github.com/juju/python-libjuju/pull/470/files#diff-54b5b6ff5ea16d0a1e4ea85dbce3bcc3bc29a7f0e09179804da5787e7892c9c0) which is in turned [used in the test you commented on](https://github.com/juju/python-libjuju/pull/470/files#diff-9405e0bc34a26ab57891171003bb7253b44c20fa921670f44b207c1971b413f1R67).  So it's testing bundle which uses both a local charm dir and a local charm file.

juju/model.py
johnsca: I might change it to `wait_for_idle` and then move `wait_for_bundle` out to `utils` as a wrapper around `wait_for_idle` which also checks that all of the desired apps and units are there.

juju/model.py
```diff
@@ -2144,6 +2146,91 @@ async def _get_source_api(self, url, controller_name=None):
...
+                    if unit.agent_status == "error":
+                        errors.setdefault("Agent", []).append(unit.name)
+                        continue
+                    if raise_on_error and unit.workload_status == "error":
+                        errors.setdefault("Unit", []).append(unit.name)
+                        continue
+                    if unit.agent_status == "idle":
+                        now = datetime.now()
+                        idle_start = idle_times.setdefault(unit.name, now)
+                        print(f"Unit {unit.name} is idle for {now - idle_start}")
```
johnsca, line 2217: Oops, errant pront.

juju/model.py
```diff
@@ -2144,6 +2146,89 @@ async def _get_source_api(self, url, controller_name=None):
...
+
+        while True:
+            busy_apps = []
+            errors = {}
+            for app in apps:
+                if app not in self.applications:
+                    busy_apps.append(app)
+                    continue
+                app = self.applications[app]
+                if raise_on_error and app.status == "error":
```
SimonRichardson, line 2202: On most new controllers this will probably just be `unset` unless the charm explicitly sets the status, so I'm glad we also check the units as well :+1:

SimonRichardson (approved):

SimonRichardson:
$$merge$$

PR #475, opened 2021-03-12
Introduce hostname property to Machine model

### LINKED ISSUE #474: Add hostname attribute to machine object
With https://github.com/juju/juju/pull/12754 being merged, and in 2.8.10, I'd like to access this new information via libjuju.  It would be great if we could have this available via machine.hostname, same as we can have machine.dns_name now.

achilleasa:
This allows users to access the hostname value reported by the machine
agent when it starts. This feature only works on 2.8.10+ controllers; in
all other cases (older controllers, machines with pending agents), the
property returns None.

Fixes #474 and works in conjunction with the changes from https://github.com/juju/juju/pull/12760

### DIFF
--- a/examples/machine_hostname.py
+++ b/examples/machine_hostname.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3.5
+
+"""
+This example:
+
+1. Connects to the current model
+2. Creates a machine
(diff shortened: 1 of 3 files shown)


SimonRichardson (approved):

achilleasa:
$$merge$$

Issue #477, opened 2021-03-15
Scaling K8s instance with python-libjuju not working


eminaktas:
#### Steps to reproduce 
1. Deploy a K8s instance with Juju
2. Connect the model
3. Get the application out of model
4. Call scale() function in Application module
```python3
from juju import loop
from juju.model import Model


async def _scale():

    application_name = "nginx"

    model = Model()
    await model.connect()
    if model.applications and application_name in model.applications:
        application = model.applications[application_name]

    await application.scale(scale=3)

    await model.disconnect()


if __name__ == '__main__':
    loop.run(_scale())
```

```bash
$ pip list | grep juju
juju                2.8.4
jujubundlelib       0.5.6
```

#### Current behaviour with python-libjuju
Returns this error
```error
Traceback (most recent call last):
  File "/home/ubuntu/tmpdir/module/N2VC/n2vc/deneme.py", line 20, in <module>
    loop.run(_scale())
  File "/home/ubuntu/.local/lib/python3.6/site-packages/juju/loop.py", line 38, in run
    raise task.exception()
  File "/home/ubuntu/tmpdir/module/N2VC/n2vc/deneme.py", line 14, in _scale
    await application.scale(scale=3)
  File "/home/ubuntu/.local/lib/python3.6/site-packages/juju/application.py", line 165, in scale
    client.ScaleApplicationParam(application_tag=self.tag,
AttributeError: module 'juju.client.client' has no attribute 'ScaleApplicationParam'

Process finished with exit code 1
```

#### Current behaviour with cli and Expected behaviour with python-libjuju
Before scaling application
```bash
$ juju status -m native-kdu-a05225f3-d946-4c09-a9ba-88df0fc7ca9b
Model                                            Controller  Cloud/Region                                  Version  SLA          Timestamp
native-kdu-a05225f3-d946-4c09-a9ba-88df0fc7ca9b  osm         d84a81e4-2631-41d2-bc3f-ac7b40a8de4a/default  2.8.6    unsupported  07:29:58Z

App    Version  Status  Scale  Charm      Store  Rev  OS          Address  Notes
nginx  nginx    active      1  nginx-k8s  local    0  kubernetes

Unit      Workload  Agent  Address       Ports   Message
nginx/0*  active    idle   10.1.128.225  80/TCP  Ready
```
Juju command to scale application (K8s)
```bash
$ juju scale-application nginx 2 -m native-kdu-a05225f3-d946-4c09-a9ba-88df0fc7ca9b
[...]

PR #478, opened 2021-03-15
Fix the typo in the param name

SimonRichardson:
The name of the param is ScaleApplicationParams not
ScaleApplicationParam. This should have really be caught in a test, but
for now, fix this so we can get a release out.

Fixes: #477


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #482, opened 2021-03-31
Find parameters

SimonRichardson:
The following adds the optional find parameters that are exposed in the
API. The find API already existed, this is just to augment what was
already there.

It's now possible to just request bundles.

```py
result = await model.charmhub.find("kuber", charm_type="bundle")
```


manadart (approved):
Works nicely.

SimonRichardson:
$$merge$$

PR #483, opened 2021-03-31
Charmhub deploy charm

SimonRichardson:
This PR allows the deploying of charmhub charms and bundles.

There are a few gaps in this PR and those will be filled in future PRs.

 1. resources (oci-images, local, etc)
 2. refresh/upgrade-charm
 3. k8s testing

The replicates as much as the juju go CLI. Although there are gaps in how
we deal with series as we currently don't have as much information as the go
CLI, but that is coming.

## Q&A Steps

Run the examples for bundles and try and deploy a charmhub charm.


SimonRichardson:
!!build!!

Makefile
```diff
@@ -20,8 +20,11 @@ client:
 	$(PY) -m juju.client.facade -s "juju/client/schemas*" -o juju/client/
 
 .PHONY: test
-test:
-	tox
+test: lint
+	tox -e py3
+	@for f in $(shell find tests/integration -maxdepth 1 -mindepth 1 -name "*.py" | grep -v "__init__.py"); do \
+		tox -e integration -- "$${f}"; \
+	done
```
petevg, line 27: I would rather see this done in the tox.ini file. There are some examples here: https://tox.readthedocs.io/en/latest/example/pytest.html

The point of tox is to make this sort of invocation unnecessary :-)

examples/deploy_bundle.py
```diff
@@ -6,35 +6,41 @@
...
-            await application.remove()
-    finally:
-        print('Disconnecting from model')
-        await model.disconnect()
-        print("Success")
+    controller = Controller()
+    # connect to current controller with current user, per Juju CLI
+    await controller.connect()
+
+    bundles = [('cs:~juju-qa/bundle/basic-0', 'beta'), ('juju-qa-bundle-test', None)]
```
petevg, line 18: This works, but I'd like to see it better organized. Instead of stuffing everything into main, I'd like to see something like:

```
async def deploy_bundle(name, risk):
    ...

async def main():
    ...
    # Deploy bundle
    await deploy_bundle('juju-qa-bundle-test', None)

    # Deploy legacy bundle
    await deploy_bundle('cs:...', 'beta')
```

It becomes clear after some inspection that this is what you're doing. But it's not clear at first glance. And I think that we should be careful to make the code in examples very, very clear.

juju/bundle.py
```diff
@@ -156,7 +177,100 @@ async def fetch_plan(self, entity_id):
         if self.plan.errors:
             raise JujuError(self.plan.errors)
 
+    async def _download_bundle(self, charm_url, origin):
+        if self.charms_facade is not None:
```
petevg, line 181: This would be cleaner if you did:

```
if self.charms_facade is None:
    raise ...

resp = await self.charms_facade ...
```

With Python, I like to see code that uses every opportunity to avoid indenting. And also it's nice to have the Exceptions thrown as close as possible, visually, to the check that causes it.

juju/bundle.py
```diff
@@ -422,15 +565,47 @@ async def run(self, context):
...
+        if Schema.LOCAL.matches(url.schema):
+            origin = client.CharmOrigin(source="local", risk="stable")
+            context.origins[self.charm] = {str(None): origin}
             return self.charm
 
-        entity_id = await context.charmstore.entityId(self.charm)
-        log.debug('Adding %s', entity_id)
-        await context.client_facade.AddCharm(channel=None, url=entity_id, force=False)
-        return entity_id
+        elif Schema.CHARM_STORE.matches(url.schema):
```
petevg, line 579: nitpick: these elifs could just be ifs. If .matches would match for multiple branches of the if, else if statements, then we have a bug, anyway.

juju/model.py
```diff
@@ -1416,33 +1419,78 @@ async def deploy(
...
-                     (not is_local and 'bundle/' in entity_id))
+            identifier = result['Id']
+            origin = client.CharmOrigin(source="charm-store",
+                                        architecture=architecture,
+                                        risk=channel)
+            is_bundle = url.series == "bundle"
+            if not series:
+                series = self._get_series(entity_url, result)
+
+        elif Schema.CHARM_HUB.matches(url.schema):
```
petevg, line 1466: All these special cases for local, charm store, and charm hub charms feel bad to me.

I don't think that it's necessarily in the scope of this PR to fix it. But I'd rather see this properly designed. Like there's a nice little class that knows how to do special processing if the charm is local or charm store, and the logic for the bundle just throws stuff at that class, and gets consistent objects back.

petevg (changes requested):

petevg (approved):
LGTM :-)

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #485, opened 2021-04-01
Fix Unit.resolved() not actually doing anything

### LINKED ISSUE #484: Unit.resolved() doesn't actually retry hooks
The value of the `tags` param [passed to `ResolveUnitErrors`](https://github.com/juju/python-libjuju/blob/d940c2c/juju/unit.py#L128) isn't of the correct format, so it doesn't actually trigger the hook to run again.  The correct format should be:

```python
return await app_facade.ResolveUnitErrors(
    all_=False,
    retry=retry,
    tags={'entities': [{'tag': self.tag}]})
```

johnsca:
The integration test for `Unit.resolved()` only confirmed that it doesn't error, but due to an issue with the format of the `tags` parameter, it wasn't actually affecting the desired unit. This fixes both the call and the test to confirm that it actually resolves the error.

Fixes #484


SimonRichardson:
!!build!!

Cynerva (approved):

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

SimonRichardson:
> python 3.5 failure, let's try again

$$merge$$

PR #492, opened 2021-05-14
Adds support for juju client proxies

tlm:
Adds support for Kubernetes client proxies introduced in Juju 2.9.

Relates to bug https://bugs.launchpad.net/juju/+bug/1926595

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -231,6 +231,7 @@ class Connection:
             retries=3,
             retry_backoff=10,
             specified_facades=None,
+            proxy=None,
     ):
         """Connect to the websocket.
 
@@ -308,6 +309,10 @@ class Connection:
             max_frame_size = self.MAX_FRAME_SIZE
         self.max_frame_size = max_frame_size
 
+        self.proxy = proxy
+        if self.proxy is not None:
+            self.proxy.connect()
+
         _endpoints = [(endpoint, cacert)] if isinstance(endpoint, str) else [(e, cacert) for e in endpoint]
         for _ep in _endpoints:
             try:
@@ -348,12 +353,23 @@ class Connection:
         else:
             url = "wss://{}/api".format(endpoint)
 
+        # We need to establish a server_hostname here for TLS sni if we are
+        # connecting through a proxy as the Juju controller certificates will
+        # not be covering the proxy
+        sock = None
+        server_hostname = None
+        if self.proxy is not None:
+            sock = self.proxy.socket()
+            server_hostname = "juju-app"
+
         return (await websockets.connect(
             url,
             ssl=self._get_ssl(cacert),
             loop=self.loop,
             max_size=self.max_frame_size,
-        ), url, endpoint, cacert)
+            server_hostname=server_hostname,
+            sock=sock,
+        )), url, endpoint, cacert
 
     async def close(self):
         if not self.ws:
@@ -363,6 +379,9 @@ class Connection:
         await self._receiver_task.stopped.wait()
         await self.ws.close()
         self.ws = None
+
+        if self.proxy is not None:
+            self.proxy.close()
 
     async def _recv(self, request_id):
         if not self.is_open:
@@ -551,11 +570,9 @@ class Connection:
         return await Connection.connect(**self.connect_params())
 
     def connect_params(self):
-        """Return a tuple of parameters suitable for passing to
+        """Return a dict of parameters suitable for passing to
         Connection.connect that can be used to make a new connection
-        to the same controller (and model if specified. The first
-        element in the returned tuple holds the endpoint argument;
-        the other holds a dict of the keyword args.
+        to the same controller (and model if specified).
         """
         return {
             'endpoint': self.endpoint,
@@ -566,6 +583,7 @@ class Connection:
             'bakery_client': self.bakery_client,
             'loop': self.loop,
             'max_frame_size': self.max_frame_size,
+            'proxy': self.proxy,
         }
 
     async def controller(self):
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -6,6 +6,7 @@ import macaroonbakery.httpbakery as httpbakery
 from juju.client.connection import Connection
 from juju.client.gocookies import GoCookieJar, go_to_py_cookie
 from juju.client.jujudata import FileJujuData
+from juju.client.proxy.factory import proxy_from_config
 from juju.errors import JujuConnectionError, JujuError
 
 log = logging.getLogger('connector')
@@ -55,6 +56,7 @@ class Connector:
 
         kwargs are passed through to Connection.connect()
         """
+
         kwargs.setdefault('loop', self.loop)
         kwargs.setdefault('max_frame_size', self.max_frame_size)
         kwargs.setdefault('bakery_client', self.bakery_client)
@@ -89,6 +91,8 @@ class Connector:
         endpoints = controller['api-endpoints']
         accounts = self.jujudata.accounts().get(controller_name, {})
 
+        proxy = proxy_from_config(controller.get('proxy-config', None))
+
         await self.connect(
             endpoint=endpoints,
             uuid=None,
@@ -97,6 +101,7 @@ class Connector:
             cacert=controller.get('ca-cert'),
             bakery_client=self.bakery_client_for_controller(controller_name),
             specified_facades=specified_facades,
+            proxy=proxy,
         )
         self.controller_name = controller_name
         self.controller_uuid = controller["uuid"]
@@ -121,8 +126,11 @@ class Connector:
         account = self.jujudata.accounts().get(controller_name, {})
         models = self.jujudata.models().get(controller_name, {}).get('models',
                                                                      {})
+
         if model_name not in models:
             raise JujuConnectionError('Model not found: {}'.format(model_name))
+
+        proxy = proxy_from_config(controller.get('proxy-config', None))
 
         # TODO if there's no record for the required model name, connect
         # to the controller to find out the model's uuid, then connect
@@ -137,6 +145,7 @@ class Connector:
             password=account.get('password'),
             cacert=controller.get('ca-cert'),
             bakery_client=self.bakery_client_for_controller(controller_name),
+            proxy=proxy,
         )
         self.controller_name = controller_name
         self.model_name = controller_name + ':' + model_name
--- a/juju/client/proxy/factory.py
+++ b/juju/client/proxy/factory.py
@@ -1,0 +1,26 @@
+from juju.client.proxy.kubernetes.proxy import KubernetesProxy
+
+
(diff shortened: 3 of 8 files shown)


juju/client/connection.py
```diff
@@ -348,12 +353,24 @@ async def _open(self, endpoint, cacert):
         else:
             url = "wss://{}/api".format(endpoint)
 
+        # We need to establish a server_hostname here for TLS sni if we are
+        # connecting through a proxy as the Juju controller certificates will
+        # not be covering the proxy
+        try:
+            sock = self.proxy.socket()
```
hpidcock, line 360: I'm confused to why the exception is being thrown here. Is it because .socket() may not be on the proxy object?

juju/client/connector.py
```diff
@@ -89,6 +91,9 @@ async def connect_controller(self, controller_name=None, specified_facades=None)
         endpoints = controller['api-endpoints']
         accounts = self.jujudata.accounts().get(controller_name, {})
 
+        proxyFactory = JujuDataFactory()
```
hpidcock, line 94: proxy_factory

juju/client/connector.py
```diff
@@ -121,9 +127,13 @@ async def connect_model(self, model_name=None):
         account = self.jujudata.accounts().get(controller_name, {})
         models = self.jujudata.models().get(controller_name, {}).get('models',
                                                                      {})
+
         if model_name not in models:
             raise JujuConnectionError('Model not found: {}'.format(model_name))
 
+        proxyFactory = JujuDataFactory()
```
hpidcock, line 134: proxy_factory

hpidcock (approved):

juju/client/connection.py
benhoyt: Yeah, agreed it seems very weird to catch `AttributeError` here, when you know the "shape" of the value. Is this to catch the case when `self.proxy` is None? In that case I'd use a plain `if self.proxy is not None` type of check instead of try/except.

juju/client/connection.py
```diff
@@ -566,6 +586,7 @@ def connect_params(self):
             'bakery_client': self.bakery_client,
             'loop': self.loop,
             'max_frame_size': self.max_frame_size,
+            'proxy': self.proxy,
```
benhoyt, line 589: Not your mistake, but it looks like the docstring for this `connect_params` function is incorrect. It says it "returns a tuple ... the first element in the tuple" but it actually returns a dict. Maybe you could fix while you're in there?

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
+from abc import ABC, abstractmethod
+from juju.client.proxy.kubernetes.proxy import KubernetesProxy
+
+class ProxyTypeNotDefinedError(Exception):
```
benhoyt, line 4: This class doesn't look like it's used - should it be removed?

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
...
+    pass
+
+class UnknownProxyTypeError(Exception):
```
benhoyt, line 7: It may make sense just to use the built-in `ValueError` exception type for this. That's typically what's raised for invalid params to a function, and callers probably don't need the extra specificity. If you really want/need to keep it, make it inherit from `ValueError`.

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
...
+    pass
+
+class Factory(ABC):
+
+    @abstractmethod
+    def proxy_from_config(self, conf):
+        raise NotImplementedError()
+
+class JujuDataFactory(Factory):
```
benhoyt, line 16: I'm not sure why this is called `JujuDataFactory` -- that implies it's a "factory for Juju data", but it's a factory for a Proxy, right? Also, the use of Factory here gives me the Java-shivers. :-) It looks like this could just be replaced with a function (which are first-class in Python), for example `proxy_from_config()` as an ordinary function. One of the give-aways is that it's a class with only one method, and it doesn't use `self`. Here's a well-known talk by a Python programmer about this (though arguably he goes a bit too far): https://www.youtube.com/watch?v=o9pEzgHorH0

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def proxy_from_config(self, conf):
+        if conf is None:
+            return None
+
+        if "type" not in conf:
```
benhoyt, line 22: Would be good to use a consistent quote style, single or double. I lean towards single as that's what's Python's `repr()` produces, and from a quick scan it looks like that's what this project prefers too. Also elsewhere in this PR.

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
...
+            return None
+
+        proxy_type = conf['type']
+        if proxy_type != "kubernetes-port-forward":
+            raise UnknownProxyTypeError()
```
benhoyt, line 27: You typically want to pass a message here, including the value that was incorrect. For example:

    raise ValueError('invalid proxy type {!r}'.format(proxy_type))

The `!r` format syntax prints the repr of the value, i.e., quoted string.

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return self.__construct_kube_proxy(conf['config'])
+
+    def __construct_kube_proxy(self, config):
```
benhoyt, line 31: FYI, I know Python supports this, but it's not common to use name-mangled, double-underscore prefix methods for privacy. You typically just use single-under-prefixed ones. This project uses single-underscore prefixed names for private methods (`rg 'def __[a-z]+\('` returns no results).

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -0,0 +1,76 @@
...
+            pods.items[0].metadata.name,
+            self.namespace,
+            ports=str(self.remote_port),
+        )
+
+
+    def __del__(self):
+        self.disconnect()
+
+    def disconnect(self):
```
benhoyt, line 66: In my experience, `close` would be a more common name for this kind of thing.

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -0,0 +1,76 @@
...
+        config.api_key = {"authorization": "Bearer " + service_account_token}
+
+        self.namespace = namespace
+        self.remote_port = remote_port
+        self.service = service
+
+        try:
+            self.remote_port = int(remote_port)
+        except ValueError:
+            raise ValueError("Invalid port number: %s" % remote_port)
```
benhoyt, line 30: You would usually just use `remote_port` directly, and the caller would have to pass in and int or int-like object. Is there a reason this should do the string parsing in this class?

benhoyt, line 30: Typically you'd use `str.format` in modern Python, as in `"Invalid port number: {}".format(remote_port)`

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if ca_cert:
+            self.temp_ca_file = tempfile.NamedTemporaryFile()
+            self.temp_ca_file.write(bytes(ca_cert, 'utf-8'))
+            self.temp_ca_file.flush()
+            config.ssl_ca_cert = self.temp_ca_file.name
```
benhoyt, line 36: I guess this has to be an actual file? If so, we should delete the temp file in close/disconnect. I believe you can do this by calling `close()` on `self.temp_ca_file`.

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def connect(self):
+        corev1 = client.CoreV1Api(self.api_client)
+        service = corev1.read_namespaced_service(self.service, self.namespace)
+
+        label_selector = ""
+        for k, v in service.spec.selector.items():
+            if label_selector:
+                label_selector += ","
+            label_selector += k + "=" + v
```
benhoyt, line 48: This is a great use for Python's `str.join` method and generator expressions / comprehensions:

```python
label_selector = ','.join(k + '=' + v for k, v in service.spec.selector.items())
```

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -0,0 +1,76 @@
...
+    def disconnect(self):
+        try:
+            self.port_forwarder.close()
+        except AttributeError:
+            pass
+
+    def socket(self):
+        try:
+            self.port_forwarder.socket(self.remote_port)._socket
+        except AttributeError:
```
benhoyt, line 75: Similar here -- it's very odd to see `AttributeError` being caught in a place like this. What are we actually catching? Could we do a `is None` check instead?

juju/client/proxy/proxy.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC, abstractmethod
```
benhoyt, line 1: You generally don't use ABC's in Python (duck typing FTW). You'd just define and document the class, and people can subclass it or define a class that quacks like it if they want. I think you should just drop this abstract class (the rest of the project doesn't use them).

juju/client/proxy/factory.py
tlm: Was trying to go for duck typing buck fell short.

juju/client/proxy/kubernetes/proxy.py
tlm: Yeah ports aren't integers. No maths is performed on them. It's theoretically possible that we may want to open this up in the future to support named ports on the Kubernetes side.

juju/client/connection.py
```diff
@@ -551,11 +570,9 @@ async def clone(self):
         return await Connection.connect(**self.connect_params())
 
     def connect_params(self):
-        """Return a tuple of parameters suitable for passing to
+        """Return a dict of parameters suitable for passing to
         Connection.connect that can be used to make a new connection
-        to the same controller (and model if specified. The first
-        element in the returned tuple holds the endpoint argument;
-        the other holds a dict of the keyword args.
+        to the same controller (and model if specified.
```
benhoyt, line 575: Nit: no right parenthesis.

juju/client/proxy/factory.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC, abstractmethod
```
benhoyt, line 1: This import is no longer needed here, correct?

benhoyt (approved):

tlm:
$$merge$$

tlm:
$$merge$$

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #493, opened 2021-05-18
Remove machine work around

SimonRichardson:
The following change removes the workaround for the 2.2.3 release and
subsequent 2.3-beta1 release. As these are older releases that users
should have moved away from, it's time to remove the workarounds.

The code in question was highly contentious anyway, as it was never
awaited on the machine, so might not have been doing what we wanted it
to do in the long run.

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -5,7 +5,7 @@ import os
 
 import pyrfc3339
 
-from . import model, tag, utils
+from . import model, tag
 from .annotationhelper import _get_annotations, _set_annotations
 from .client import client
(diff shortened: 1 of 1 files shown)


petevg (approved):
+1

Good to remove old cruft :-)

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

> RuntimeError: There is no current event loop in thread 'Dummy-1'.

We'll keep fighting this until we work out how to fix this.

PR #497, opened 2021-05-21
Honor charm channel in bundles

### LINKED ISSUE #496: Bundle deploy doesn't honor charm channels
The bundle handler's `AddCharmChange` implementation is [ignoring the channel](https://github.com/juju/python-libjuju/blob/d6fae9e/juju/bundle.py#L439) even though it [does have the channel available](https://github.com/juju/python-libjuju/blob/d6fae9e/juju/bundle.py#L447-L448).  E.g., a deploy logged this:

```
INFO     juju.bundle:bundle.py:166 Applying change: upload charm cs:~containers/kubernetes-worker for series focal from channel edge
```

But in fact deployed the charm rev from stable.

johnsca:
Fixes #496

### DIFF
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -434,7 +434,7 @@ class AddCharmChange(ChangeInfo):
         if self.charm.startswith('local:'):
             return self.charm
 
-        entity_id = await context.charmstore.entityId(self.charm)
+        entity_id = await context.charmstore.entityId(self.charm, channel=self.channel)
         log.debug('Adding %s', entity_id)
         await context.client_facade.AddCharm(channel=self.channel, url=entity_id, force=False)
         return entity_id
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -436,7 +436,7 @@ class AddCharmChange(ChangeInfo):
 
         entity_id = await context.charmstore.entityId(self.charm)
         log.debug('Adding %s', entity_id)
-        await context.client_facade.AddCharm(channel=None, url=entity_id, force=False)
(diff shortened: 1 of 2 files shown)


juju/bundle.py
```diff
@@ -436,7 +436,7 @@ async def run(self, context):
 
         entity_id = await context.charmstore.entityId(self.charm)
```
Cynerva, line 437: I suspect the channel might need to be passed into charmstore.entityId too: https://github.com/juju/theblues/blob/a8cf043507cde71eba77727f2e6952d323349760/theblues/charmstore.py#L329

SimonRichardson:
Also be aware this isn't always the case, it's a 2.7 feature I believe.

johnsca:
@SimonRichardson I'm not sure what you mean.  Even the `ClientFacade.AddCharm` in `_client1.py` [supports the channel arg](https://github.com/juju/python-libjuju/blob/eebf2b3/juju/client/_client1.py#L4177).  If you mean that it's not supported in the bundle, then I think that it would be ignored or rejected by the `BundleFacade` (if that is even available there).

SimonRichardson:
> If you mean that it's not supported in the bundle, then I think that it would be ignored or rejected by the `BundleFacade` (if that is even available there).

This ☝️

SimonRichardson (approved):

SimonRichardson:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #504, opened 2021-06-02
Fix series being required for local charms

johnsca:
Local charms may no longer need a series if they are metadata v2 container charms. The Juju CLI uses the model default instead.

Fixes [lp:1929076][]

[lp:1929076]: https://bugs.launchpad.net/juju/+bug/1929076

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1483,7 +1483,9 @@ class Model:
                 series = series or get_charm_series(charm_dir)
                 if not series:
                     config = await self.get_config()
-                    series = config.get("default-series")
+                    default_series = config.get("default-series")
+                    if default_series:
+                        series = default_series.value
                 if not series:
                     raise JujuError(
                         "Couldn't determine series for charm at {}. "
--- a/juju/model.py
+++ b/juju/model.py
@@ -1482,8 +1482,8 @@ class Model:
                     os.path.expanduser(entity_id))
                 series = series or get_charm_series(charm_dir)
                 if not series:
-                    config = await self.get_config()
-                    default_series = config.get("default-series")
+                    model_config = await self.get_config()
+                    default_series = model_config.get("default-series")
                     if default_series:
                         series = default_series.value
                 if not series:
--- a/juju/model.py
+++ b/juju/model.py
@@ -1482,6 +1482,9 @@ class Model:
                     os.path.expanduser(entity_id))
                 series = series or get_charm_series(charm_dir)
                 if not series:
+                    config = await self.get_config()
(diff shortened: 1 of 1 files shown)


johnsca:
Testing with this got me the following, which I don't understand:

```
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/tests/integration/test_charm.py", line 16, in test_build_and_deploy
    await ops_test.model.deploy(
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1505, in deploy
    return await self._deploy(
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1659, in _deploy
    result = await app_facade.Deploy(applications=[app])
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 480, in wrapper
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/client/_client13.py", line 1102, in Deploy
    reply = await self.rpc(msg)
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 623, in rpc
  File "/home/johnsca/juju/envs/dev/charms/kube-state-metrics/.tox/integration/lib/python3.8/site-packages/juju/client/connection.py", line 512, in rpc
    raise errors.JujuError(err_results)
juju.errors.JujuError: ['unknown option "max-status-history-size"']
```

The actual API call there is `await app_facade.Deploy(applications=[app])` which doesn't reference that option at all, though it is set on the controller:

```
max-status-history-size            default     5G
```

jameinel:
That feels like we are mixing the config of the application with the config
of the model. As in, we're reading the model-config and accidentally
shoving it into the application config.

jameinel:
My guess is that this code:
```
config = await self.get_config()
series = config.get("default-series")
default_series = config.get("default-series")
if default_series:
series = default_series.value
```
Is overwriting a 'config' variable that is actually application config when
you read the model config.
So you want
```
model_config = await self.get_config()
...
```

johnsca:
@jameinel You're absolutely right.  Thanks for the catch.  It seems to be all working now.

SimonRichardson (approved):

SimonRichardson:
$$merge$$

PR #507, opened 2021-06-09
Add Local Charm Upgrade support

DomFleischmann:
This commit will add support for upgrading from a locally built charm with local resources.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -15,12 +15,17 @@
 import asyncio
 import json
 import logging
+import zipfile
+import os
+import yaml
+from pathlib import Path
 
 from . import model, tag
 from .status import derive_status
 from .annotationhelper import _get_annotations, _set_annotations
 from .client import client
 from .errors import JujuError
+from .bundle import get_charm_series
 from .placement import parse as parse_placement
 
 log = logging.getLogger(__name__)
@@ -578,9 +583,10 @@ class Application(model.ModelEntity):
         :param str switch: Crossgrade charm url
 
         """
-        # TODO: Support local upgrades
         if path is not None:
-            raise NotImplementedError("path option is not implemented")
+            await self.local_refresh(channel, force, force_series, force_units,
+                                     path, resources)
+            return
         if resources is not None:
             raise NotImplementedError("resources option is not implemented")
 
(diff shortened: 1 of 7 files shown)


johnsca:
LGTM, but would be good to have test coverage in [`tests/integration/test_application.py`](https://github.com/juju/python-libjuju/blob/master/tests/integration/test_application.py).

SimonRichardson:
!!build!!

tests/integration/test_application.py
```diff
@@ -143,6 +144,20 @@ async def test_upgrade_charm_switch(event_loop):
...
+async def test_upgrade_local_charm(event_loop):
+    async with base.CleanModel() as model:
+        tests_dir = Path(__file__).absolute().parent
+        charm_path = tests_dir / 'upgrade-charm'
+        app = await model.deploy('ubuntu-0')
+        await model.block_until(lambda: (len(app.units) > 0 and
+                                         app.units[0].machine))
+        assert app.data['charm-url'] == 'cs:ubuntu-0'
+        await app.upgrade_charm(path=charm_path)
+        assert app.data['charm-url'] == 'local:ubuntu'
```
johnsca, line 158: When I ran this test, I got:

```
juju.errors.JujuAPIError: cannot upgrade application "ubuntu" to charm "local:focal/ubuntu-0": cannot change an application's series
```

Also, this doesn't actually wait for the charm to deploy or upgrade to confirm that it completes successfully.  My suggestion:

```suggestion
        app = await model.deploy('ubuntu', series='focal')
        await model.wait_for_idle(wait_for_active=True)
        assert app.data['charm-url'].startswith('cs:ubuntu')
        await app.upgrade_charm(path=charm_path)
        await model.wait_for_idle()  # nb: can't use wait_for_active because test charm goes to "waiting"
        assert app.data['charm-url'] == 'local:focal/ubuntu-0'
```

johnsca (changes requested):

SimonRichardson:
!!build!!

juju/application.py
```diff
@@ -684,6 +690,68 @@ async def refresh(
...
+
+        charm_url = await self.model.add_local_charm_dir(charm_dir, series)
+        if resources is not None:
+            if str(path).endswith('.charm'):
+                with zipfile.ZipFile(path, 'r') as charm_file:
+                    metadata = yaml.load(charm_file.read('metadata.yaml'), Loader=yaml.FullLoader)
+            else:
+                entity_path = Path(path)
+                metadata_path = entity_path / 'metadata.yaml'
+                metadata = yaml.load(metadata_path.read_text(), Loader=yaml.FullLoader)
```
SimonRichardson, line 731: This feels like it could be extracted somewhere so we could reuse this.

DomFleischmann: I agree, would this make sense somewhere like `utils.py`?

SimonRichardson: I think we should be specific about this and put it in `juju/charm.py`

DomFleischmann: Updated, I didn't add it as a class method as i don't think it makes sense to initialize a charm object in my use case,

johnsca (approved):
LGTM, and test passes locally.

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

PR #517, opened 2021-08-06
add wait_for_status instead of wait_for_active

sed-i:


johnsca (commented):
Looks reasonable to me. However, there are some integration tests which use `wait_for_idle(wait_for_active=True)` and if that is going to be deprecated, those should be updated (and maybe a regression test included).

sed-i:
Updated integration tests and added unit tests.

sed-i:
@johnsca how about renaming `wait_for_status` to simply `status`, such that `await ops_test.model.wait_for_idle(status="active")`

johnsca:
+1 to `status`.  That seems plenty clear and much less typing.

sed-i:
@johnsca: renamed to status and patched timedelta

johnsca (approved):

tox.ini
```diff
@@ -10,6 +10,9 @@ skipsdist=True
 [pytest]
 markers =
     serial: mark a test that must run by itself
+filterwarnings =
+    ignore::DeprecationWarning:asynctest
+    ignore::DeprecationWarning:websockets
```
SimonRichardson, line 15: :clap:

SimonRichardson:
!!build!!

SimonRichardson (approved):

SimonRichardson:
$$merge$$

SimonRichardson:
$$merge$$

PR #531, opened 2021-09-03
Upgrade websockets version.

juanmanuel-tirado:
Upgrade websockets to version 9.1 to skip security warning from dependabot regarding [CVE 2021-33880](https://github.com/advisories/GHSA-8ch4-58qp-g3mp)


setup.py
```diff
@@ -35,8 +35,8 @@
         'pyRFC3339>=1.0,<2.0',
         'pyyaml>=5.1.2,<=6.0',
         'theblues>=0.5.1,<1.0',
-        'websockets>=7.0,<8.0 ; python_version<"3.9"',
-        'websockets>=8.0,<9.0 ; python_version>="3.9"',
+        'websockets>="9.1" ; python_version<"3.9"',
```
SimonRichardson, line 38: We can't do this, as we use a patched version of websockets, see: https://github.com/juju/python-libjuju/blob/2.9/tox.ini#L35

SimonRichardson (changes requested):

setup.py
juanmanuel-tirado: !!build!!

PR #539, opened 2021-09-09
Fix issue 530: Check the controller for unsynched models

cderici:
### Description

The issue occurs when a `model.connect` runs before the model is synced with the local cache (`~/.local/share/juju/models.yaml`). This patch fixes it by consulting to the controller for the model `uuid` in case the model is not synced with the local cache yet.

#### Contains 3 commits

1. implements the patch
2. adds a test that targets this issue/patch
3. fixes a typo

### QA Steps

A new test is added just for this so the following should pass:

```
$ tox -e integration -- tests/integration/test_model.py::test_model_cache_update
```

### Discussion:

- After succesfully getting the `uuid` of the model (or after connecting to it) it might be nice to trigger an update on the local cache. 
- If we have a way of triggering an update on the local cache (before connecting to anything), then an alternative solution to all of this might be to just update the local cache before querying it for the model `uuid` (instead of going through the controller).

### DIFF
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -141,9 +141,6 @@ class Connector:
             for user_model in response.user_models:
                 if 'admin/' + user_model.model.name == model_name:
                     model_uuid = user_model.model.uuid
-            # TODO refresh the local cache
-            # then self.jujudata.refresh()
-            # then self.jujudata.models()
 
         if model_uuid is None:
             raise JujuConnectionError('Model not found: {}'.format(model_name))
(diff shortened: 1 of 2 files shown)


cderici:
@SimonRichardson I couldn't find a way to use the `AllWatcher` to trigger an update on the local cache, if you can point me in the right direction then I can add it to this PR too.

juju/client/connector.py
```diff
@@ -125,26 +126,42 @@ async def connect_model(self, model_name=None, **kwargs):
...
+                self.connection())
+            # get all the user models from the api
+            response = await controller_facade.AllModels()
+            # search the one that contains admin/model_name
+            for user_model in response.user_models:
+                if 'admin/' + user_model.model.name == model_name:
+                    model_uuid = user_model.model.uuid
+            # TODO refresh the local cache
+            # then self.jujudata.refresh()
+            # then self.jujudata.models()
```
SimonRichardson, line 146: Delete me.

SimonRichardson (approved):

cderici:
!!build!!

SimonRichardson:
!!build!!

cderici:
$$merge$$

PR #540, opened 2021-09-09
Allow ApplicationFacade set_config with non-string values

### LINKED ISSUE #388: ApplicationFacade doesn't allow set_config with non-string values
In [this](https://github.com/openstack-charmers/zaza-openstack-tests/pull/156) PR, I discovered that either python-libjuju or libjuju itself has an issue with application set_config when the values of the dictionary are not strings. 

```
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/zaza/model.py", line 531, in async_set_application_config
2020-01-24 12:40:44 [INFO]     .set_config(configuration))
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/application.py", line 383, in set_config
2020-01-24 12:40:44 [INFO]     return await app_facade.Set(application=self.name, options=config)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/facade.py", line 471, in wrapper
2020-01-24 12:40:44 [INFO]     reply = await f(*args, **kwargs)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/_client8.py", line 1124, in Set
2020-01-24 12:40:44 [INFO]     reply = await self.rpc(msg)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/facade.py", line 607, in rpc
2020-01-24 12:40:44 [INFO]     result = await self.connection.rpc(msg, encoder=TypeEncoder)
2020-01-24 12:40:44 [INFO]   File "/home/ubuntu/charms/layers/keystone-ldap/build/builds/keystone-ldap/.tox/func-smoke/lib/python3.7/site-packages/juju/client/connection.py", line 456, in rpc
2020-01-24 12:40:44 [INFO]     raise errors.JujuAPIError(result)
2020-01-24 12:40:44 [INFO] juju.errors.JujuAPIError: json: cannot unmarshal bool into Go struct field ApplicationSet.options of type string
```

interestingly enough, i don't think you could execute which maybe isn't a bug, but its certainly confusing for a developer.  

```python
    config = await application.get_config()
    await application.set_config(config)
```

here's what seems to be required
```python
    config = await application.get_config()
    str_config = {k: str(v.get('value')) for config.items()}
    await application.set_config(str_config)
```

cderici:
### Description

So we want to be able to do the following:

```python
    config = await application.get_config()
    await application.set_config(config)
```

However, the issue was, that on the Juju side while the `get_config` returns a `map[string]value`, the `set_config` expects a `map[string]string`. This patch fixes it by casting everything into `str` on `pylibjuju` side before it's passed into the api.

Fixes #388

### QA Steps

There's a test included in the patch that targets this, so the following should pass:
```
 $ tox -e integration -- tests/integration/test_application.py::test_get_set_config
```

### Notes & Discussion

So clearly this is not the ideal solution, which would be to have Juju accept configs that look like `map[string]value` and cast things individually as it sees fit. However, @SimonRichardson suggests that that has the potential of breaking a lot of things, so looks like we'll leave this as a future work. 

Note that this also means that any library (other than `pylibjuju`) that wants to use `Set` has to do the same thing on their own, or abandon their project and start using `pylibjuju` instead, which would be the cooler option.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -517,7 +517,12 @@ class Application(model.ModelEntity):
         log.debug(
             'Setting config for %s: %s', self.name, config)
 
-        return await app_facade.Set(application=self.name, options=config)
+        str_config = {}
+        for k, v in config.items():
+            if v.get('value') is not None:
+                str_config[k] = str(v.get('value'))
+
+        await app_facade.Set(application=self.name, options=str_config)
 
     async def reset_config(self, to_default):
         """
--- a/tests/integration/test_application.py
+++ b/tests/integration/test_application.py
@@ -48,6 +48,31 @@ async def test_action(event_loop):
 
 @base.bootstrapped
 @pytest.mark.asyncio
+async def test_get_set_config(event_loop):
+    async with base.CleanModel() as model:
+        ubuntu_app = await model.deploy(
+            'percona-cluster',
+            application_name='mysql',
+            series='xenial',
+            channel='stable',
+            config={
+                'tuning-level': 'safest',
+            },
+            constraints={
+                'arch': 'amd64',
+                'mem': 256 * MB,
+            },
+        )
+
+        config = await ubuntu_app.get_config()
+        await ubuntu_app.set_config(config)
+
+        config2 = await ubuntu_app.get_config()
+        assert config == config2
+
+
+@base.bootstrapped
+@pytest.mark.asyncio
 async def test_status_is_not_unset(event_loop):
     async with base.CleanModel() as model:
         app = await model.deploy(
--- a/tests/unit/test_bundle.py
+++ b/tests/unit/test_bundle.py
@@ -1001,29 +1001,45 @@ class TestBundleHandler:
 
         bundle = await handler._handle_local_charms(bundle, bundle_dir)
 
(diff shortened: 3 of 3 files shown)


SimonRichardson (approved):

SimonRichardson:
$$merge$$

SimonRichardson:
Can you rebase against master and we'll try again.

cderici:
!!build!!

cderici:
$$merge$$

PR #541, opened 2021-09-09
a random small bug fix

cderici:
### Description

Directly modifying the inputs to the functions should be avoided, as Python is call-by-reference. 

Basically, if we modify the argument directly, and if the caller of the function still wants to do something with it after the call to callee is completed, then it would see a different (mutated) object.


SimonRichardson (approved):

cderici:
$$merge$$

PR #542, opened 2021-09-10
Skip macaroon tests issue 534

### LINKED ISSUE #534: A blocking test case is causing the whole test build to time out
There's at most two test cases that seem to be blocking the test builds to complete, so the build times out after 120 minues. Here's an example console output from [such a build](https://jenkins.juju.canonical.com/job/github-integration-tests-pylibjuju-python3.7/86/console) :

image

Note that, because of this time out, `tox` can't get any chance to print any stderr output from any of the failing test cases. (which was the first symptom that lead us to this issue in the first place)

This seems to be related to test cases `test_macaroon_auth_with_bad_key` and `test_macaroon_auth_with_unauthorized_user`, as they are the only ones that were initiated but never finished with a line such as `PASSED` or `FAILED`. They both use the environment variable `TEST_AGENTS`, **disabling which allows the build to complete without a timeout**, because pytest skips them: 

image

No other test cases seem to depend on this environment variable. 

Running both cases on local with `TEST_AGENTS` defined as in the CI build (just grep for it on the console output) results in `juju.errors.JujuAPIError: no credentials provided` and completes just fine, since both tests are marked as `xfail`.

It is unclear why these cases block on CI.

cderici:
### Description

Marks the `test_macaroon_auth` tests (there are 3 of them) with `skip`.

Fixes #534 

### Notes & Discussion

These tests are several years old, and they were always failing, in fact, they were also marked with `xfail` by their author, currently they seem to have started to block the entire test run on the CI, so they need to be revised. 
Until then, skipping them seems to be reasonable.

### DIFF
--- a/tests/integration/test_macaroon_auth.py
+++ b/tests/integration/test_macaroon_auth.py
@@ -16,7 +16,7 @@ log = logging.getLogger(__name__)
 
 @base.bootstrapped
 @pytest.mark.asyncio
-#@pytest.mark.xfail
+# @pytest.mark.xfail
 @pytest.mark.skip('one of old macaroon_auth tests, needs to be revised')
 async def test_macaroon_auth(event_loop):
     auth_info, username = agent_auth_info()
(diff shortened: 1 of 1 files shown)


cderici:
[skip the macaroon_auth tests](https://trello.com/c/SRIfwOnT/3951-skip-the-macaroonauth-tests)

SimonRichardson (approved):

cderici:
!!build!!

SimonRichardson:
!!build!!

SimonRichardson:
$$merge$$

cderici:
$$merge$$

PR #560, opened 2021-10-08
remove the event loop arguments

cderici:
### Description

Don't need to thread the loop object all around since
asyncio can create tasks without the loop argument.

This solves the intermittent issue with calling asyncio.get_event_loop
in a worker thread (which errors), as well as gets rid of deprecation
warnings for the loop argument. Furthermore, the `get_event_loop`
is [getting deprecated](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) in `python3.10` anyways.

Kept all the modifications in a single commit to preserve the code
integrity (it's non-trivial to have the tests pass without all the
modifications in place).

### QA Steps

The tests should pass.

```
make test
```

PR #562, opened 2021-10-13
add debug-log

### LINKED ISSUE #549: Enable debug_log
It would be nice to have debug_log in python-libjuju.

From @SimonRichardson: This is what needs to be implemented: https://github.com/juju/juju/blob/develop/api/common/logs.go#L107-L146

cderici:
### Description

This PR adds `debug-log` functionality into pylibjuju, by adding a separate Connection object into the Connector of a Model. 

User runs `await model.debug_log()` with a writable target (defaults to `stdout`), we create a new connection with `/log` and start the receiving tasks. As the user continues to interact with the model, the debug log receiver tasks write the logs onto whatever target is provided. 

Fixes #549 

### QA Steps

```
python examples/debug-log.py
```

### Notes & Discussion

There are two things need to be done before this is ready to merge:

1. All the tasks need to be flushed before ending the program (that's why we're getting `ERROR:asyncio:Future exception was never retrieved` errors at the end when we run the example above). This is also closely related to #560 , which I'd prefer to complete and land before I fix this PR.

2. The filtering parameters of the debug-log (e.g. `no-tail`, `exclude-module`, `level`, `lines`, etc.) need to be passed through.

### DIFF
--- a/examples/debug-log.py
+++ b/examples/debug-log.py
@@ -1,0 +1,38 @@
+"""
+This example demonstrate how debug-log works
+
+"""
+import asyncio
+from juju import loop
+import logging
+import sys
+from logging import getLogger
+from juju.model import Model
+
+LOG = getLogger(__name__)
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+
+async def main():
+    model = Model()
+    await model.connect_current()
+
+    await model.debug_log()
+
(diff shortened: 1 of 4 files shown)


cderici:
[debug_log in pylibjuju](https://trello.com/c/8tvvR8Dp/4031-debuglog-in-pylibjuju)

PR #573, opened 2021-11-04
Consistently getting a unit's public address

### LINKED ISSUE #551: Getting a unit's public_address is unreliable in Juju 2.9
Calling [unit.public_address](https://github.com/juju/python-libjuju/blob/master/juju/unit.py#L66-L71) works reliably on juju 2.8 but fails intermittently, but frequently, on juju 2.9.

This method calls into the safe_data attribute on the object which, when successful, looks like:

```python
{'agent-status': {'current': 'idle',
                  'message': '',
                  'since': '2021-09-30T10:05:27.491219135Z',
                  'version': '2.8.10'},
 'application': 'keystone',
 'charm-url': 'cs:~openstack-charmers-next/keystone-563',
 'life': 'alive',
 'machine-id': '1',
 'model-uuid': '5beccf88-9f05-402c-8c97-584b490dceab',
 'name': 'keystone/0',
 'port-ranges': [{'from-port': 5000, 'protocol': 'tcp', 'to-port': 5000}],
 'ports': [{'number': 5000, 'protocol': 'tcp'}],
 'principal': '',
 'private-address': '10.5.0.141',
 'public-address': '10.5.0.141',
 'series': 'bionic',
 'subordinate': False,
 'workload-status': {'current': 'active',
                     'message': 'Unit is ready',
                     'since': '2021-09-30T09:48:11.443378654Z',
                     'version': ''}}
```

As the method is defined as a `@property`, the moment it returns None, it will never re-evaluate, meaning that the unit will _never_ have a public address.

cderici:
### Description

Assigning a unit's public address may take a little time, which makes the result of `unit.public_address` depend on when it's called, thereby creating inconsistency. This PR adds a method, namely `get_public_address`, that waits (with a timeout) until the unit is assigned an address and returns the address when it's assigned.

Fixes #551

Jira card [#141](https://warthogs.atlassian.net/browse/JUJU-141)

### QA Steps

```sh
tox -e integration -- tests/integration/test_unit.py::test_unit_public_address
```

### Notes & Discussion

**A minor design decision:** is that in the case of a timeout, `get_public_address` returns a `None`, rather than a `TimeoutError`.

### DIFF
--- a/tests/integration/test_unit.py
+++ b/tests/integration/test_unit.py
@@ -19,7 +19,7 @@ async def test_unit_public_address(event_loop):
         )
 
         for unit in app.units:
-            addr = await unit.get_public_address()
+            addr = await unit.get_public_address(timeout=60 * 4)
             assert addr, 'unit public address not set'
 
 
--- a/tests/integration/test_unit.py
+++ b/tests/integration/test_unit.py
@@ -16,7 +16,15 @@ async def test_unit_public_address(event_loop):
             application_name='ubuntu',
             series='trusty',
             channel='stable',
+            num_units=2,
         )
+
+        await model.block_until(
+            lambda: app.units,
+            timeout=480,
+        )
+
+        assert len(app.units) >= 1
 
         for unit in app.units:
             addr = await unit.get_public_address(timeout=60 * 4)
--- a/tests/integration/test_unit.py
+++ b/tests/integration/test_unit.py
@@ -19,15 +19,14 @@ async def test_unit_public_address(event_loop):
             num_units=2,
         )
 
-        await model.block_until(
-            lambda: app.units,
-            timeout=480,
-        )
+        # wait for the units to come up
+        await model.block_until(lambda: app.units, timeout=480)
 
+        # make sure we have some units to test with
         assert len(app.units) >= 1
 
         for unit in app.units:
-            addr = await unit.get_public_address(timeout=60 * 4)
+            addr = await unit.get_public_address(timeout=480)
             assert addr, 'unit public address not set'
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_unit.py
```diff
@@ -7,6 +7,22 @@
...
+async def test_unit_public_address(event_loop):
+    async with base.CleanModel() as model:
+        app = await model.deploy(
+            'cs:ubuntu-0',
+            application_name='ubuntu',
+            series='trusty',
+            channel='stable',
+        )
+
+        for unit in app.units:
```
SimonRichardson, line 21: Can we assert we got at least one unit, otherwise we skip the assertions

cderici: Not sure what you mean, are you saying that getting the public_address of one the units is enough?

SimonRichardson: You need to ensure that units is >= 1, otherwise the test will pass and you won't know it's failed as the for loop looks for units. What happens if units is zero for example?

cderici: Yeah I'm dumb, you're right I should probably use `num_units=2` or something with the `deploy` and then wait until the units are up, that way it's ensured we'll have some units. 
An assertion about units afterwards certainly wouldn't hurt, though might be redundant, because if we have no units after using `num_units`, then we have much bigger problems, so some other test (prob. about deploy) should already be failing. I'll make the changes :+1:

cderici:
[Getting a unit's public-address in libjuju](https://trello.com/c/ijdPEFa7/4136-getting-a-units-public-address-in-libjuju)

cderici:
!!build!!

SimonRichardson (approved):

cderici:
$$merge$$

SimonRichardson:
We should potentially revert this. The fix is not in pylibjuju but in juju AllWatcher implementation.

PR #575, opened 2021-11-04
[JUJU-140] Model.wait_for_idle -- for apps with no units yet

### LINKED ISSUE #574: wait_for_idle ignores idle_period if app has no units
If an app has no units, then `wait_for_idle` would not wait `idle_period` before returning.
This may also be a problem when the host is slow enough (e.g. github actions) such that python-libjuju sees an app status of "active-idle" before a unit is started.

https://github.com/juju/python-libjuju/blob/c68ee681995dc4b966ed14160d61a72f249c3b7b/juju/model.py#L2624

cderici:
#### Description

`Model.wait_for_idle`'s behavior is not specified for applications that have no units when `wait_for_idle` is called. In that case, it just returns without waiting for anything at all. This PR changes that behavior into waiting for units to show up (with timeout that already exists).

Fixes #574 

#### QA Steps

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_without_units
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_not_enough_units
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_enough_units
```

#### Notes & Discussion

### DIFF
--- a/juju/jasyncio.py
+++ b/juju/jasyncio.py
@@ -30,7 +30,7 @@ ROOT_LOGGER = logging.getLogger()
 
 from asyncio import Event, TimeoutError, Queue, ensure_future, \
     gather, sleep, wait_for, create_subprocess_exec, subprocess, \
-    wait, FIRST_COMPLETED, Lock, as_completed, \
+    wait, FIRST_COMPLETED, Lock, as_completed, new_event_loop, \
     get_event_loop_policy, CancelledError # noqa
 
 try:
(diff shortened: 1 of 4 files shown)


cderici:
[wait_for_idle's behavior when no units in app](https://trello.com/c/gtx0Nmfy/4140-waitforidles-behavior-when-no-units-in-app)

SimonRichardson (approved):

PR #579, opened 2021-11-05
Facade schemas for 2.9.17

SimonRichardson:
The following builds the client facades for 2.9.17. Unfortunately, the
schema generator doesn't currently work with 3.8. The fix is that kind
Any shouldn't be exploded when getting the type.

The clients are just updated with the latest changes.


juju/client/_client1.py
```diff
@@ -2371,11 +2460,12 @@ class CAASFirewallerEmbeddedFacade(Type):
                                                                         'type': 'array'}},
                                              'required': ['results'],
                                              'type': 'object'}},
-     'properties': {'ApplicationCharmURLs': {'description': 'ApplicationCharmURLs '
```
SimonRichardson, line 2374: This is scary that it was around and that it isn't any longer?

juju/client/_client1.py
```diff
@@ -2732,6 +3049,29 @@ async def ApplicationsConfig(self, entities=None):
...
+
+        url : str
+        Returns -> Charm
+        '''
+        if url is not None and not isinstance(url, (bytes, str)):
+            raise Exception("Expected url to be a str, received: {}".format(type(url)))
+
+        # map input types to rpc msg
+        _params = dict()
+        msg = dict(type='CAASFirewaller',
```
SimonRichardson, line 3065: What's the CAASFirewaller doing here?

juju/client/facade.py
```diff
@@ -256,6 +256,8 @@ def __init__(self, schema, defs):
                     self.append((name, rtype))
 
     def do_explode(self, kind):
+        if kind is Any:
+            return False
```
SimonRichardson, line 260: This is the fix for generating the facades.

cderici (approved):

SimonRichardson:
$$merge$$

PR #138, opened 2017-06-09
Add type info to Model.add_machine docs

johnsca:
Resolves #137

PR #580, opened 2021-11-05
[JUJU-138] Streamlining asyncio tasks/events

### LINKED ISSUE #576: Repeated `asyncio` errors
I've been using `pylibjuju` through the `pytest-operator` for testing, and I'm seeing this error *a lot*:

```
ERROR    asyncio:base_events.py:1738 Task exception was never retrieved
future: <Task finished name='Task-30' coro=<Connection._pinger() done, defined at /home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/juju/client/connection.py:502> exception=Exception('No facade Pinger in facades {}')>
  File "/home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/juju/client/connection.py", line 516, in _pinger
    pinger_facade = client.PingerFacade.from_connection(self)
  File "/home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/juju/client/_client.py", line 65, in from_connection
    raise Exception('No facade {} in facades {}'.format(facade_name,
Exception: No facade Pinger in facades {}
```

and

```
ERROR    asyncio:base_events.py:1738 Task exception was never retrieved
future: <Task finished name='Task-402' coro=<WebSocketCommonProtocol.recv() done, defined at /home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/websockets/protocol.py:443> exception=ConnectionClosedOK('code = 1000 (OK), no reason')>
  File "/home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/websockets/protocol.py", line 509, in recv
    await self.ensure_open()
  File "/home/jon/code/canonical/operator-libs-linux/.tox/integration/lib/python3.9/site-packages/websockets/protocol.py", line 803, in ensure_open
    raise self.connection_closed_exc()
websockets.exceptions.ConnectionClosedOK: code = 1000 (OK), no reason
```

These occur consistently on every run. More examples can be found here: https://github.com/jnsgruk/charm-kubernetes-dashboard/runs/4069754517?check_suite_focus=true#step:4:256

cderici:
### Description

A bunch of background noise is happening in the concurrent asyncio events resulting in repeated error logs to build up, despite that nothing in the foreground is failing. These are often:

1. Actual exceptions that are ignored in the main task/thread.
2. Error logs bleeding from other libraries (potentially when trying to handle the ignored exceptions in (1)
(b/c `asyncio` manages their events as well).
3. Tasks that are not gathered properly at teardown.

This PR attempts to streamline all the tasks and events that are happening on our `connection.py` and make sure we're:
* not producing any background exceptions,
* not producing any unnecessary/useless/excess error logs, and 
* tearing down properly.

Hopefully will fix #576 

Jira card [#138](https://warthogs.atlassian.net/browse/JUJU-138)

### QA Steps

In the CI test runs (especially in the ones with failed tests), we should only see the pass info and exception stacktraces that are relevant to the test logic (i.e. there shouldn't be any asyncio related errors in there).

### Notes & Discussion 

* This first commit ensures the `Pinger` facade is available when the `_pinger_task` is created, by moving the `_pinger_task` creation after `_build_facade` calls.

* Also disables the `_pinger_task` completely for `debug_log` connections, because debug_log connections are not doing any RPCs, they directly connect through the WebSocketClientProtocol, so we're not getting any facades from the apiserver (like we do in the case of `_connect_with_login` etc)

* A relevant bug here is https://bugs.python.org/issue36709 : which is about `asyncio.sslproto.SSLProtocol` raising an error whenever the event_loop is closed during an alive websocket connection. So in our `client/connection.py` we avoid keeping the connection alive (by removing `ping_interval=None`).

* The websocket object (_ws) is tightly coupled with the asyncio-managed tasks (e.g. Pinger) as well as the Monitor, so interacting with it directly (e.g. doing conn._ws.close() instead of conn.close()) pushes things out of sync/control. So it shouldn't be used directly (e.g. avoid `conn.ws.close()`, instead use `conn.close()`)

* Functions such as `model.get_controller` returns a newly constructed Controller object that is already connected with the model's own connection parameters. This puts the responsibility of closing out the connection (and destroying the spawned asyncio tasks etc) to the caller. If not done, everyone (asyncio, websockets etc) yells at pylibjuju about lingering tasks and open connections. So we introduce a `ConnectedController` object that when it's used with the `async with` form automatically disconnects after done.

* asyncio loop event_handler is only called on Task exceptions when the Task object is cleared from the memory. But the garbabe collector doesn't clear the Task if we keep a reference for it (e.g. putting it in a neverending Task ---like Pinger) until the very end. So installing our own handler makes sure that the exceptions are retrieved and properly handled and logged regardless of when the Task is being destroyed.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -425,7 +425,10 @@ class Connection:
     async def _recv(self, request_id):
         if not self.is_open:
             raise websockets.exceptions.ConnectionClosed(0, 'websocket closed')
-        return await self.messages.get(request_id)
+        try:
+            return await self.messages.get(request_id)
+        except GeneratorExit:
+            return {}
 
     def _close_debug_log_target(self):
         if self.debug_log_target is not sys.stdout:
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -389,7 +389,6 @@ class Connection:
             max_size=self.max_frame_size,
             server_hostname=server_hostname,
             sock=sock,
-            ping_interval=None
         )), url, endpoint, cacert
 
     async def close(self, to_reconnect=False):
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -690,11 +690,15 @@ class Connection:
         async with monitor.reconnecting:
             await self.close(to_reconnect=True)
             connector = self._connect if self.is_debug_log_connection else self._connect_with_login
-            await connector(
+            res = await connector(
                 [(self.endpoint, self.cacert)]
                 if not self.endpoints else
                 self.endpoints
             )
+            if connector is self._connect_with_login:
+                self._build_facades(res.get('facades', {}))
+                if not self._pinger_task:
+                    self._pinger_task = jasyncio.create_task(self._pinger())
 
     async def _connect(self, endpoints):
         if len(endpoints) == 0:
@@ -752,10 +756,6 @@ class Connection:
         elif not self.is_debug_log_connection and not self._receiver_task:
             self._receiver_task = jasyncio.create_task(self._receiver())
 
-        #  In any type of connection, if we don't have a _pinger_task
-        #  yet, then schedule a new one
-        if not self._pinger_task:
-            self._pinger_task = jasyncio.create_task(self._pinger())
 
         log.debug("Driver connected to juju %s", self.addr)
         self.monitor.close_called.clear()
@@ -799,8 +799,6 @@ class Connection:
         finally:
             if not success:
                 await self.close()
-            else:
-                self._pinger_task = jasyncio.create_task(self._pinger())
 
     async def _connect_with_redirect(self, endpoints):
         try:
@@ -811,6 +809,8 @@ class Connection:
                 raise
             login_result = await self._connect_with_login(e.endpoints)
         self._build_facades(login_result.get('facades', {}))
+        if not self._pinger_task:
+            self._pinger_task = jasyncio.create_task(self._pinger())
 
     def _build_facades(self, facades):
         self.facades.clear()
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -186,7 +186,7 @@ class Monitor:
             return self.DISCONNECTED
 
         # connection cleanly disconnected or not yet opened
-        if not connection.ws:
+        if not connection._ws:
             return self.DISCONNECTED
 
         # close called but not yet complete
@@ -199,7 +199,7 @@ class Monitor:
         else:
             stopped = connection._receiver_task.cancelled()
 
-        if stopped or not connection.ws.open:
+        if stopped or not connection._ws.open:
             return self.ERROR
 
         # everything is fine!
@@ -287,7 +287,7 @@ class Connection:
         # _connect_with_redirect method, but create them here
         # as a reminder that they will exist.
         self.addr = None
-        self.ws = None
+        self._ws = None
         self.endpoint = None
         self.endpoints = None
         self.cacert = None
@@ -392,7 +392,7 @@ class Connection:
         )), url, endpoint, cacert
 
     async def close(self, to_reconnect=False):
-        if not self.ws:
+        if not self._ws:
             return
         self.monitor.close_called.set()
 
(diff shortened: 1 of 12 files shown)


jnsgruk:
Happy to help with QA - I can trigger the same builds but pointing at your branch and see how it goes?

cderici:
Hey @jnsgruk, things seem to be a lot smoother here with the asyncio stuff (there were some rather intricate bugs in there -- see PR notes if you're curious), so if you wanna run your builds against this and help with QA'ing, I think now is a good point to do so :+1:

jnsgruk:
Starting now.... https://github.com/jnsgruk/charm-kubernetes-dashboard/runs/4191343613?check_suite_focus=true

tests/integration/test_crossmodel.py
```diff
@@ -96,9 +96,9 @@ async def test_remove_saas(event_loop):
 async def test_add_relation_with_offer(event_loop):
     async with base.CleanModel() as model_1:
         application = await model_1.deploy(
-            'cs:mysql-58',
+            'ch:mysql',
             application_name='mysql',
-            series='bionic',
+            series='xenial',
```
SimonRichardson, line 101: I'd rather we didn't go backward here. Can we pick a better charm or better series?

jnsgruk:
Looks good! Still got some [facade errors](https://github.com/jnsgruk/charm-kubernetes-dashboard/runs/4191343613?check_suite_focus=true#step:4:261) but the asyncio stuff seems fixed.

tests/integration/test_crossmodel.py
cderici: Oh looks like I mixed up the version names again, for some reason that's a problem for me :) gotta keep a post-it or something

cderici:
> Looks good! Still got some [facade errors](https://github.com/jnsgruk/charm-kubernetes-dashboard/runs/4191343613?check_suite_focus=true#step:4:261) but the asyncio stuff seems fixed.

I'm not sure what's going on about the facade warnings, #579 should've fixed it and I rebased this one on it, but that doesn't look like a complicated fix.

Good to know the asyncio doesn't make any noise anymore, thanks for checking!

juju/client/connection.py
```diff
@@ -287,7 +287,7 @@ async def connect(
         # _connect_with_redirect method, but create them here
         # as a reminder that they will exist.
         self.addr = None
-        self.ws = None
+        self._ws = None
```
SimonRichardson, line 290: This would be classed as a breaking change i.e. we're making something public, private. For now, let's move this back to public.

SimonRichardson (changes requested):
I know users should be using `ws`, but somebody will be, so we need to keep it around for now.

cderici:
!!build!!

SimonRichardson (approved):
Can we add to the release notes that `ws` has become a read-only property.

SimonRichardson:
Also, why are the integration tests failing?

cderici:
!!build!!

cderici:
> Also, why are the integration tests failing?

@SimonRichardson tl;dr - b/c of intermittent errors that are out of scope to this PR, and the bundle overlay stuff that are waiting for Juju release. I also see an asyncio bug I'm working on now (but that doesn't fail a test, it's just an intermediate error producing output, which is the very subject of this PR).

If you wanna see what we should see consistently as a qa here, take a look at the output of [`3.8`](https://jenkins.juju.canonical.com/job/github-integration-tests-pylibjuju-python3.8/250/consoleText) and [`3.6`](https://jenkins.juju.canonical.com/job/github-integration-tests-pylibjuju-python3.6/248/consoleText), that's what we want to see: no outputs/errors about `asyncio` task stuff for both passing and failing tests.

Details if you wanna see:

I see in the logs some errors I wasn't seeing a couple days ago:
- intermittent `http.client.RemoteDisconnected: Remote end closed connection without response`  (grep for `test_deploy_local_bundle_dir` in `3.9` output)
- intermittent mongodb dump while creating a backup (grep for `test_backups` in `3.9` or `3.5`)
- intermittent long standing adding relation with offer problem (that I think I should open a separate issue about it at this point)

The only thing related to this PR is that I see is one intermittent `RuntimeError: Event loop is closed` error there in `3.9` output, seemingly about `tests/integration/test_unit.py::test_run_action`, which means there's still a bug in there about `asyncio` that's evading me. I'm working on it :+1: (the other intermittent errors are just out of scope for this PR)

cderici:
$$merge$$

PR #582, opened 2021-11-08
Added boolean entries to normalize values.

### LINKED ISSUE #581: Unable to set allocate-public-ip constraint
### Summary

When using `model.deploy()` to deploy an application, including the `allocate-public-ip=true` constraint fails with the following error:

```
  File "test_integration.py", line 60, in test_deploy_cluster
    app = await ops_test.model.deploy(
  File "/.tox_env/lib/python3.8/site-packages/juju/model.py", line 1662, in deploy
    return await self._deploy(
  File "/.tox_env/lib/python3.8/site-packages/juju/model.py", line 1897, in _deploy
    app = client.ApplicationDeploy(
  File "/.tox_env/lib/python3.8/site-packages/juju/client/_definitions.py", line 1806, in __init__
    constraints_ = Value.from_json(constraints) if constraints else None
  File "/.tox_env/lib/python3.8/site-packages/juju/client/facade.py", line 640, in from_json
    return cls(**d)
  File "/.tox_env/lib/python3.8/site-packages/juju/client/_definitions.py", line 27110, in __init__
    raise Exception("Expected allocate_public_ip_ to be a bool, received: {}".format(type(allocate_public_ip_)))
Exception: Expected allocate_public_ip_ to be a bool, received: <class 'str'>
```

I believe the issue occurs due to https://github.com/juju/python-libjuju/blob/master/juju/constraints.py#L75-L85 not doing any handling for boolean values.

### Proposed Fix

Extend `normalize_value` with something along the lines of:

```python
if value in ['true', 'True']:
    return True
if value in ['false', 'False']:
    return False
```

I doubt this is going to break any other constraints.

juanmanuel-tirado:
Fixes #581 by including bool values in the `normalize_value` function.


cderici (approved):
We might need to add some more values for yes/no as well, see https://github.com/juju/python-libjuju/issues/581#issuecomment-963274930

PR #592, opened 2021-11-25
Ensure all watchers validate for the Id

SimonRichardson:
Currently all watchers in Juju are expected to ensure that they pass
back an id. This is currently done per watcher, what should actually
happen is we should correctly wire up the calls to the facades, but this
doesn't seem easy. Instead, this approach ensures that all watchers when
doing an RPC call have a Id. If no Id is present raise an error so that
the correct watcher id can be passed in.

### DIFF
--- a/juju/client/_client1.py
+++ b/juju/client/_client1.py
@@ -239,6 +239,20 @@ class AllWatcherFacade(Type):
 
 
 
+    async def rpc(self, msg):
+        '''
+        Patch rpc method to add Id.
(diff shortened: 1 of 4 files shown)


cderici (approved):
I assume the problems in the CI tests are not related. Do you already know what's going on there or should we investigate?

cderici:
!!build!!

SimonRichardson:
$$merge$$

PR #596, opened 2021-12-02
[JUJU-289] Use provided series in deploy if supported

jack-w-shaw:
Previously any provided series was ignored and the default provided by
CharmStore API was always used. Now, if a provided series is supported,
use that instead

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -542,17 +542,26 @@ class CharmhubDeployType:
         ch = Channel('latest', 'stable')
         if channel is not None:
             ch = Channel.parse(channel).normalize()
+
         origin = client.CharmOrigin(source="charm-hub",
                                     architecture=architecture,
                                     risk=ch.risk,
                                     track=ch.track)
-        charm_url, origin = await self.charm_resolver(url, origin)
+
+        charm_url, origin, supported_series = await self.charm_resolver(url, origin)
 
         if app_name is None:
             app_name = url.name
 
+        if series:
+            if series in supported_series:
+                origin.series = series
+                charm_url.series = series
+            else:
+                raise JujuError("Series {} not supported for {}. Only {}".format(series, url, supported_series))
+
         return DeployTypeResult(
-            identifier=charm_url,
+            identifier=str(charm_url),
             app_name=app_name,
             origin=origin,
             is_bundle=origin.type_ == "bundle",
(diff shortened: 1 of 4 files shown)


SimonRichardson (commented):
This looks great, but can we get an integration test?

juju/model.py
```diff
@@ -542,17 +542,26 @@ async def resolve(self, url, architecture, app_name=None, channel=None, series=N
...
 
         if app_name is None:
             app_name = url.name
 
+        if series:
+            if series in supported_series:
+                origin.series = series
+                charm_url.series = series
+            else:
+                raise JujuError(f"Series {series} not supported for {str(url)}. Only {supported_series}")
```
cderici, line 561: I'd change this to a slightly older syntax (because we still're supporting `Python 3.5` - until 3.0).

`"Series % not supported for %s. Only %s" % (series, str(url), supported_series)`

SimonRichardson: We're using `"{}".format()` syntax in pretty much every other place.

tests/integration/test_model.py
```diff
@@ -277,6 +277,30 @@ async def test_deploy_channels_revs(event_loop):
...
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_deploy_from_ch_with_series(event_loop):
+    async with base.CleanModel() as model:
+        charm = 'ch:ubuntu'
+        for series in ['xenial', 'bionic', 'focal']:
+            async with base.CleanModel() as model:
+                app_name = f"ubuntu-{series}"
```
cderici, line 287: :point_up: This as well, is not gonna go well with py3.5

juju/model.py
cderici: Oh that works with py35 too you're right. Though, I've been using the other syntax, I should probably go ahead and make a consistency update at some point

cderici (approved):

jack-w-shaw:
$$merge$$

tests/integration/test_model.py
```diff
@@ -277,6 +277,30 @@ async def test_deploy_channels_revs(event_loop):
         ]
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_deploy_from_ch_with_series(event_loop):
+    async with base.CleanModel() as model:
+        charm = 'ch:ubuntu'
+        for series in ['xenial', 'bionic', 'focal']:
+            async with base.CleanModel() as model:
```
jack-w-shaw, line 286: Oops

jack-w-shaw:
$$merge$$

PR #597, opened 2021-12-03
[JUJU-292] Update the charms in the tests to use Charmhub

### LINKED ISSUE #564: tests using charmstore should be updated to use charmhub
We have intermittent connection errors on the CI tests that are seem to be originating from accessing to the charmstore (http).

Access to the charmstore from pylibjuju is done by [theblues](https://github.com/juju/theblues) library and it doesn't seem to be maintained anymore.

We should update the tests to use charms from charmhub instead.

If we want to keep making sure that the access to the charmstore is working, then we should have a separate `test_charmstore.py`, just like the [test_charmhub.py](https://github.com/juju/python-libjuju/blob/master/tests/integration/test_charmhub.py).

cderici:
#### Description

This updates the charms, bundles & overlays used in the `model` and `crossmodel` integration tests to use the CharmHub instead of CharmStore.

Fixes #564

#### QA Steps

`tox -e integration -- tests/integration/test_model.py`
and
`tox -e integration -- tests/integration/test_crossmodel.py`

should be consistently passing.


cderici:
!!build!!

juanmanuel-tirado (approved):
R2GO

Finally green tests :)

cderici:
$$merge$$

PR #612, opened 2022-01-04
Fallback to 'local-fan'

dparv:
If 'public' or 'local-cloud' are missing from the unit data, fallback to
'local-fan', in case of containers using solely fan networking and no
provider network.

Fixes juju#611

### DIFF
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -239,6 +239,8 @@ class Machine(model.ModelEntity):
             scope = address['scope']
             if scope == 'public' or scope == 'local-cloud':
                 return address['value']
+            if scope == 'local-fan':
+                return address['value']
         return None
 
     @property
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -234,11 +234,17 @@ class Machine(model.ModelEntity):
 
         May return None if no suitable address is found.
         """
-        return_addresses = []
         addresses = self.safe_data['addresses'] or []
-        for address in addresses:
+        return_addresses = []
+        ordered_addresses = []
+        ordered_scopes = ['public', 'local-cloud', 'local-fan']
+        for scope in ordered_scopes:
+            for address in addresses:
+                if scope == address['scope']:
+                    ordered_addresses.append(address)
+        for address in ordered_addresses:
             scope = address['scope']
-            for check_scope in ['public', 'local-cloud', 'local-fan']:
+            for check_scope in ordered_scopes:
                 if scope == check_scope:
                     return_addresses.append(address['value'])
         if return_addresses:
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -234,13 +234,15 @@ class Machine(model.ModelEntity):
 
         May return None if no suitable address is found.
         """
+        return_addresses = []
         addresses = self.safe_data['addresses'] or []
         for address in addresses:
             scope = address['scope']
-            if scope == 'public' or scope == 'local-cloud':
-                return address['value']
-            if scope == 'local-fan':
-                return address['value']
+            for check_scope in ['public', 'local-cloud', 'local-fan']:
+                if scope == check_scope:
+                    return_addresses.append(address['value'])
+        if return_addresses:
+            return return_addresses[0]
         return None
 
     @property
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -235,7 +235,6 @@ class Machine(model.ModelEntity):
         May return None if no suitable address is found.
         """
         addresses = self.safe_data['addresses'] or []
-        return_addresses = []
         ordered_addresses = []
         ordered_scopes = ['public', 'local-cloud', 'local-fan']
         for scope in ordered_scopes:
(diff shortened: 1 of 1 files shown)


juanmanuel-tirado:
This will have to wait until we have further discussion on how to proceed with dns names.

juju/machine.py
```diff
@@ -239,6 +239,8 @@ def dns_name(self):
             scope = address['scope']
             if scope == 'public' or scope == 'local-cloud':
```
cderici, line 240: Python's logic ops like `or` are short-circuiting (e.g. `1 or 1/0 or 3` will produce 1), so it might make it cleaner if you just add `or scope == 'local-fan'` at the end of this line.

juju/machine.py
```diff
@@ -239,6 +239,8 @@ def dns_name(self):
...
                 return address['value']
+            if scope == 'local-fan':
```
manadart, line 242: I don't believe there is a guaranteed order for these addresses, so this makes it possible to return a `local-fan` address for a non-container machine if Fan is the container networking method in the model.

We should only be returning the Fan address is no public or local-cloud addresses are present.

manadart (changes requested):

juju/machine.py
dparv: Indeed this is the reason I put it after. So would that patch be acceptable?

manadart: For consistent results we should:
- Filter the list for public/local-cloud/local-fan scopes.
- Return `None` if empty.
- Sort the list by scope in order of public/local-cloud/local-fan.
- Return list index 0.

dparv: I've changed some things for the above recommendations with the latest commit.

manadart: We're missing the sort to ensure that the item at index zero prioritised by scope.
Something like this should do it (untested):
```diff
diff --git a/juju/machine.py b/juju/machine.py
index 44a2a6a..583115d 100644
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -1,5 +1,6 @@
 import ipaddress
 import logging
+from operator import attrgetter

 import pyrfc3339

@@ -234,12 +235,16 @@ class Machine(model.ModelEntity):

         May return None if no suitable address is found.
         """
-        addresses = self.safe_data['addresses'] or []
-        for address in addresses:
-            scope = address['scope']
-            if scope == 'public' or scope == 'local-cloud':
-                return address['value']
-        return None
+        addresses = self.safe_data['addresses']
+        if not addresses:
+            return None
+
+        valid_scopes = ['public', 'local-cloud', 'local-fan']
+        filtered = filter(lambda a: a['scope'] in scopes)
+        if not filtered:
+            return None
+
+        return sorted(filtered, key=valid_scopes.index(attrgetter('scope'))[0]

     @property
     def hostname(self):
```

dparv: I have tried to test and make this work but I couldn't. But I get the idea, because there might be multiple public or local-cloud addresses which might cause the indices to become [public, local-cloud, public, local-fan, etc.]. We can use something simpler, like (this I have tested):

```diff
diff --git a/juju/machine.py b/juju/machine.py
index 44a2a6a..150899d 100644
--- a/juju/machine.py
+++ b/juju/machine.py
@@ -235,10 +235,20 @@ class Machine(model.ModelEntity):
         May return None if no suitable address is found.
         """
         addresses = self.safe_data['addresses'] or []
-        for address in addresses:
+        return_addresses = []
+        ordered_addresses = []
+        ordered_scopes = ['public', 'local-cloud', 'local-fan']
+        for scope in ordered_scopes:
+            for address in addresses:
+                if scope == address['scope']:
+                    ordered_addresses.append(address)
+        for address in ordered_addresses:
             scope = address['scope']
-            if scope == 'public' or scope == 'local-cloud':
-                return address['value']
+            for check_scope in ordered_scopes:
+                if scope == check_scope:
+                    return_addresses.append(address['value'])
+        if return_addresses:
+            return return_addresses[0]
         return None

```

juju/machine.py
```diff
@@ -235,10 +235,20 @@ def dns_name(self):
...
         """
         addresses = self.safe_data['addresses'] or []
-        for address in addresses:
+        return_addresses = []
+        ordered_addresses = []
+        ordered_scopes = ['public', 'local-cloud', 'local-fan']
+        for scope in ordered_scopes:
+            for address in addresses:
+                if scope == address['scope']:
+                    ordered_addresses.append(address)
```
manadart, line 244: Since you're now looking for the scopes in order, and exhausting the addresses for each, you could return the first one you find here and drop the rest up until `return None`.

Happy to approve then.

manadart (changes requested):

juju/machine.py
dparv: Changed as requested.

manadart (approved):
Thanks for working on this. Much appreciated.

cderici:
!!build!!

cderici:
@juanmanuel-tirado The problems in the CI are known intermittent problems (i.e. not related to this PR), you can go ahead and merge this :+1:

PR #613, opened 2022-01-04
[JUJU-376] `wait_for_idle` to support scale down

### LINKED ISSUE #608: `wait_for_idle` does not work for scale down
`wait_for_idle` immediately returns if used for scaledown when all the units are already active/idle.
```python
await ops_test.model.applications[app_name].scale(scale=1)                                     
await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=1000)
# wait_for_idle immediately returns and the test function completes before any scale down took place
```

It is also not possible to use `wait_for_units` to provide a smaller number, because of the conditional:

https://github.com/juju/python-libjuju/blob/8d62315acc978b6b0375931440ec9b85bfaa5a49/juju/model.py#L2635-L2637

The only workaround I am aware of is to use `block_until` with a precalculated target unit count.

cderici:
#### Description

This adds a new parameter `wait_for_exact_units` to the `wait_for_idle` function to wait for an exact number of units.

Fixes #608 

#### QA Steps

```sh
tox -e integration -- tests/integration/test_model::test_wait_for_idle_with_exact_units
```

```sh
tox -e integration -- tests/integration/test_model::test_wait_for_idle_with_exact_units_scale_down
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2554,7 +2554,7 @@ class Model:
 
     async def wait_for_idle(self, apps=None, raise_on_error=True, raise_on_blocked=False,
                             wait_for_active=False, timeout=10 * 60, idle_period=15, check_freq=0.5,
-                            status=None, wait_for_units=1):
+                            status=None, wait_for_units=1, wait_for_exact_units=-1):
         """Wait for applications in the model to settle into an idle state.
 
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado (approved):
LG2M

cderici:
$$merge$$

PR #629, opened 2022-02-04
[JUJU-566] `safe_data` checks the connection before access in ModelEntity

### LINKED ISSUE #627: [JUJU-566] `safe_data` needs to check connection status
Currently, even after a `Model.disconnect()`, some data like the unit public addresses etc are still accessible from the `safe_data` method. That data should only be accessible when the connection is still alive.

cderici:
#### Description

`safe_data` is a property access method on `ModelEntity` objects (such as `Unit`s). This PR makes every model entity to check their model's connection before allowing access to the underlying data. If the model is disconnected, any access to data in the associated model entities will raise a `DeadEntityException`.

Fixes #627 

#### QA Steps

```sh
tox -e integration -- tests/integration/test_model.py::test_disconnect_clears_safe_data
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -363,11 +363,12 @@ class ModelEntity:
     def safe_data(self):
         """The data dictionary for this entity.
 
-        If this `ModelEntity` points to the dead state, it will
-        raise `DeadEntityException`.
+        If this `ModelEntity` points to the dead state, or the model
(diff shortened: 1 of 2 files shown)


SimonRichardson (commented):
Is this classified as a breaking change, esp. if users are accessing the model after a disconnect?

cderici:
!!build!!

PR #633, opened 2022-02-09
[JUJU-573] Fix charm resolution for Juju 2.8.11

### LINKED ISSUE #623: Can't deploy charms with juju 2.8.11
Deploying charms seems to be broken if using juju version 2.8.11. This is my setup:

```console
$ pip3 show juju | grep -i version
Version: 2.9.6
$ juju version
2.8.11-bionic-amd64
$ juju controllers --format json | jq -r '.controllers["localhost-localhost"]["agent-version"]'
2.8.11
```

And this is the minimal test I was able to come up with:

```python
import sys
import asyncio
import juju.controller

async def controller():
    c = juju.controller.Controller()
    await c.connect_current()
    return c

async def model(controller):
    return await controller.add_model('test')

async def deploy_app(model):
    await model.deploy('ubuntu')

loop = asyncio.get_event_loop()

c = loop.run_until_complete(controller())
m = loop.run_until_complete(model(c))
loop.run_until_complete(m.deploy(sys.argv[1], channel='stable'))

loop.run_until_complete(m.disconnect())
loop.run_until_complete(c.destroy_model('test'))
loop.run_until_complete(c.disconnect())
```

Then this happens, which I believe is expected since juju 2.8 can't deploy charmhub charms:

```console
$ python3 test.py ch:ubuntu       
  File "test.py", line 18, in <module>
    loop.run_until_complete(m.deploy(sys.argv[1], channel='stable'))                       
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File "/home/faq/lp1943748/charm-hw-health/.build/hw-health/.tox/func/lib/python3.6/site-packages/juju/model.py", line 1614, in deploy
    res = await self.deploy_types[str(url.schema)].resolve(url, architecture, application_name, channel, series, entity_url)
  File "/home/faq/lp1943748/charm-hw-health/.build/hw-health/.tox/func/lib/python3.6/site-packages/juju/model.py", line 551, in resolve
    charm_url_str, origin, supported_series = await self.charm_resolver(url, origin)
  File "/home/faq/lp1943748/charm-hw-health/.build/hw-health/.tox/func/lib/python3.6/site-packages/juju/model.py", line 1714, in _resolve_charm
    raise JujuError("resolve charm")                                
juju.errors.JujuError: resolve charm                                               
```

The same thing happens if no prefix is specified (ie, plain `ubuntu`). I don't know if this is expected or not. I would expect it to try charmhub and fallback to the charmstore if that doesn't work, but I'm not sure.

But then this also happens:

```console
$ python3 test.py cs:ubuntu
  File "test.py", line 18, in <module>                                             
    loop.run_until_complete(m.deploy(sys.argv[1], channel='stable'))
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()                       
  File "/home/faq/lp1943748/charm-hw-health/.build/hw-health/.tox/func/lib/python3.6/site-packages/juju/model.py", line 1644, in deploy       
    charm_origin = add_charm_res.charm_origin
AttributeError: 'dict' object has no attribute 'charm_origin'                                                                                  
```

which I don't think it's expected.

cderici:
#### Description

This is in response to some charm deployment issues with `Juju 2.8.11`. It changes two things:

1 - It patches the deployment of `CharmStore` charms
2 - Normally pylibjuju assumes the `CharmHub` when it receives a charm url without a prefix. This change adds an override for the case where the `CharmsFacade` version is too low to resolve the charm.

Fixes #623 

#### QA Steps

You'll need the [juju 2.8.11](https://github.com/juju/juju/archive/refs/tags/juju-2.8.11.tar.gz), and to put the following script in a file (slightly modified version of what is provided in the reporting issue #623):

```python
import sys
import asyncio
import juju.controller

async def controller():
    c = juju.controller.Controller()
    await c.connect_current()
    return c

async def model(controller):
    return await controller.add_model('test')

loop = asyncio.get_event_loop()

try:
    c = loop.run_until_complete(controller())
    m = loop.run_until_complete(model(c))
    loop.run_until_complete(m.deploy(sys.argv[1], channel='stable'))

    loop.run_until_complete(m.disconnect())
    loop.run_until_complete(c.destroy_model('test'))
    loop.run_until_complete(c.disconnect())
    print("success")
except:
    loop.run_until_complete(c.destroy_model('test'))
    loop.run_until_complete(c.disconnect())
    raise

```

Bootstrap with Juju `2.8.11` and run the file containing the script above like `python3 test.py ubuntu`, you should see `success`.

#### Notes & Discussion

`_resolve_charm` method literally has a failing check for this because lower versions (such as Juju 2.8) cannot deploy from the CharmHub in the first place, so it's safe to put in this override and try to get the charm from the CharmStore instead of failing.

### DIFF
--- a/juju/url.py
+++ b/juju/url.py
@@ -42,17 +42,11 @@ class URL:
 
         if Schema.LOCAL.matches(u.scheme):
             c = URL(Schema.LOCAL, name=u.path)
+        elif Schema.CHARM_STORE.matches(u.scheme) or \
+             (u.scheme == "" and Schema.CHARM_STORE.matches(default_store)):
+            c = parse_v1_url(Schema.CHARM_STORE, u, s)
         else:
-            cs_match = Schema.CHARM_STORE.matches(u.scheme)
-            default_is_cs = Schema.CHARM_STORE.matches(default_store)
-
-            ch_match = Schema.CHARM_HUB.matches(u.scheme)
-            default_is_ch = Schema.CHARM_HUB.matches(default_store)
-
-            if cs_match or (not ch_match and default_is_cs):
-                c = parse_v1_url(Schema.CHARM_STORE, u, s)
-            elif ch_match or default_is_ch:
-                c = parse_v2_url(u, s)
+            c = parse_v2_url(u, s)
 
         if not c or not c.schema:
             raise JujuError("expected schema for charm or bundle URL {}".format(u))
--- a/juju/url.py
+++ b/juju/url.py
@@ -29,11 +29,12 @@ class URL:
         self.architecture = architecture
 
     @staticmethod
-    def parse(s):
+    def parse(s, force_v1=False):
         """parse parses the provided charm URL string into its respective
             structure.
 
             A missing schema is assumed to be 'ch'.
+            If force_v1 is True, then it is assumed to be 'cs'.
 
         """
         u = urlparse(s)
@@ -42,7 +43,7 @@ class URL:
 
         if Schema.LOCAL.matches(u.scheme):
             c = URL(Schema.LOCAL, name=u.path)
-        elif Schema.CHARM_STORE.matches(u.scheme):
+        elif force_v1 or Schema.CHARM_STORE.matches(u.scheme):
             c = parse_v1_url(Schema.CHARM_STORE, u, s)
         else:
             c = parse_v2_url(u, s)
--- a/juju/url.py
+++ b/juju/url.py
@@ -43,7 +43,7 @@ class URL:
         if Schema.LOCAL.matches(u.scheme):
             c = URL(Schema.LOCAL, name=u.path)
         elif Schema.CHARM_STORE.matches(u.scheme) or \
-             (u.scheme == "" and Schema.CHARM_STORE.matches(default_store)):
+                (u.scheme == "" and Schema.CHARM_STORE.matches(default_store)):
             c = parse_v1_url(Schema.CHARM_STORE, u, s)
         else:
             c = parse_v2_url(u, s)
--- a/juju/url.py
+++ b/juju/url.py
@@ -29,12 +29,11 @@ class URL:
         self.architecture = architecture
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado (approved):
LGTM

cderici:
> I'm not fond of the `URL.parse` having `force_v1`. I think we should have `URL.defaultStore` that we can then pass into `URL.parse`. If the return type of `URL.defaultStore` is None, then fallback to charmhub.

Agreed, I think that's a better way of handling this, though parse being a static method doesn't help. Check out the changes. First I actually put a `URL.default_store` variable and changed that but I don't think modifying a global static class constant is a good idea, so I just made a new variable `default_store` to parse (defaulting to `Schema.CHARM_HUB`).

cderici:
!!build!!

juju/url.py
```diff
@@ -42,12 +42,19 @@ def parse(s):
...
-        if not c.schema:
+            ch_match = Schema.CHARM_HUB.matches(u.scheme)
+            default_is_ch = Schema.CHARM_HUB.matches(default_store)
+
+            if cs_match or (not ch_match and default_is_cs):
+                c = parse_v1_url(Schema.CHARM_STORE, u, s)
+            elif ch_match or default_is_ch:
+                c = parse_v2_url(u, s)
+
+        if not c or not c.schema:
```
SimonRichardson, line 57: I think we should write this like the following:

```python
if Schema.LOCAL.matches(u.scheme):
            c = URL(Schema.LOCAL, name=u.path)
elif Schema.CHARM_STORE.matches(u.scheme) or (u.Scheme == "" and default_store == Schema.CHARM_STORE):
            c = parse_v1_url(Schema.CHARM_STORE, u, s)
else:
            c = parse_v2_url(u, s)
```

SimonRichardson (approved):

cderici:
$$merge$$

PR #648, opened 2022-03-10
[JUJU-731] Subordinate charm num unit

### LINKED ISSUE #639: model.deploy(subordinate_charm) should work out of the box
Not sure this is more of an issue or a feature request, but here's my thought.

`juju deploy subordinate_charm`  and the charm will be automagically deployed with scale=0/0.
While if I do:
`model.deploy(subordinate_charm)`
it will fail citing `subordinate application must be deployed without units`. 

User is then required to do: `deploy(subordinate_charm, num_units=0)` and it will work.

Two problems with this:
1) I would expect the behaviour of `ops_test.deploy` to match that of `juju deploy`
2) Given the error message, we do know exactly how to fix the issue. Why don't we then? A dynamic default checking `if charm.is_subordinate` would do.

cderici:
#### Description

Subordinate charms are by definition don't require additional units. So num-units should be 0 at deploy.

Normally in libjuju num_units defaults to 1, this change lets libjuju to check the metadata from charmhub to see if we're deploying a subordinate charm. If that's the case then we adjust the num-units to 0 to be able to deploy it without any problems from Juju's side.

Fixes #639 

#### QA Steps

```sh
tox -e integration -- tests/integration/test_charmhub.py::test_subordinate_charm_zero_units
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1622,6 +1622,16 @@ class Model:
                     resources = await self._add_charmhub_resources(res.app_name,
                                                                    identifier,
                                                                    add_charm_res.charm_origin)
+                    charm_info = await self.charmhub.info(res.app_name)
+                    is_subordinate = False
+                    try:
+                        is_subordinate = charm_info.result.charm.subordinate
+                    except AttributeError:
+                        log.warning('CharmHub.Info : unable to retrieve the subordinate information')
+                    if is_subordinate:
+                        if num_units > 1:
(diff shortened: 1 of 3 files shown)


juanmanuel-tirado (approved):
LG2M

cderici:
!!build!!

cderici:
!!build!!

cderici:
!!build!!

cderici:
!!build!!

PR #659, opened 2022-03-29
[JUJU-843] Attach-resource to check if given binary file

### LINKED ISSUE #654: Problem when trying to attach local binary resource
Hi Juju team!
I’m using this lib through [pytest-operator](https://github.com/charmed-kubernetes/pytest-operator) in my charm integration tests where I add one `.tar.gz` as a resource in the deployment.

What happens is that I couldn’t find a way to attach the resource using the `attach_resource` method. I’m getting an error when I try to pass the resource as a binary content to that method:
```python
resources = {"patroni": "patroni.tar.gz"}
app = await ops_test.model.deploy(
    charm, resources=resources, application_name=application_name, series=series
)
with open("patroni.tar.gz", 'rb') as f:
    app.attach_resource('patroni', "patroni.tar.gz", f)
```
```sh
  File "/home/neppel/Canonical/postgresql-operator/tests/integration/test_charm.py", line 60, in test_deploy
    app.attach_resource('patroni', "patroni.tar.gz", f)
  File "/home/neppel/Canonical/postgresql-operator/.tox/integration/lib/python3.8/site-packages/juju/application.py", line 454, in attach_resource
    headers['Content-Sha384'] = hashlib.sha384(bytes(data, 'utf-8')).hexdigest()
TypeError: encoding without a string argument
```

I could attach the resource only when using the `juju` function:
```sh
await ops_test.juju("attach-resource", "postgresql-focal", "patroni=patroni.tar.gz")
```
Am I doing something wrong when using `attach_resource` from the lib?

Some more details:
* It's a machine charm.
* I'm testing locally with a local charm (not present in charmhub).
* The test works if I use a plain text file instead of a binary file (it looks like that the lib code is trying to convert the content to bytes to generate a sha checksum, which breaks because the content is already as bytes).

cderici:
#### Description

The `attach-resource` method on pylibjuju assumes the resource read is plain text, so it borks when it gets a binary file as a resource.

Fixes #654 

#### QA Steps

```sh
tox -e integration -- tests/integration/test_model.py::test_attach_resource
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -441,7 +441,8 @@ class Application(model.ModelEntity):
 
         headers['Content-Type'] = 'application/octet-stream'
         headers['Content-Length'] = len(data)
-        headers['Content-Sha384'] = hashlib.sha384(bytes(data, 'utf-8')).hexdigest()
+        data_bytes = data if isinstance(data, bytes) else bytes(data, 'utf-8')
+        headers['Content-Sha384'] = hashlib.sha384(data_bytes).hexdigest()
 
         file_name = str(file_name)
         if not file_name.startswith('./'):
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -660,6 +660,9 @@ async def test_attach_resource(event_loop):
         with open(str(charm_path / 'test.file')) as f:
             app.attach_resource('file-res', 'test.file', f)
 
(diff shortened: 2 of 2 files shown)


SimonRichardson (approved):

cderici:
!!build!!

cderici:
$$merge$$

PR #660, opened 2022-03-30
[JUJU-796] Add relate method and deprecate add-relation

jack-w-shaw:
#### Description

This is to follow the pattern we are following going ahead with Juju.
change the name of add-relation to relate, but keep add-relation, which
calls relate for backwards compatibility

This method should then be removed at the next major version bump

See this PR for more details
https://github.com/juju/juju/pull/13907

#### QA Steps

```sh
tox -e integration
tox -e unit
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -107,6 +107,13 @@ class Application(model.ModelEntity):
         return tag.application(self.name)
 
     async def add_relation(self, local_relation, remote_relation):
+        """
+        .. deprecated:: 2.9.9
+           Use ``relate()`` instead
+        """
+        return await self.relate(local_relation, remote_relation)
+
+    async def relate(self, local_relation, remote_relation):
         """Add a relation to another application.
 
         :param str local_relation: Name of relation on this application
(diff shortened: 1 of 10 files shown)


juju/application.py
```diff
@@ -109,6 +109,18 @@ def tag(self):
     async def add_relation(self, local_relation, remote_relation):
         """Add a relation to another application.
 
+        DEPRCATED: Use `relate` instead
```
juanmanuel-tirado, line 112: typo: DEPRECATED

juanmanuel-tirado:
I label this as "do-not-merge" until a we have a consistent message in the juju CLI.

juju/application.py
jack-w-shaw: Good spot, fixed

I have also change formatting to be consistent with the rest of pylibjuju

cderici:
!!build!!

cderici (changes requested):
Other than my comments about the duplication in tests, this lgtm, the CI failures seem to be intermittent glitches, I think we're good there too :+1:

cderici (approved):
This lgtm. The CI fails are unrelated to this PR, so you're clear for landing.

cderici:
I just ran the CI manually for this one and all the tests pass, this is clear for landing :+1:

PR #671, opened 2022-04-19
[JUJU-981] Get series from deployed app instead of metadata when charm upgrade

### LINKED ISSUE #668: Cannot specify series for charm upgrade
In application.py there exists the `local_refresh` method which allows for upgrading a charm from the filesystem.  Unfortunately, this method does not allow for a series to be specified.  Instead it uses the bundle.py `get_charm_series` method, which simply opens the metadata.yaml file and returns the first series found in the list.  As there can be multiple series present, the first one present might not match the series of the unit to which the application is deployed, resulting in an error.

There should be a method to specify the series when executing `local_refresh`.

cderici:
#### Description

Charmhub is modeling the charm data separately from where the charm is deployed, therefore the series is not a part of the metadata anymore. So instead of getting the series from the metadata we now get it from the deployed application first, then we try the model config, if everything fails then we look at the metadata as a last resort (mostly to keep backwards ompatibility).

Fixes #668 

#### QA Steps

The charm used for the integration test `test_upgrade_local_charm` has been updated to simulate what's happening in #668 (now it has multiple series where the first one is not the one that the charm was initially deployed on), so along with the rest of the CI tests, the following should pass with this change (which fails without the change):

```sh
tox -e integration -- tests/integration/test_application.py::test_upgrade_local_charm
```

#### Notes & Discussion

I initially thought that this would be a breaking change (needs a version bump), however, we're still keeping the old way of getting the series in case the new ways fail, so this shouldn't break anything.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -339,6 +339,18 @@ class Application(model.ModelEntity):
         log.debug("Unexposing %s", self.name)
         return await app_facade.Unexpose(application=self.name)
 
+    async def get_series(self):
+        """Return the series on which the application is deployed
+
+        :return: str series
+        """
+        app_facade = self._facade()
+
+        log.debug(
+            'Getting series for %s', self.name)
+
+        return (await app_facade.Get(application=self.name)).series
+
     async def get_config(self):
         """Return the configuration settings dict for this application.
         """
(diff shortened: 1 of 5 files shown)


juanmanuel-tirado (approved):
LG2M

cderici:
All green on the manual CI run, clear to land :+1:

PR #687, opened 2022-07-04
Replace deprecated juju.loop() calls from examples and documentation

ittner:


juanmanuel-tirado:
I will merge this PR independently of the ongoing issues with the integration tests. That has to be addressed separately.

PR #689, opened 2022-07-15
[JUJU-1439] Initial fixes for `test_model` to pass with juju 3.0

cderici:
#### Description

This PR is the first of a bunch of PRs to enable pylibjuju to support Juju 3.0. 
In particular, the aim of this PR is to have all the tests in the  `test_model.py` to pass, which constitutes a significant portion of libjuju's operational logic.

Includes 12 commits. The summary of the changes is:
- Facade versions are updated with the new schema, and the new clients are generated.
- Changed incorrect uses of `ClientFacade` for a bunch of functions to use their respective correct Facades.
- Update the tests/bundles that are using charmstore charms to use the charmhub instead.
- There's a couple of spots where we needed `CharmHub.Info()` to get some basic charm info to deploy stuff, however, since CharmHub facade doesn't exist anymore in juju 3.0, I replaced it with a simple api call to the charmhub itself to get the necessary info. 

#### QA Steps

All the tests in `test_model.py` should pass on a `juju 3.0` controller.

`tox -e integration -- tests/integration/test_model.py`

#### Notes & Discussion

- This PR targets the `juju-3.0-compatibility` branch, not the main branch.
- Currently the tests in the CI don't make much sense, since the CI is being run against the latest stable juju (which is still 2.9). I expect most of them to fail until we address the problem of supporting both 2.9 and 3.0 at the same time.


juanmanuel-tirado (approved):
LG2M :+1:

PR #691, opened 2022-07-20
[JUJU-1464] More fixes for 3.0 compatibility

cderici:
#### Description

This PR completes the fixes for 3.0 compatibility. With this, `pylibjuju` should be able to work with `juju 3.0`.

Includes 8 commits. The summary of the changes is:

* `ClientFacade` -> `CharmsFacade` for upgrades
* Correct facade functions for setting and unsetting application config.
* Added a call to charmhub api to check for subordinate
* Changes to use the correct facade functions with correct input/result parameters
* Small fixes for tests

#### QA Steps

All the integration tests should pass on a `juju 3.0` controller.

```sh
make test
```

#### Notes & Discussion

* I observe in my local runs that some tests get stuck, running them individually makes them pass, there might be intermittent issues, but this is not related to 3.0 compatibility, a further investigation about that might be done later on.
* Again, this targets the `juju-3.0-compatibility` branch. Next step on that is to make sure that the `pylibjuju` can work with both `2.9` and `3.0` facades at the same time.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -619,7 +619,7 @@ class Application(model.ModelEntity):
         if switch is not None and revision is not None:
             raise ValueError("switch and revision are mutually exclusive")
 
-        client_facade = client.ClientFacade.from_connection(self.connection)
+        charms_facade = client.CharmsFacade.from_connection(self.connection)
         resources_facade = client.ResourcesFacade.from_connection(
             self.connection)
         app_facade = self._facade()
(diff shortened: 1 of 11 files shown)


juanmanuel-tirado:
None of the github integration tests pass. This does not have to be addressed to pass this PR but we have to take it into consideration.

```
00:43:43 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
```

juanmanuel-tirado:
As mentioned in the description some of the tests get stuck when run in batches. We can consider this PR to be valid. However, concurrency seems to be compromised and deserves some investigation.

juanmanuel-tirado (approved):
LG2M

cderici:
> None of the github integration tests pass. This does not have to be addressed to pass this PR but we have to take it into consideration.

This is to be expected, since the CI tests run against juju stable (i.e. `2.9`), it'll be fine after we have support for both 2.9 and 3.0 at the same time. Note that the integration test `github-integration-tests-pylibjuju-edge` is passing (that's the one with juju edge).

juanmanuel-tirado:
That's correct. But see the pasted error there, that's a dependency issue.

PR #692, opened 2022-07-22
[JUJU-1457] Merge 3.0 compatibility branch onto master

cderici:
#### Description

This PR introduces the mechanism to have both the old and the new clients generated from `2.9` and `3.0` schema respectively. Based on the `server_version'` information within the connection, correct facade is picked from the correct client among the correct client group. With this, `pylibjuju` should be able to work with both `juju 2.9` and `juju 3.0`.

This PR also merges the `juju-3.0-compatibility` branch onto the `master`.

Most important changes include:
- we have both the clients generated from `juju 3.0` schema, and from the `juju 2.9` schema. The `facade` code is updated to generate the `_client` code to have both sets of clients and facades inside, and whenever a facade is needed, to dynamically pick the correct facade from the correct client based on the version set in the underlying connection with juju.
- `client.py` module is no longer a regular Python module, it's a class instance that acts like a module within Python runtime. Based on the juju version, it pulls all the bindings (`_client`s and the `_definition`s) dynamically from respective correct modules, which are named `_client` and `_definitions` for `juju 3.0`, and `_2_9_client` and `_2_9_definitions` for `juju 2.9`. So whenever in the code we pull a binding from the client, like `client.CharmOrigin`, the `CharmOrigin` object we get comes from the correct `_definitions` module.

#### QA Steps

All the integration tests should pass here on CI, and also on a `juju 3.0` controller.

```sh
make test
```

#### Notes & Discussion

- Everything on the CI tests should pass except the intermittent stuff we already know about, and also some charmhub stuff that we observed before I started working on this.

- There will be some tests failing for sure, I couldn't check all the tests yet, so I'm labeling this as `do not merge` until we see all green (or almost all green).


cderici:
Some notes on the tests:

- [ ] The regular integration tests are against `juju latest/stable` (i.e. `2.9`)
- [ ] The `github-integration-tests-pylibjuju-edge` one is against `juju latest/edge` (i.e. `3.0-beta2`)

So both of them passing means that with this change the `python-libjuju` will be working correctly with both the APIs :+1: 

(the `github-integration-tests-pylibjuju-python3.10` job didn't fully run for some reason, I'm pretty sure it's just a glitch on the jenkins side.)

juanmanuel-tirado (approved):
LG2M

Good job :+1:

PR #697, opened 2022-08-01
[JUJU-1534] Fix `model.connect_current()`

cderici:
#### Description

This PR fixes the `model.connect_current()` call which currently fails at the after_connect where we pull up the model info. The reason of the failure is that we neither have a `model_name`, nor a `model_uuid` for the controller to pull up the info. To solve it, this change gets the `model_uuid` from the connector during connect (if we don't use the endpoints, in which case we don't have a problem to begin with).

#### QA Steps

This change also introduces an integration test, namely, `test_connect_current`, so the following should pass.

```
tox -e integration -- tests/integration/test_model.py::test_connect_current
```

### DIFF
--- a/examples/config.py
+++ b/examples/config.py
@@ -37,12 +37,12 @@ async def main():
     # update and check app config
     await ubuntu_app.set_config({'tuning-level': 'fast'})
     config = await ubuntu_app.get_config()
-    assert(config['tuning-level']['value'] == 'fast')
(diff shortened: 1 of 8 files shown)


juanmanuel-tirado (approved):
LG2M

juanmanuel-tirado:
Some tests are still running random issues. I will ignore and proceed with the merge.

PR #698, opened 2022-08-01
[JUJU-1542] Fix run actions on units

cderici:
#### Description

This changes the function that's used for running an action on a unit from `Action.Enqueue` to `Action.EnqueueOperation`, because it's the most up to date one and also the former one is deprecated in `Action v7` facade.

- [ ] Should be merged after #697 has landed.

#### QA Steps

The qa steps I followed were with a k8s model. So get a controller on a k8s cloud. 
The following then should work fine.

```python
async def _get_password():
    model = Model()
    await model.connect()
    await model.deploy('zinc-k8s')
    await model.wait_for_idle(status="active")

    unit = model.applications['zinc-k8s'].units[0]
    action = await unit.run_action("get-admin-password")
    results = await action.wait()
    print(results["admin-password"])
    await model.disconnect()
```

#### Notes & Discussion

* Maybe we should also add an integration test with an example action to make sure we won't regress from this in the future.

### DIFF
--- a/juju/action.py
+++ b/juju/action.py
@@ -7,4 +7,4 @@ class Action(model.ModelEntity):
         return self.data['status']
 
     async def wait(self):
-        return await self.model.wait_for_action(self.id)
+        return await self.model.get_action_output(self.id)
--- a/juju/unit.py
+++ b/juju/unit.py
@@ -164,19 +164,19 @@ class Unit(model.ModelEntity):
 
         log.debug('Starting action `%s` on %s', action_name, self.name)
 
-        res = await action_facade.Enqueue(actions=[client.Action(
+        res = await action_facade.EnqueueOperation(actions=[client.Action(
             name=action_name,
             parameters=params,
             receiver=self.tag,
         )])
-        action = res.results[0].action
-        error = res.results[0].error
+        action = res.actions[0].result
+        error = res.actions[0].error
         if error and error.code == 'not found':
             raise ValueError('Action `%s` not found on %s' % (action_name,
                                                               self.name))
         elif error:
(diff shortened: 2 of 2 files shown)


juanmanuel-tirado:
Rerun tests after merging #697

juanmanuel-tirado (approved):
LG2M

sed-i:
I'm now [getting](https://github.com/canonical/prometheus-k8s-operator/runs/7638474213?check_suite_focus=true#step:5:1440)
```
  File "/home/runner/work/prometheus-k8s-operator/prometheus-k8s-operator/.tox/integration/lib/python3.8/site-packages/juju/unit.py", line 167, in run_action
    res = await action_facade.EnqueueOperation(actions=[client.Action(
AttributeError: 'ActionFacade' object has no attribute 'EnqueueOperation'
```
Not sure if it's because of this change.

cderici:
It's for sure because of this change, but afaik the juju `2.9`'s `Action v6` facade should have an `EnqueueOperation` function. Is it possible for you to upgrade to `juju 2.9/stable` (that's `2.9.32` I think)?

sed-i:
Oh nice, I didn't know.
Most (all?) observability tests are pinned to 2.9.29 because of https://bugs.launchpad.net/juju/+bug/1977582, so only if 2.9.33 fixes it we would unpin.

BTW would it be easy for pylibjuju to raise a runtime error if the controller version is too old?

juanmanuel-tirado:
We have not decided so far the list of Juju controllers to be supported by every pylibjuju version. But that makes perfect sense.

Issue #699, opened 2022-08-03
Missing way to inspect relation databags


PietroPasotti:
Hi,
I've been looking for a way to access relation data contents (same as via `juju unit-info`) but I couldn't find one.
What is the recommended way to do it?

Issue #644, opened 2022-03-02
Need means of retrieving ingress addresses per unit


mmanciop:
When using libjuju for writing integration tests for sidecar charms, there seems to be no ergonomic way of retrieving the ingress address (i.e., the unit pod ip) through libjuju. Ideally, we would have something like the [`public_address`](https://github.com/juju/python-libjuju/blob/1f50f544c9753d867ac33cf05543cf9002f1cf03/juju/unit.py#L69) property (which, btw, seems to always return `None` with sidecar charms).

cderici:
@mmanciop Sorry for the delayed response, did you try to use the `unit.get_public_address()`? https://github.com/juju/python-libjuju/blob/e8456e118f1f2e494af9d885001063277e7cdfd1/juju/unit.py#L92

cderici:
Closing this as incomplete (non-reproducible), feel free to re-open if needed.

PR #700, opened 2022-08-03
added get_relation_data in Unit

PietroPasotti:
This PR adds to `Unit` a `get_relation_data(endpoint:str, related_endpoint:str, remote:str)` method that returns 'raw' RelationData objects.

Fixes https://github.com/juju/python-libjuju/issues/699
Fixes https://github.com/juju/python-libjuju/issues/644
(somewhat) Fixes https://github.com/charmed-kubernetes/pytest-operator/issues/81

It would be nice, as a next step, to add a shortcut to access this data via the Relation object.

Note: this is untested code. Will finalize after a green light that this is indeed a good addition

### DIFF
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -59,7 +59,7 @@ ubuntu charm, then exits:
       model = Model()
 
       # Connect to the currently active Juju model
-      await model.connect_current()
+      await model.connect()
 
       try:
           # Deploy a single unit of the ubuntu charm, using the latest revision
@@ -119,7 +119,7 @@ and then, to connect to the current model and fetch status:
 
   >>> from juju.model import Model
   >>> model = Model()
-  >>> await model.connect_current()
+  >>> await model.connect()
   >>> status = await model.get_status()
 
 
--- a/juju/unit.py
+++ b/juju/unit.py
@@ -104,6 +104,41 @@ class Unit(model.ModelEntity):
             raise JujuAPIError("expected one result")
         return defResult.results[0].result.get('public-address', None)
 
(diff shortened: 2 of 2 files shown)


cderici:
Thanks for the PR @PietroPasotti , I'll review this when you mark it as `Ready for review` (i.e. when it's not a draft) :+1: 

This will need a test as well to be approved. Also, I don't see how this fixes the #644 , are you adding more functionality? If you aren't, then feel free to remove that line (so that the issue won't close automatically when we land this), I have that one in my todo list anyways.

Thanks again!

sed-i:
This is a great addition @PietroPasotti.
I spotted this PR after I came across your [`get_relation_data`](https://github.com/canonical/traefik-k8s-operator/blob/6fddc92d5ddca4be24c8184dc0124257df93e4b8/tests/integration/conftest.py#L254) helper.

Would this work for peer data as well?

PietroPasotti:
> This is a great addition @PietroPasotti. I spotted this PR after I came across your [`get_relation_data`](https://github.com/canonical/traefik-k8s-operator/blob/6fddc92d5ddca4be24c8184dc0124257df93e4b8/tests/integration/conftest.py#L254) helper.
> 
> Would this work for peer data as well?
> 
> Currently I'm using
> 
> ```python
> rc, stdout, stderr = await ops_test.run("juju", "show-unit", "app/0", "--format=json")
> data = json.loads(stdout)
> peer_data = next(filter(lambda d: d['endpoint'] == 'replicas', data["app/0"]["relation-info"]))
> ```

Ugh, had forgotten this was even here. I think peer data requires a little addition. I have it working in jhack, that's where this code originally came from. Putting it again in my todo list

PR #703, opened 2022-08-08
[JUJU-1577] Replace k8s bundles with machine bundles for tests

cderici:
#### Description

Some of the tests in the CI are failing because we’re using k8s bundles while running the tests on lxd (don’t ask why they were passing before). This change replaces them with a machine bundle.

#### QA Steps

```
tox -e integration -- tests/integration/test_model.py::test_deploy_trusted_bundle
```

```
tox -e integration -- tests/integration/test_model.py::test_deploy_bundle
```


juanmanuel-tirado (approved):
LG2M

Issue #705, opened 2022-08-10
v.3.0.0: AttributeError: 'ActionFacade' object has no attribute 'EnqueueOperation'


sed-i:
With [pylibjuju 3.0.0](https://github.com/canonical/prometheus-k8s-operator/runs/7771894069?check_suite_focus=true#step:5:31) there is now a [facade error](https://github.com/canonical/prometheus-k8s-operator/runs/7771894069?check_suite_focus=true#step:5:1394):

```
   File "/home/runner/work/prometheus-k8s-operator/prometheus-k8s-operator/.tox/integration/lib/python3.8/site-packages/juju/unit.py", line 233, in run_action
    res = await action_facade.EnqueueOperation(actions=[client.Action(
AttributeError: 'ActionFacade' object has no attribute 'EnqueueOperation'
```

cderici:
Yeah I'm currently working on a fix on actions that'll solve this one as well, thanks for reporting :+1:

PR #706, opened 2022-08-10
[JUJU-1593] Fix `run_actions` and facade issues

cderici:
#### Description

This change attempts to fix the `Unit.run_actions` issue where the resulting object is different based on the used Facade (which is based on whether the `2.9` or `3.0` juju client is being used). With this change both ways will return the same type of object as a result of running the action on the unit. It's also a continuation/possible-regression-fix of #698.

Also adding the `EnvironUpgrader` facade that's introduced in `Juju 3.0`, so we shouldn't get the "unexpected facade" warning anymore.

Also updating the clients with the latest schema from juju 3.0. In particuılar the Application facade version is bumped up `13` -> `14`.

With this #705 should be fixed too. 

#### QA Steps

QA steps for this is sort of a simulation of one of the real world libjuju uses in a function on [mongodb-operator](https://github.com/canonical/mongodb-operator/blob/8796be02e10b84f990da0426ad1cff5922dc506a/tests/integration/helpers.py#L39).

Get a controller on a k8s cloud (I did it on `microk8s`). The following then should work fine.
Full QA should be done with 2 separate controllers (at least that's what I did): one bootstrapped with `juju 2.9` and another with `juju 3.0`.

```python
async def _get_password():
    model = Model()
    await model.connect()
    await model.deploy('zinc-k8s')
    await model.wait_for_idle(status="active")

    unit = model.applications['zinc-k8s'].units[0]
    action = await unit.run_action("get-admin-password")
    action = await action.wait()
    print(action.results["admin-password"])
    await model.disconnect()
```

#### Notes & Discussion

This is also related (and most likely a fix) to:
*  https://github.com/openstack-charmers/zaza/issues/545
*  https://github.com/openstack-charmers/zaza/pull/546
* https://github.com/canonical/mongodb-operator/runs/7761747331?check_suite_focus=true


cderici:
@juanmanuel-tirado I haven't got the change to run the full test suite locally yet (only ran the linter and the qa steps above), so keep in mind before approving/landing this, it's important that we don't see any critical failures in the CI about facades etc, especially after updating the schema and the clients, if it's full green, then it's safe to go ahead and land after QAing :+1:

juanmanuel-tirado:
Manual QA runs OK with 2.9.32 and 3.0-rc1 controllers.

juanmanuel-tirado (approved):
LG2M

Issue #707, opened 2022-08-11
`unit.run` output doesn't contain "results" field in v3.0.0, which breaks `run_on_unit` function in zaza


agileshaw:
With python-libjuju 3.0.0, `unit.run` returns less information than before, notably the absent of the `results` field. I think the commit causes this change is [a1c4f20](https://github.com/juju/python-libjuju/commit/a1c4f20611cc014c7ec78e88840da1fa9ba75fd2) (when `self.connection.is_using_old_client` is true).

With 3.0.0
```
>>> print(action.data)
{'model-uuid': '12f9d804-4369-4071-86ee-3dfaaae00ed4', 'id': '26', 'receiver': 'ubuntu/0', 'name': 'juju-run', 'status': 'completed', 'message': '', 'enqueued': '2022-08-10T23:10:02Z', 'started': '2022-08-10T23:10:06Z', 'completed': '2022-08-10T23:10:06Z'}
```

With 2.9.11
```
>>> action = await unit.run("df -h", timeout=None)
>>> print(action.data)
{'model-uuid': '12f9d804-4369-4071-86ee-3dfaaae00ed4', 'id': '28', 'receiver': 'ubuntu/0', 'name': 'juju-run', 'parameters': {'command': 'df -h', 'timeout': 0, 'workload-context': False}, 'status': 'completed', 'message': '', 'results': {'Code': '0', 'Stdout': 'Filesystem      Size  Used Avail Use% Mounted on\nudev            975M     0  975M   0% /dev\ntmpfs           199M  1.1M  198M   1% /run\n/dev/vda1        20G  2.0G   18G  11% /\ntmpfs           992M     0  992M   0% /dev/shm\ntmpfs           5.0M     0  5.0M   0% /run/lock\ntmpfs           992M     0  992M   0% /sys/fs/cgroup\n/dev/loop0       62M   62M     0 100% /snap/core20/1518\n/dev/loop1       68M   68M     0 100% /snap/lxd/22753\n/dev/loop2       47M   47M     0 100% /snap/snapd/16292\n/dev/vda15      105M  5.2M  100M   5% /boot/efi\n/dev/vdb         40G   24K   38G   1% /mnt\ntmpfs           1.0M     0  1.0M   0% /var/snap/lxd/common/ns\n'}, 'enqueued': '2022-08-10T23:46:05Z', 'started': '2022-08-10T23:46:06Z', 'completed': '2022-08-10T23:46:07Z'}
```

[...]

PR #710, opened 2022-08-11
[JUJU-1593] Fix `unit.run()` and update the old client codes

cderici:
#### Description

This is PR complements #706, and tries to address the issues being discussed in #707, by updating the `old_client`s with `2.9.34` schema and fixing the `unit.run()` function to be asynchronous, as it's (breakingly) changed to be in the `Juju 3.0`. This also brings `unit.run` and `unit.run_action` functions very close, the only difference is that the former is using `Action.Run` while the latter is using `Action.Enqueue` functions. It's plausible that one of these  ways of doing the similar thing will disappear in the future since the newer versions of `Action.Run` is asynchronous anyway.

Should fix the #707.

#### QA Steps

I updated the integration tests for `run()`, so basically all the tests in the following suite (that includes `test_run` and `test_run_action`) should pass:

```python
tox -e integration -- tests/integration/test_unit.py
``` 

Additionally I added the example that I have been using myself to test this into the examples folder, so it should be running fine:

```
 $ python examples/run_action.py
```

Both the integration tests (`test_run` & `test_run_action`) and the example module are tested in the following controllers:

```
Controller       Model       User   Access     Cloud/Region         Models  Nodes    HA  Version
caner-k8s        test        admin  superuser  microk8s/localhost        2      -     -  2.9.33     
caner-k8s-2929   testhede    admin  superuser  microk8s/localhost        2      -     -  2.9.29     
caner-k8s-2931   testhaylaz  admin  superuser  microk8s/localhost        2      -     -  2.9.31     
caner-k8s-2934*  haylazlar   admin  superuser  microk8s/localhost        2      -     -  2.9.34     
caner-lxd-2.9    default     admin  superuser  localhost/localhost       3      2  none  2.9.34.1   
caner-lxd-2929   default     admin  superuser  localhost/localhost       2      1  none  2.9.29     
caner-lxd-2931   default     admin  superuser  localhost/localhost       2      1  none  2.9.31     
caner-lxd-2932   default     admin  superuser  localhost/localhost       2      1  none  2.9.32     
caner-lxd-3.0    canhiras    admin  superuser  localhost/localhost       1      1  none  3.0-rc1.1  
k8s-30           test4       admin  superuser  microk8s/localhost        2      -     -  3.0-rc1    
```

Note that, all the clients that I used  `<2.9.34` are using the `ActionFacade v6`, and the ones `>=2.9.34` are using the `ActionFacade v7` (regardless of `2.9` vs` 3.0`).

#### Notes & Discussion

One of the requests was to keep the `unit.run()` function synchronous for any client using `<ActionFacade v6`. While this is a valid request from the user perspective, it also means that we would have the same function returning different types (either a dictionary or an action object) depending on the facade being used, which would make it way harder to maintain. That's why I kept the change for all the versions.


juanmanuel-tirado (approved):
LG2M 

Thanks for consolidating this.

PR #714, opened 2022-08-18
[JUJU-1664] Add force, no-wait, destroy-storage params to app.destroy

### LINKED ISSUE #645: Application.destroy() should have force, no_wait, destroy_storage options
I would expect the api to reflect more closely the juju cli.
In particular I would expect Application.destroy() to expose options equivalent to `--force --no-wait --destroy-storage`, or to make clear in the docstring why it doesn't and what the actual behaviour is equivalent to.

cderici:
#### Description

This PR adds the parameters `force`, `no_wait` and `destroy_storage` parameters to `Application.destroy()`.

Fixes #645

#### QA Steps

```
tox -e integration -- tests/integration/test_application.py::test_app_destroy
```

```
tox -e integration -- tests/integration/test_application.py::test_app_remove_wait_flag
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -227,8 +227,7 @@ class Application(model.ModelEntity):
         """
 
         if no_wait and not force:
-            log.warning("Invalid parameters. Automatically setting the --force parameter to True.")
-            force = True
+            raise JujuError("--no-wait without --force is not valid")
 
         app_facade = self._facade()
 
--- a/juju/application.py
+++ b/juju/application.py
@@ -217,16 +217,31 @@ class Application(model.ModelEntity):
         return await self.model.destroy_units(*unit_names)
     destroy_units = destroy_unit
 
-    async def destroy(self):
+    async def destroy(self, destroy_storage=False, force=False, no_wait=False):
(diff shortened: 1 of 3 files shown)


juju/application.py
```diff
@@ -217,16 +217,31 @@ async def destroy_unit(self, *unit_names):
...
+    async def destroy(self, destroy_storage=False, force=False, no_wait=False):
         """Remove this application from the model.
 
+        :param bool destroy_storage: Destroy storage attached to application unit. (=false)
+        :param bool force: Completely remove an application and all its dependencies. (=false)
+        :param bool no_wait: Rush through application removal without waiting for each individual step to complete (=false)
+        :param bool block: Blocks until the application is removed from the model
         """
+
+        if no_wait and not force:
```
juanmanuel-tirado, line 229: Following the behaviour from the CLI, `no-wait` and `force` should not be set by default.

juanmanuel-tirado (changes requested):
We must not set `force` and `no-wait` by default.

juanmanuel-tirado (approved):
LG2M

PR #717, opened 2022-08-23
action.wait will block until complete/failed -- update to the latest entity

addyess:
`action.wait()` returns itself with updated `self.results`, but the `self.data` is stale.  

this can result in situations where the results have valid data, but `action.success` may say `pending` or `running` when in reality it's complete or failed.

to address this issue, return `self.latest()` which should return a new Action object that's got a valid `completed` or `failed` action

### DIFF
--- a/juju/action.py
+++ b/juju/action.py
@@ -16,4 +16,4 @@ class Action(model.ModelEntity):
 
     async def wait(self):
         self.results or await self.fetch_output()
-        return self
+        return self.latest()


juanmanuel-tirado:
Please provide a description for the PR.

jameinel:
ok to test

jameinel:
add to whitelist

addyess:
@cderici , agree, this isn't a correct solution.  My tests proved that this PR isn't a fully baked conclusion because `action.latests()` didn't  update the model entity so the `data` property still had the old values.  

is interesting that `model._get_completed_action(...)` gets there and sees the entity as being completed, but the action object describing the entity is still out of data.  I'm wondering if the model state data isn't updated while it's polling the model to see if the entity is "completed or failed".  Your genius is better than my poor rags.  Thanks for looking into it

PR #720, opened 2022-08-23
[JUJU-1681] Add --attach-storage parameter to model.deploy

### LINKED ISSUE #716: Enable functionality for `juju deploy <charm> --attach-storage=<storage id>`
Enable an `attach-storage` option so that this is possible: `await ops_test.model.deploy(my_charm, num_units=1, attach-storage=storage_id)`

cderici:
#### Description

As the name suggests, this adds `--attach-storage` parameter to the `model.deploy()`.

Fixes #716 

#### QA Steps

The following test is added for this. However, see the Notes below.

```
tox -e integration -- tests/integration/test_model.py::test_model_attach_storage_at_deploy
```

All CI tests need to pass.

#### Notes & Discussion

Currently the `attach-storage` and `detach-storage` is not working consistently on the Juju side. So we're skipping the storage tests that need those functions. (Investigating on the Juju side) The reason we need them is that the `--attach-storage` parameter on the `deploy()` expects an already existing storage. And the only way that I know how to get one that's attachable is to `add_storage()` to a unit and `unit.detach_storage()` it, which makes the storage available and retain it in the model. For some reason, however, either the `detach` doesn't detach properly I'm not sure, the `remove_application` wipes the storage out of the model (in the juju cli as well).

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1670,7 +1670,7 @@ class Model:
             self, entity_url, application_name=None, bind=None,
             channel=None, config=None, constraints=None, force=False,
             num_units=1, overlays=[], plan=None, resources=None, series=None,
-            storage=None, to=None, devices=None, trust=False):
+            storage=None, to=None, devices=None, trust=False, attach_storage=[]):
         """Deploy a new service or bundle.
 
         :param str entity_url: Charm or bundle url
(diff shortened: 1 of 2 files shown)


MiaAltieri:
Awesome stuff here! Big thanks @cderici

juanmanuel-tirado (approved):
LG2M

PR #722, opened 2022-08-24
[JUJU-1663] Drop Python 3.5 support from python-libjuju

cderici:
#### Description

Python-libjuju no longer supports Python 3.5 (:tada:), following the Juju dropping support for xenial.

#### QA Steps

The PR on `juju-qa-jenkins` to remove the `Python3.5` tests is up https://github.com/juju/juju-qa-jenkins/pull/67, and I ran the job builder with the changes, so we shouldn't see any (unit, integration) tests spawn for `Python 3.5`.

Even if they spawn somehow, we can ignore them, but we need to make sure all the other tests are passing for other Python versions. (Modulo, of course we might have some known intermittent bugs that are still under investigation.)

### DIFF
--- a/tests/integration/test_connection.py
+++ b/tests/integration/test_connection.py
@@ -147,12 +147,6 @@ class RedirectServer:
         if hasattr(ssl, 'PROTOCOL_TLS_SERVER'):
             # python 3.6+
             protocol = ssl.PROTOCOL_TLS_SERVER
-        elif hasattr(ssl, 'PROTOCOL_TLS'):
-            # python 3.5.3+
-            protocol = ssl.PROTOCOL_TLS
-        else:
-            # python 3.5.2
-            protocol = ssl.PROTOCOL_TLSv1_2
         self.ssl_context = ssl.SSLContext(protocol)
         crt_file = Path(__file__).with_name('cert.pem')
         key_file = Path(__file__).with_name('key.pem')
--- a/tests/unit/test_bundle.py
+++ b/tests/unit/test_bundle.py
@@ -1122,10 +1122,6 @@ class TestBundleHandler:
 
         bundle = await handler._handle_local_charms(bundle, bundle_dir)
 
-        # TODO: for some reason 'assert_has_calls' is failing in
-        # Python3.5, refactor this with 'assert_has_calls' when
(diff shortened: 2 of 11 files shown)


cderici:
Two initial observations on the CI:
* Unit tests didn't spawn, probably because we're modifying the `.github/workflow/tox.yaml` (which is what is used to spawn the unit tests)
* We seem to be getting all kinds of weirdness from Jenkins for the integration tests (Not related to the PR). There was a recent maintenance on that, might be the cause of what's going on. We should probably wait to land this until we see the tests running and passing. (except the `github-integration-tests-pylibjuju-python3.5`)

tests/unit/test_bundle.py
```diff
@@ -1122,10 +1122,6 @@ async def __call__(self, *args, **kwargs):
 
         bundle = await handler._handle_local_charms(bundle, bundle_dir)
 
-        # TODO: for some reason 'assert_has_calls' is failing in
-        # Python3.5, refactor this with 'assert_has_calls' when
-        # Python3.5 support is dropped
-
```
jack-w-shaw, line 1128: But has this refactor been done?

tests/integration/test_connection.py
```diff
@@ -147,12 +147,6 @@ def __init__(self, destination):
         if hasattr(ssl, 'PROTOCOL_TLS_SERVER'):
```
jack-w-shaw, line 147: Will this `if` ever fail? Do we really want to do nothing if it is false? This seems like a potential way in for a silent failure

tests/unit/test_bundle.py
cderici: No, because I think keeping the explicit checks helps to make it clear what is being tested and how. I'm not a huge fan making things concise in testing, it only conceals stuff and introduces more magic imho (... turns around and runs `tox` :facepalm:).

Here's the original change that switched to explicit https://github.com/juju/python-libjuju/commit/2dc598adbfb93d23279952f17a23a26cf3478459 , I think keeping it like this is better, but not a big deal really to change it back, let me know what you think.

jack-w-shaw (approved):
LGTM

PR #724, opened 2022-08-31
[JUJU-1671] Charmhub url from model config

cderici:
#### Description

Before this change, the api host address for the charmhub was hard-coded. This change allows fetching the api address from the model config.

#### QA Steps

```
tox -e integration -- tests/integration/test_charmhub.py
```

### DIFF
--- a/juju/charmhub.py
+++ b/juju/charmhub.py
@@ -18,18 +18,22 @@ class CharmHub:
             jasyncio.sleep(5)
         raise JujuError("Got {} from {}".format(_response.status_code, url))
 
-    def get_charm_id(self, charm_name):
+    async def get_charm_id(self, charm_name):
         conn, headers, path_prefix = self.model.connection().https_connection()
 
-        url = "https://api.snapcraft.io/v2/charms/info/{}".format(charm_name)
+        model_conf = await self.model.get_config()
+        charmhub_url = model_conf['charmhub-url']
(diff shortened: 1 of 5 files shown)


juju/charmhub.py
```diff
@@ -18,18 +18,22 @@ def request_charmhub_with_retry(self, url, retries):
...
         return response['id'], response['name']
 
-    def is_subordinate(self, charm_name):
+    async def is_subordinate(self, charm_name):
         conn, headers, path_prefix = self.model.connection().https_connection()
 
-        url = "https://api.snapcraft.io/v2/charms/info/{}?fields=default-release.revision.subordinate".format(charm_name)
+        model_conf = await self.model.get_config()
+        charmhub_url = model_conf['charmhub-url']
+        url = "{}/v2/charms/info/{}?fields=default-release.revision.subordinate".format(charmhub_url.value, charm_name)
```
jack-w-shaw, line 36: Now that we've dropped Python 3.5 support we can use fstrings right?

jack-w-shaw (approved):
LGTM. One small q

juju/charmhub.py
cderici: Yeah I agree :+1:  we should probably turn all the formattings into fstrings at some point

cderici:
I think there's something going on in the tests, none of them is finished, it's more than likely about this change, it's blocking for some reason. Will investigate :+1:

PR #726, opened 2022-09-07
add support for ipv6

jdkandersson:
#### Description

For machine charms, every now and then during testing libjuju will pick an IPv6 address to communicate with the machine. This results in problems with the part of the code changed because it just splits the endpoint to a host and port on the first `:` character which, for an IPv6 address, does not separate the host and port since an IPv6 address has many `:` characters in it. After applying this patch locally, libjuju is able to communicate with the machine again.

*\<Fixes: add support for ipv6 connections\>*

#### QA Steps

*\<Commands / tests / steps to run to verify that the change works: I ran tests for a charm that ocassionaly used ipv6 with these changes and they worked\>*

```
tox -e py3 -- tests/unit/...
```

```
```

All CI tests need to pass.

#### Notes & Discussion

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -678,7 +678,9 @@ class Connection:
 
         """
         endpoint = self.endpoint
-        host, remainder = endpoint.split(':', 1)
+        # Support IPv6 by right splitting on : and removing [] around IP address for host
+        host, remainder = endpoint.rsplit(':', 1)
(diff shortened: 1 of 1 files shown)


cderici (approved):
CI failures on this seem to be intermittent glitches with the charmhub api, unrelated to this change. We can go ahead and land this. The merger might need a rebase on master to complete successfully.

PR #727, opened 2022-09-07
Add wait_for_app flag to wait_for_idle

nsklikas:
#### Description

The wait_for_idle function wasn't waiting for the application status to become active. This meant that after it returned the units would be in an `active` state while the application could be still in `waiting`. 

I added a flag to make this inactive by default, in case it breaks something. Not sure if that's needed.

#### QA steps

To run the unit tests:
```
tox -e unit -- -k test_wait_for_idle_wait_for_app_status
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2530,7 +2530,7 @@ class Model:
 
     async def wait_for_idle(self, apps=None, raise_on_error=True, raise_on_blocked=False,
                             wait_for_active=False, timeout=10 * 60, idle_period=15, check_freq=0.5,
-                            status=None, wait_for_units=1, wait_for_exact_units=None):
+                            status=None, wait_for_units=1, wait_for_exact_units=None, wait_for_app_status=False):
         """Wait for applications in the model to settle into an idle state.
 
         :param apps (list[str]): Optional list of specific app names to wait on.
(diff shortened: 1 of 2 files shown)


cderici:
Hi @nsklikas, thanks for the change! :wave:  I see that you added some unit tests, thanks for that (please also add them in your PR description as QA steps). But we're gonna need some integration tests for this one. In particular, I'd like to see some integration tests that demonstrate the scenario you're talking about (and fail without your changes) and pass when your changes are applied. Thanks again!

nsklikas:
I encountered this bug when I was developing the https://charmhub.io/gatekeeper-controller-manager charm. In that charm the application becomes active quite some time after the units do. I am not very familiar with how juju decides on the application status and why that is happening, my guess is that it has something to do with the pebble readiness checks. 

I will try to create a simple charm that proves the issue and add it to the integration tests. Is adding a dummy charm in the integration tests ok with you?

cderici:
> I will try to create a simple charm that proves the issue and add it to the integration tests. Is adding a dummy charm in the integration tests ok with you?

Absolutely. Any test that demonstrates the problem and that the change fixes it will do just fine. We have a couple of integration tests that are using dummy charms, might serve well as examples. Thanks for working on this! :+1:

Issue #728, opened 2022-09-13
[Bug]: application.refresh can't switch channels on a ch charm


addyess:
### Description

It's come to my attention [this method](https://github.com/juju/python-libjuju/blob/master/juju/application.py#L617) it quite broken for upgrading an application in a juju model. 

it does seem to rely heavily on charmstore apis 

i think the issues i'm running into can be reproduced by installing `ubuntu` `--channel=stable` and try to just update it `candidate`

But just reading through this code it seems to believe the current running charm is `ch:amd64/jammy/ubuntu`  via `self.data['charm-url']` -- and it's trying to look up this charm in charmstore but it's not gonna find anything at [this link](https://api.jujucharms.com/charmstore/v5/ch:amd64/jammy/ubuntu/meta/any\?include\=bundle-machine-count\&include\=bundle-metadata\&include\=bundle-unit-count\&include\=charm-actions\&include\=charm-config\&include\=charm-metadata\&include\=common-info\&include\=extra-info\&include\=owner\&include\=published\&include\=resources\&include\=supported-series\&include\=terms\&include\=id\&include\=stats\&channel\=candidate)

### Urgency

Blocker for our release

### Python-libjuju version

2.9.11 (seems to be the same in current 3.x versions)

### Juju version

2.9.33

### Reproduce / Test

```python
`juju deploy ubuntu --channel=stable`

try to use 
`Application.refresh(channel="candidate")` to upgrade the charm.

I'll need a solution which can also refresh all the resources attached to a charm.
```

PR #729, opened 2022-09-13
[JUJU-1800] Revise the `application.upgrade_charm()` (refresh)

cderici:
#### Description

The `application.refresh()` method was crashing due to the resources checks because of the old CharmStore code. This adds a new method, namely `list_resources` into the charmhub module and replaces the old code with the new CharmHub checks using the new method.

Flixes: #728 

#### QA Steps

Two tests were added, one for the `list_resources` method, and one for the `application.upgrade_charm`. Both of them need to pass, along with the other tests (modulo, we might still have some of those known intermittent CI failures).

```
tox -e integration -- tests/integration/test_charmhub.py::test_list_resources
```

```
tox -e integration -- tests/integration/test_application.py::test_upgrade_charm_switch_channel
```

The `test_upgrade_charm_switch_channel` also includes the reproduce/test section from #728, so no need for a manual check.

All CI tests need to pass.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -647,6 +647,10 @@ class Application(model.ModelEntity):
         resources_facade = client.ResourcesFacade.from_connection(self.connection)
         charms_facade = client.CharmsFacade.from_connection(self.connection)
 
+        # 1 - Figure out the destination origin and destination charm_url
+        # 2 - Then take care of the resources
+        # 3 - Finally execute the upgrade
+
         # Get the charm URL and charm origin of the given application is running at present.
         charm_url_origin_result = await app_facade.GetCharmURLOrigin(application=self.name)
         if charm_url_origin_result.error is not None:
@@ -664,6 +668,8 @@ class Application(model.ModelEntity):
         if revision is not None:
             origin.revision = revision
 
+        # Make the source-specific changes to the origin/channel/url
+        # (and also get the resources necessary to deploy the (destination) charm -- for later)
         if Schema.CHARM_HUB.matches(parsed_url.schema):
             origin.source = 'charm-hub'
             if channel:
@@ -675,21 +681,22 @@ class Application(model.ModelEntity):
             charm_resources = await charmhub.list_resources(charm_name)
         else:
             charmstore = self.model.charmstore
-            charmstore_entity = await charmstore.entity(charm_url, channel=channel)
-
+            charmstore_entity = None
             if switch is None:
                 charm_url = charm_url.rpartition('-')[0]
                 if revision is not None:
                     charm_url = "%s-%d" % (charm_url, revision)
                 else:
+                    charmstore_entity = await charmstore.entity(charm_url, channel=channel)
                     charm_url = charmstore_entity['Id']
             origin.source = 'charm-store'
             if channel:
                 origin.risk = channel
-
+            if charmstore_entity is None:
+                charmstore_entity = await charmstore.entity(charm_url, channel=channel)
             charm_resources = charmstore_entity['Meta']['resources']
 
-        # resolve the given charm URLs with an optionally specified preferred channel.
+        # Resolve the given charm URLs with an optionally specified preferred channel.
         # Channel provided via CharmOrigin.
         resolved_charm_with_channel_results = await charms_facade.ResolveCharms(
             resolve=[client.ResolveCharmWithChannel(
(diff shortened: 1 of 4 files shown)


tests/integration/test_application.py
```diff
@@ -167,6 +171,48 @@ async def test_upgrade_charm_channel(event_loop):
         assert app.data['charm-url'] != 'cs:ubuntu-0'
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_upgrade_charm_switch_channel(event_loop):
+    # Note for future:
+    # This test requires a charm that has different
+    # revisions in channels 'stable' and 'candidate'.
+    # Currently, we use mongodb, but eventually
```
hmlanigan, line 180: You can use the juju-qa-test charm for this. There are 2 tracks currently in use for it. Using that charm would also allow to disambiguation of the channels "stable", "candidate" etc. 

Suggest also adding a track to the risk for the channel. Risk only channels are more typical of charmstore charms.

juju/charmhub.py
```diff
@@ -41,6 +41,16 @@ async def is_subordinate(self, charm_name):
     # TODO (caner) : we should be able to recreate the channel-map through the
```
hmlanigan, line 41: Does this TODO apply to list_resources or info?

cderici: All of it, really, since if we create a channel-map dict then all of these methods are gonna change, maybe moving this to the top would be more appropriate

juju/application.py
```diff
@@ -641,48 +643,85 @@ async def refresh(
...
-            charm_url = charm_url.rpartition('-')[0]
-            if revision is not None:
-                charm_url = "%s-%d" % (charm_url, revision)
+            charms_facade = client.CharmsFacade.from_connection(self.connection)
+            charmstore = self.model.charmstore
+            charmstore_entity = None
+            if switch is not None:
+                charm_url = switch
+                if not charm_url.startswith('cs:'):
+                    charm_url = 'cs:' + charm_url
```
hmlanigan, line 697: What happens with local charms? I see a method below which is localRefresh?

I'd say it's an error if the url has no schema for refresh unless switch is used. With the cli, a url without a schema defaults to charmhub.

hmlanigan: Is there a test with refresh doing a switch from cs to ch? Or refresh with switch to local?

juju/application.py
```diff
@@ -641,48 +643,85 @@ async def refresh(
...
 
-        # Update resources
-        if not charmstore_entity:
-            charmstore_entity = await charmstore.entity(charm_url,
-                                                        channel=channel)
-        store_resources = charmstore_entity['Meta']['resources']
+            if charm_url == self.data['charm-url']:
+                raise JujuError('already running charm "%s"' % charm_url)
 
+            dest_origin = client.CharmOrigin(source="charm-store", risk=channel)
```
hmlanigan, line 710: I'm curious what the charm-origin in the db for this application looks like after AddCharm/SetCharm are called below. Thinking of the bug I just put a pr up to fix: LP1990182

juju/application.py
```diff
@@ -641,48 +643,85 @@ async def refresh(
...
+        # First we need to make sure we have the resources for the charm that's
+        # coming
+
+        # Get the list of resources needed to deploy this charm
+        if Schema.CHARM_HUB.matches(parsed_url.schema):
+            # Charmhub charms
+            charmhub = self.model.charmhub
+            charm_resources = await charmhub.list_resources(charm_name)
+
+            charm_url_origin_result = await app_facade.GetCharmURLOrigin(application=charm_name)
```
hmlanigan, line 663: GetCharmURLOrigin should be used for all charms, not just charmhub charms. 

An application name does not have to be the same as a charm name. Does pylibjuju not allow for an application name different from the charm name?

juju/application.py
cderici: We do have a (passing) test for local upgrade https://github.com/juju/python-libjuju/blob/bd01785ab0ee3ba2374ecc60d93bd9bc1e6c070e/tests/integration/test_application.py#L196, but I agree I should probably revise the logic for switch and add some tests for it :+1:

tests/integration/test_application.py
```diff
@@ -187,7 +235,9 @@ async def test_upgrade_charm_switch(event_loop):
         await model.block_until(lambda: (len(app.units) > 0 and
                                          app.units[0].machine))
         assert app.data['charm-url'] == 'cs:ubuntu-0'
```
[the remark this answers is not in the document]
cderici: Yeah good call, I'll update and run the tests with both `2.9` and `3.0` and make sure both are happy :+1:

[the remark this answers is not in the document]
cderici: Well looks like the pylibjuju is able to use juju `3.0` to deploy that charm successfully, and all the other tests seem to pass as well, so I think we're good for now :+1:

hmlanigan (approved):

PR #737, opened 2022-09-30
[JUJU-1893] Revisit `charmhub.info()`

### LINKED ISSUE #734: [Feature]: Enable access to `juju info`
### Description

It would be useful sometimes in scripts or charms to be able to do `juju info somecharm` to get information on a charms channels, etc.  This would be great when interacting with charm channels, such as for release and promotion CI.

### Urgency

It would save from a weird workaround

### Code I'd Like to Run

```python
from ___ import info

charm_info = info("someCharmName")
channels = charm_info.channels # or charm_info['channels']
...
```
```

### PR DESCRIPTION
**Pull request by cderici**
#### Description

This PR revisits the `info` method in `charmhub.py`, allowing users to get information for a charm (like `juju info juju-qa-test`) in pylibjuju using either old or new clients. For old clients it calls `Info` from the `Charmhub` facade, and for the new clients it directly calls the charmhub api and processes the result. The channel-map representations are different between the `2.9` and `3.0` clients, this change introduces extra effort to get them as close as possible.

Fixes #734 

#### QA Steps

The existing charmhub tests are updated, and the ones that we used to skip (due to charmhub facade not existing anymore) are re-activated since we have the working code now (only the `info` ones). QA should be done for both `2.9` and `3.0` clients. So get yourself `juju 2.9` and `juju 3.0` and bootstrap two controllers and run the relevant tests:

```
tox -e integration -- tests/integration/test_charmhub.py::test_info
```
```
tox -e integration -- tests/integration/test_charmhub.py::test_info_with_channel
```
```
tox -e integration -- tests/integration/test_charmhub.py::test_info_not_found
```

You may also play with the `examples/charmhub_info.py` with different controllers if you wanna be extra pedantic.

### DIFF
--- a/juju/charmhub.py
+++ b/juju/charmhub.py
@@ -76,14 +76,39 @@ class CharmHub:
             result = res.result
             result.channel_map = self._channel_map_to_dict(result.channel_map)
             result = result.serialize()
-            return result
         else:
-            result = {}
+            charmhub_url = await self._charmhub_url()
+            url = "{}/v2/charms/info/{}?fields=channel-map".format(
+                charmhub_url.value, name)
+            _response = self.request_charmhub_with_retry(url, 5)
+            result = json.loads(_response.text)
         return result
+
+    def _channel_list_to_map(self, channel_list_map):
+        """Charmhub API returns the channel map as a list of channel objects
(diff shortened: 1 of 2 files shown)


jack-w-shaw (approved):
LGTM

PR #741, opened 2022-10-12
Wait for idle arg type check

### LINKED ISSUE #732: `wait_for_idle` should raise if `apps` is not a list
It is a common error to pass an str instead of List[str] to `apps=` in `wait_for_idle`, and as a result the test fails for timeout:

```
INFO     juju.model:model.py:2645 Waiting for model:
  p (missing)
  r (missing)
  o (missing)
  m (missing)
  e (missing)
  t (missing)
  h (missing)
  e (missing)
  u (missing)
  s (missing)
```

It would be handy if the test fails with a more obvious message.

https://github.com/juju/python-libjuju/blob/bd01785ab0ee3ba2374ecc60d93bd9bc1e6c070e/juju/model.py#L2536

cderici:
#### Description

Currently there's no check for the `apps` argument for the `wait_for_idle()` method. So it goes through any iterable that's passed to it, including strings. As a result `wait_for_idle(apps="test")` produces something like the following:

```
INFO     juju.model:model.py:2645 Waiting for model:
  t (missing)
  e (missing)
  s (missing)
  t (missing)
```

This introduces a check for the argument to be a `List[str]`.

Fixes #732 

#### QA Steps

Unit tests are added, should be a simple QA.

```
tox -e py3 -- tests/unit/test_model.py::TestModelWaitForIdle
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2581,6 +2581,12 @@ class Model:
         timeout = timedelta(seconds=timeout) if timeout is not None else None
         idle_period = timedelta(seconds=idle_period)
         start_time = datetime.now()
+        # Type check against the common error of passing a str for apps
+        if apps is not None and (not isinstance(apps, list) or
+                                 any(not isinstance(o, str)
+                                     for o in apps)):
+            raise JujuError(f'Expected a List[str] for apps, given {apps}')
+
         apps = apps or self.applications
         idle_times = {}
         last_log_time = None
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado (approved):
Apart from the problems regarding the testing phase, this LG2M

cderici:
> Apart from the problems regarding the testing phase, this LG2M

This CI issue is slightly more than trivial so it'll be fixed in a new seperate PR, already up at #742 .

PR #742, opened 2022-10-13
[JUJU-1970] Revise local refresh

cderici:
#### Description

This change revises the `local_refresh()` (i.e. `application.upgrade_charm()` with a local charm) to use the `app_facade.GetCharmURLOrigin()` to get the origin of the currently running application charm, instead of trying to pass `channel` to the `SetCharm` like before.

We already implemented this when revising `upgrade_charm` in #729, but the upgrade local was missed (we talked about it but I forgot to include it).

This should fix the `tests/integration/test_application.py::test_upgrade_local_charm` that's currently failing in the CI runs.

#### QA Steps

The integration test mentioned above is revised a little bit, so it should pass. QA with different types and shapes of local charms would be appreciated.

```
tox -e integration -- tests/integration/test_application.py::test_upgrade_local_charm
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -633,10 +633,6 @@ class Application(model.ModelEntity):
         :param str switch: Crossgrade charm url
 
         """
-        if path is not None:
-            await self.local_refresh(channel, force, force_series, force_units,
-                                     path, resources)
-            return
         if resources is not None:
             raise NotImplementedError("resources option is not implemented")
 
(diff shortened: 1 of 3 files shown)


cderici:
Didn't realize this doesn't have a review yet :)

juanmanuel-tirado (approved):
LG2M

I did not reviewed earlier because another reviewer was already assigned.

PR #752, opened 2022-10-21
[JUJU-2018] Update 2.9.36 facades & clients

cderici:
#### Description

This updates the `old_client`s with the juju `2.9.36` schema.

#### QA Steps

Bootstrap with juju --channel `2.9/edge`

```bash
 $ python -m asyncio
 > from juju import model  <----------- this line ensures that all the _clients (old and new are loaded successfully)
 > m = model.Model()
 > m.connect()
 > m.deploy('ubuntu') <---------------------- these 3 lines ensure that majority of the facades are working correctly
```

PR #754, opened 2022-10-21
[JUJU-2019] Small fixes for 3.0

cderici:
#### Description

This includes a bunch of small fixes/patches to make sure that everything's working fine with the `Juju 3.0` (along with the `2.9` track -- e.g. coming `2.9.36`), tested currently against `3.0/candidate`.

#### QA Steps

Bootstrap with juju `3.0/edge` and run:

```
tox -e integration
```

which goes through all the integration tests. The `tests/integration/test_model.py::test_storage_pools` might fail if you're not using lxd, but other than that everything should be all green.

#### Notes & Discussion

* So this PR makes sure that we're good with `juju 3.0` (I manually ran the integration tests on aws)
* #752 makes sure we're good with `juju 2.9.36`
* And the CI tests are currently running against `latest/stable` (juju `2.9.35`), so we should be good there too.

So the `pylibjuju 3.0.3` should be ready to go after this lands.


juanmanuel-tirado (approved):
LG2M

PR #774, opened 2022-11-18
[JUJU-2175] Remove juju 2.9 support on 3.1.0

juanmanuel-tirado:
#### Description

Python-libjuju will no longer support juju 2.9.x This PR removes the existing support and attempts some initial cleaning.

* `old_clients` were removed
* Clients were regenerated to support juju 3.1.0-beta
* Other juju schemas were removed
* Version was set to 3.1.0-beta1
* Added Github workflow to run tests on GithubActions
* Added singleton class to retrieve `asyncio` event loops. This may reduce problems regarding multiple event loops used as the same time when using `jasyncio.run`
* Integration tests were revisited.
  * Old charms were removed
  * Series set to focal or jammy were possible
  * Some bundles were changed 
  * Charmstore tests were removed
* `base.py` now forces the use of one single event loops instance.

#### QA Steps
Running QA against a 3.0 controller MUST work
Running QA against a 2.9 controller MUST fail

#### Notes & Discussion
* A large number of tests are checking juju behaviour rather than the correctness of the interaction between python-libjuju and juju.
* Obviously running integration tests while possible are particularly time consuming. We should consider when to run these tests.
* At this point any connection with a 2.9.x controller will simply continue up to a point where facades are not correctly recognised. We should add a check before connections are established.


jnsgruk:
That is an appealing looking diff statistic 😅

juanmanuel-tirado:
The largest negative contribution ever :)

jameinel:
Wow, I've never seen a 7 digit negative before, very impressive.

juanmanuel-tirado:
GithubActions may take about 50 minutes... running a local QA can be more convenient at this point.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,80 @@
+name: Testing
+
+on: [push, pull_request]
+
+jobs:
```
cderici, line 5:
```suggestion
jobs:
  lint:
    name: Lint
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python:
          - "3.8"
          - "3.9"
          - "3.10"
    steps:
    - name: Check out code
      uses: actions/checkout@v2
    - name: Setup Python
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python }}
    - name: Install dependencies
      run: |
        pip install tox
    - name: Run lint
      run: tox -e lint
```

cderici (changes requested):
This looks mostly good, and seems to be working as intended. Here are some thoughts:

- We should not remove the `schemas-juju-latest.json` link and make sure it points to the current latest which is `schemas-juju-3.0.3.json`, otherwise when we have a new schema, the `facade.py` [will pick the fir](https://github.com/juju/python-libjuju/blob/fd89d1a0bb4f162969b3c479cb4701c501cc6786/juju/client/facade.py#L940)st schema in the directory that matches with `re.compile(r'[0-9]+\.[0-9-]+[\.\-][0-9a-z]+(\.[0-9]+)?')` 

- We should give a good error message if 2.9 client is being tried with this, instead of the weird `AttributeError: 'CharmOrigin' object has no attribute 'os'` error which is a random error that looks like a bug from the user's perspective.

- Some lint fixes needed:

```
juju/jasyncio.py:70:1: E302 expected 2 blank lines, found 1
juju/jasyncio.py:75:5: E301 expected 1 blank line, found 0
juju/jasyncio.py:77:11: E111 indentation is not a multiple of 4
juju/jasyncio.py:78:11: E111 indentation is not a multiple of 4
juju/jasyncio.py:82:1: E302 expected 2 blank lines, found 1
juju/model.py:2299:1: W293 blank line contains whitespace
tests/base.py:15:1: E302 expected 2 blank lines, found 1
tests/integration/test_model.py:55:42: E226 missing whitespace around arithmetic operator
tests/integration/test_model.py:56:19: E231 missing whitespace after ','
tests/integration/test_model.py:258:1: W293 blank line contains whitespace
tests/integration/test_unit.py:114:5: E306 expected 1 blank line before a nested definition, found 0
```


- We should take out the `github-integration-tests-pylibjuju-2.9-edge`, and for the remaining integration tests we should track the juju `3.0/stable` instead of the `latest/stable` until the `latest/stable` is no longer `2.9` (i.e. all the integration tests will always fail until then), though I just saw that maybe the #775 is already fixing this one

juju/client/connector.py
```diff
@@ -73,6 +78,10 @@ async def connect(self, **kwargs):
                 await self._connection.close()
             self._connection = await Connection.connect(**kwargs)
 
+        # Check if we support the target controller
+        if not self._connection.info['server-version'].startswith(SUPPORTED_JUJU_API_PREFIX):
```
cderici, line 83: This will also break for any version that's >=3.1, so I'd do `SUPPORTED_JUJU_API_PREFIX = "3."`

cderici (approved):

juju/client/connector.py
juanmanuel-tirado: We will update it once we are sure we can support 3.1

PR #775, opened 2022-11-24
[JUJU-2256] Add integration tests to github actions pinning juju 2.9

juanmanuel-tirado:
#### Description

Run integration tests with GitHub actions.

#### QA Steps

Wait for github actions to pass.

#### Notes & Discussion


tests/integration/test_application.py
```diff
@@ -213,21 +201,10 @@ async def test_upgrade_charm_switch_channel(event_loop):
...
-                                         app.units[0].machine))
-        assert app.data['charm-url'] == 'cs:ubuntu-0'
-        await app.upgrade_charm(revision=8)
-        assert app.data['charm-url'] == 'cs:ubuntu-8'
-
-
 @base.bootstrapped
 @pytest.mark.asyncio
 async def test_upgrade_charm_switch(event_loop):
+    pytest.skip('This upgrading is not the recommended approach')
```
cderici, line 207: can we just remove this test?

tests/integration/test_application.py
```diff
@@ -147,26 +137,24 @@ async def test_deploy_charmhub_charm(event_loop):
 @pytest.mark.asyncio
 async def test_upgrade_charm(event_loop):
     async with base.CleanModel() as model:
-        app = await model.deploy('cs:ubuntu-0')
+        app = await model.deploy('ubuntu')
         await model.block_until(lambda: (len(app.units) > 0 and
                                          app.units[0].machine))
-        assert app.data['charm-url'] == 'cs:ubuntu-0'
+        assert app.data['charm-url'] == 'ubuntu'
```
cderici, line 143: this should probably be something like `assert 'ubuntu' in app.data['charm-url']` as the url will be something like `ch:amd64/focal/ubuntu-21`, but more importantly, after removing the `cs:` stuff this test is a tautology unless we try to deploy an earlier version of the charm. It currently tests only that the `app.upgrade_charm()` call doesn't error.

We might be able to use the `juju-qa-test` charm for this, e.g. deploy from the `2.0/stable` channel and upgrade to `2.0/edge`, that'll bring it from rev 22 to rev 23 (`juju info juju-qa-test`).

tests/integration/test_application.py
```diff
@@ -147,26 +137,24 @@ async def test_deploy_charmhub_charm(event_loop):
...
 @pytest.mark.asyncio
 async def test_upgrade_charm_channel(event_loop):
     async with base.CleanModel() as model:
-        app = await model.deploy('cs:ubuntu-0')
+        app = await model.deploy('ubuntu')
         await model.block_until(lambda: (len(app.units) > 0 and
                                          app.units[0].machine))
-        assert app.data['charm-url'] == 'cs:ubuntu-0'
+        assert app.data['charm-url'] == 'ubuntu'
         await app.upgrade_charm(channel='stable')
```
cderici, line 156: same in this one, the deploy will default to the `stable` risk, so this test wouldn't be upgrading anything.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,79 @@
+name: Testing
+
+on: [push, pull_request]
+
+jobs:
+  unit-tests:
```
cderici, line 6: https://github.com/juju/python-libjuju/pull/774#discussion_r1034096482 might be added here in this PR instead too

cderici (changes requested):

tests/integration/test_application.py
juanmanuel-tirado: Done

tests/integration/test_application.py
juanmanuel-tirado: I will simply leave `test_upgrade_charm_switch_channel` which has all the logic for this kind of testing.

tests/integration/test_application.py
juanmanuel-tirado: I have removed this in favour of `test_upgrade_charm_switch_channel`.

PR #782, opened 2023-01-05
[JUJU-2391] Fix wrong bases analysis.

juanmanuel-tirado:
#### Description

This should fix a base recognition for local charms as described [here](https://bugs.launchpad.net/juju/+bug/1992833).

#### QA Steps
Using a 3.0 controller run:
```
juju download parca-k8s --channel=edge
python3 -m asyncio
>>> model = Model()
>>> await model.connect_current()
>>> await model.deploy("./parca-k8s_9258025.charm")
```

#### Notes & Discussion

This has to be ported to the 2.9 branch.

### DIFF
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -11,8 +11,8 @@ Documentation: https://pythonlibjuju.readthedocs.io/en/latest/
 Requirements
 ------------
 
-* Python 3.6+
-* Juju 2.0+
+* Python 3.8/3.9/3.10
+* Tested using Juju 3.0.2
 
 
 Design Notes
--- a/juju/utils.py
+++ b/juju/utils.py
@@ -385,9 +385,9 @@ def get_local_charm_base(series, channel_from_arg, charm_metadata,
         else:
             # Also check the charmcraft.yaml
             charmcraft_yaml = get_local_charm_charmcraft_yaml(charm_path)
-        if 'bases' in charmcraft_yaml:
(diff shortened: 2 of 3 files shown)


cderici (approved):
The change LGTM. I'm a little surprised that I didn't include a test that'd catch this when I initially fixed it. Maybe it would be a good idea to add it.

About the unit tests, there are couple of things going on in [the failing unit test run](https://github.com/juju/python-libjuju/actions/runs/3847757279/jobs/6555761674):
* we no longer are using tox 3.x, the latest tox 4.x looks like has a bunch of changes
* I have no idea why we see some integration tests are failing here (e.g. `ERROR tests/integration/test_application.py`), they shouldn't be running in the unit tests, I'm a little confused about that tbh, we may need to take a closer look at the github action.
* some cross-dependencies are broken. Let's say we depend on module A and B. If A depends on B itself, then tox will fetch B as a dependency for A and cache it. If A no longer depends on B, then we still need B, but aren't getting it. So far I detected that the following are in that position (we need them, but they're not explicitly stated in tox)
```
macaroonbakery
theblues
toposort
typing-inspect
paramiko
```
so I added https://github.com/juju/python-libjuju/pull/782/commits/b93951e445a4f6ce1c6d14c9dca8546e8dce7e31 and the unit tests seem to be working again :+1:

cderici:
The integration tests are failing here with `juju server-version 2.9.37 not supported` because our tests install juju with `snap install juju --channel=stable` and the juju `latest/stable` is still `2.9.37` (`snap info juju`)

PR #788, opened 2023-01-17
[JUJU-2381] Facade updated to 2.9.38

juanmanuel-tirado:
#### Description

This PR prepares the 2.9 branch to run against Juju 2.9.38.

#### QA Steps

All CI tests need to pass.

#### Notes & Discussion

A 2.9.38 version should be released after this PR lands.


VERSION
```diff
@@ -1 +1 @@
-2.9.11
+2.9.12
```
cderici, line 1: This should only be done in a release PR, as we don't have a changelog for `2.9.12` yet (see [docs/changelog.rst](https://github.com/juju/python-libjuju/blob/2.9/docs/changelog.rst)).

cderici (approved):
In the `2.9` branch we still use the `schema-juju-latest.json` for finding the "latest" schema (to be able to run `make client`). It should be a link, but looks like we accidentally copied the `2.9.37`'s schema over it, so the following should done on this PR:

```
$ cd juju/client
$ rm schemas-juju-latest.json
$ ln -s schemas-juju-2.9.38.json schemas-juju-latest.json
```

cderici:
Also why do we still have tests for `python3.6` etc? I thought we already dropped them

juanmanuel-tirado:
The python 3.6 tests are launched from Jenkins. We have to disable them. Anyway, they were removed from the github actions.

cderici (approved):

PR #791, opened 2023-01-18
[JUJU-2426] Secrets support

juanmanuel-tirado:
#### Description

This PR initially adds support for secrets. Added functionality:
* model `list-secrets`
* controller `add-secret-backends`, `list-secret-backends`, `remove-secret-backends`, `update-secret-backends`

Additionally, this PR includes fixes for the utilization of the `Application` facade and the `remove-unit` function.

#### QA Steps

The CI has to pass as expected.

#### Notes & Discussion

In this PR we include an additional packaged charm (`charm-secret_ubuntu-22.04-amd64.charm`) to test the `list-secrets` functionality. Secrets have to generated by charms and unfortunately, we cannot deploy local charm folders and make them work as expected. See #797 for more details.


VERSION
```diff
@@ -1 +1 @@
-3.1.0-b1
+3.1.0b1
```
cderici, line 1: Is there any significance in this formatting that I don't know yet, or just an arbitrary choice?

juju/client/connection.py
```diff
@@ -26,7 +27,7 @@
     'AllWatcher': {'versions': [1, 2, 3]},
     'Annotations': {'versions': [2]},
     'Application': {'versions': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,
-                                 14, 15]},
+                                 14, 15, 16]},
```
cderici, line 30: This should be Application Facade `v17`, right?

juju/model.py
```diff
@@ -1982,19 +1982,29 @@ async def _deploy(self, charm_url, application, series, config,
...
+        log.debug(
+            'Destroying unit %s', unit_id)
+
+        return await app_facade.DestroyUnit(units=[{
+            'unit-tag': unit_tag,
+            'destroy-storage': destroy_storage,
+            'force': force,
+            'max-wait': max_wait,
+            'dry-run': dry_run,
+        }])
```
cderici, line 2007: The tests that are using this code in `test_model.py` (through `application.destroy_unit`), namely the `test_wait_for_idle_with_exact_units_scale_down` and the `test_wait_for_idle_with_exact_units_scale_down_zero` need to be updated. We probably should have at least 1 test just for the `model.destroy_unit` itself.

juju/utils.py
```diff
@@ -403,3 +403,27 @@ def base_channel_to_series(channel):
...
+    ------
+    unit_tag with the corresponding equivalence or None
+    if the conversion was not possible
+    """
+
+    pos = unit_id.find("/")
+    if pos == 1:
+        return None
+    unit_id.replace("/", "-")
+    return "unit-" + unit_id
```
cderici, line 429: This is redundant, see `tag.unit()` in `juju/tag.py`.

juju/controller.py
```diff
@@ -871,6 +871,96 @@ async def _watcher(stop_event):
...
+           a list of errors if any
+        """
+        facade = client.SecretBackendsFacade.from_connection(self.connection())
+        return await facade.AddSecretBackends([{
+            'id': id,
+            'backend-type': backend_type,
+            'config': config,
+            'name': name,
+            'token-rotate-interval': config.get('token-rotate-interval', None),
+        }])
```
cderici, line 900: Need a test for `add_secret_backends` in `tests/integration/test_controller.py`

juju/controller.py
```diff
@@ -871,6 +871,96 @@ async def _watcher(stop_event):
...
+        reveal : boolean
+            include sensitive backend config content if true
+
+        Returns
+        -------
+        list
+           a list of available secret backends
+        """
+        facade = client.SecretBackendsFacade.from_connection(self.connection())
+        return await facade.ListSecretBackends(None, reveal)
```
cderici, line 917: Need a test for `list_secret_backends` in `tests/integration/test_controller.py`

juju/controller.py
```diff
@@ -871,6 +871,96 @@ async def _watcher(stop_event):
...
+
+        Returns
+        -------
+        error if any
+        """
+        facade = client.SecretBackendsFacade.from_connection(self.connection())
+        return await facade.RemoveSecretBackends([{
+            'name': name,
+            'force': force
+        }])
```
cderici, line 936: Need a test for `remove_secret_backends` in `tests/integration/test_controller.py`

juju/controller.py
```diff
@@ -871,6 +871,96 @@ async def _watcher(stop_event):
...
+            token rotation interval
+        """
+        facade = client.SecretBackendsFacade.from_connection(self.connection())
+        return await facade.UpdateSecretBackends([{
+            'name': name,
+            'config': config,
+            'force': force,
+            'token-rotate-interval': token_rotate_interval,
+            'name-change': name_change,
+        }])
```
cderici, line 962: Need a test for `update_secret_backends` in `tests/integration/test_controller.py`

cderici (changes requested):
Need to bring back the `schemas-juju-latest.json`, so:

```
cd juju/client && ln -s schemas-juju-3.1-beta1.json schemas-juju-latest.json
```

VERSION
juanmanuel-tirado: 3.1.0-b1 is not compatible with the pip versioning and automatically converts that into 3.1.0b1. This is simply to adopt pip's versioning.

juju/client/connection.py
juanmanuel-tirado: No. Finally, it's 16. See [here](https://github.com/juju/juju/blob/173c214fe69660ef9597cae44883d1ce330d7a62/apiserver/facades/client/application/application.go#L63)

juanmanuel-tirado:
In order to test the new secrets_backend facade I have added the `test_secrets_backend_lifecycle` in the `test_controller.py` file. Something like this is needed to test the `add_secret_backend` and additionally, we evaluate the lifecycle of the other elements from the facade. We could split this into several tests if required though.

juju/model.py
juanmanuel-tirado: We should probably address this in a different PR.

juju/client/connection.py
jameinel: we should look to support 17 when it is released, but we can also just support the old method on 14,15,16 (note that we can eventually drop 14 as that is a 2.9 vs 3.0)

cderici (approved):
LGTM

PR #794, opened 2023-01-19
[JUJU-2456] Fix compatibility with app facade version 6

### LINKED ISSUE #792: [Bug]: Nightly release edge testing is failing
### Description

It looks like somewhere around `juju 3.0.3` the `DestroyUnits` in `Application` facade disappears. And it causes the nightly edge to fail.

https://github.com/juju/python-libjuju/actions/runs/3944665126/jobs/6750814571

### Urgency

Casually reporting

### Python-libjuju version

edge

### Juju version

3.0.3-2c6b7fb

### Reproduce / Test

```python
The nightly release edge testing should be passing.

https://github.com/juju/python-libjuju/actions/workflows/test_edge.yaml
```

juanmanuel-tirado:
#### Description

Application facade was updated to version 16 and `DestroyUnits` function was renamed `DestroyUnit`

This fix #792 

#### QA Steps

Nightly build should pass.

#### Notes & Discussion

NA

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -35,7 +35,7 @@ client_facades = {
     'CharmHub': {'versions': [1]},
     'CharmRevisionUpdater': {'versions': [2]},
     'CharmDownloader': {'versions': [1]},
-    'Charms': {'versions': [2, 3, 4, 5]},
+    'Charms': {'versions': [2, 3, 4, 5, 6]},
     'Cleaner': {'versions': [2]},
     'Client': {'versions': [1, 2, 5, 6]},
     'Cloud': {'versions': [1, 2, 3, 4, 5, 7]},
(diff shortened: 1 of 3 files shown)


cderici (changes requested):
So I see two options to move forward here, 

1) safely move to `Application` facade `v16` *without* changing  `DestroyUnits -> DestroyUnit` (the test will pass after  https://github.com/juju/juju/pull/15077 lands), 
2) move further to `v17` and keep your change `DestroyUnits -> DestroyUnit` in which case we also need to get the new schema and regenerate the clients (after https://github.com/juju/juju/pull/15077 lands of course).

juanmanuel-tirado:
We will have to go for the second option because we are promising to support the latest version.

cderici:
> We will have to go for the second option because we are promising to support the latest version.

agreed.

juanmanuel-tirado:
I will close this PR in favour of #791

PR #798, opened 2023-01-31
[JUJU-2573] Base argument for model deploy

cderici:
#### Description

This adds the `base` argument to the `model.Deploy()`, in accordance with the upcoming Juju `3.1`.

#### QA Steps

Adds a unit test for the new `parse_base_arg` utility and an integration test for deploying with base, so the following tests should be passing:

```
tox -e py3 -- tests/unit/test_utils.py::TestBaseArgument::test_parse_base_arg
```

```
tox -e integration -- tests/integration/test_model.py::test_deploy_with_base
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1676,7 +1676,8 @@ class Model:
 
         charm_series = series
         charm_origin = res.origin
-        charm_origin.base = utils.parse_base_arg(base) if base else None
+        if base:
+            charm_origin.base = utils.parse_base_arg(base)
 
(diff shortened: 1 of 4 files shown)


juanmanuel-tirado:
Bases are not properly set for simple deployments on the tests. Setting the base param works. Something is missing in the function logic.

` juju.errors.JujuAPIError: base required for charm-hub charms`

juanmanuel-tirado (approved):
LG2M

PR #806, opened 2023-03-01
Bump allwatcher versions for 29

cderici:
#### Description

Small change to add  the  `AllWatcher` facade versions we support in juju `2.9`. Schema seems to be up to date with that, so no need to recreate clients etc.

#### QA Steps

The tests need to pass. Additionally the following should work fine:

```python
from juju.model import Model
import asyncio

async def juju_stats():
    model = Model()
    await model.connect()
    print(model.is_connected())
    status = await model.get_status()
    print(status)
    await model.disconnect()

asyncio.run(juju_stats())
```

- [ ] Before landing: check and confirm that we actually support the newly added features in `AllWatcher` versions 2 and 3.


jameinel (commented):
looking at the change history, 99% of this PR is actually just a 'sort' on the keys, is that correct? And I believe we are already sorting in the newer versions of python-libjuju, but it does make *this* PR  a bit harder to be sure it is correct.
(I did look at the commit history, thank you for making them 2 separate commits)

SimonRichardson (approved):
Tests need fixing, but I've approved based on the code changes.

cderici:
Closing this, because adding this will make everyone using `unit.run` on `2.9` branch to have to change their codes, this can be observed in the failing CI test https://github.com/juju/python-libjuju/actions/runs/4306236228/jobs/7509795941#step:6:300

Issue #780, opened 2022-12-15
[Bug]: ApplicationStatus of subordinates have no units


tbaumann:
### Description

I noticed that all `ApplicationStatus` results for subordinate charms do not contain any unit information in `units`.

```
╭──────────────────────────────────────────────────────────────────────────────────────────────── <class 'juju.client._definitions.ApplicationStatus'> ────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ │ <class 'juju.client._definitions.ApplicationStatus'>({'can_upgrade_to': '', 'charm': 'local:focal/nrpe-11', 'charm_channel': None, 'charm_profile': '', 'charm_version': '', 'endpoint_bindings': {'': 'alpha', 'general-info': 'alpha',         │ │
│ │ 'local-monitors': 'alpha', 'monitors': 'alpha', 'nrpe': 'alpha', 'nrpe-external-master': 'alpha'}, 'err': None, 'exposed': False, 'exposed_endpoints': {}, 'int_': None, 'life': '', 'meter_statuses': {}, 'provider_id': None,                  │ │
│ │ 'public_address': '', 'relations': {'nrpe-external-master': ['cinder', 'mysql']}, 'series': 'focal', 'status': <class 'juju.client._definitions.DetailedStatus'>({'data': {}, 'err': None, 'info': 'Ready', 'kind': '', 'life': '', 'since':     │ │
│ │ '2022-12-15T13:31:29.79249518Z', 'status': 'active', 'version': '', 'unknown_fields': {}}), 'subordinate_to': ['cinder', 'mysql'], 'units': {}, 'workload_version': '', 'unknown_fields': {'charm-verion': '', 'base': {'name': 'ubuntu',        │ │
│ │ 'channel': '20.04/stable'}}})                                                                                                                                                                                                                    │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                                                                                                                                                                      │
│    can_upgrade_to = ''                                                                                                                                                                                                                               │
│             charm = 'local:focal/nrpe-11'                                                                                                                                                                                                            │
│     charm_channel = None                                                                                                                                                                                                                             │
[...]

PR #808, opened 2023-03-06
Add regression test for ApplicationStatus units empty

tbaumann:
This only adds the regression test, not the fix!

#### Description
Tests bug https://github.com/juju/python-libjuju/issues/780

The units list is empty for all subordinates

#### QA Steps

```
python -m pytest --tb native -rP -v -k 'subordinate_units'
```
or
```
tox -e integration -- tests/integration/test_model.py -k 'subordinate_units' -m 'not serial'
..
```

#### Notes & Discussion

See https://github.com/juju/python-libjuju/issues/780


juanmanuel-tirado (changes requested):
I think it looks OK, I've just updated the branch with the latest changes from master. Just one minor request. Could we use jammy series instead of bionic?

Thanks for the PR

tbaumann:
> I think it looks OK, I've just updated the branch with the latest changes from master. Just one minor request. Could we use jammy series instead of bionic?

Sure, I will change that. 

Do you have an idea on the fix yet? Because if you merge this it will fail the integration until it's fixed. I hope that intention was clear.

cderici:
As I mentioned in https://github.com/juju/python-libjuju/issues/780#issuecomment-1570988234, this looks like expected behavior.

tbaumann:
> I think it looks OK, I've just updated the branch with the latest changes from master. Just one minor request. Could we use jammy series instead of bionic?

Of course

Issue #816, opened 2023-03-15
Error deploying a bundle with offers


sed-i:
The following line seems to be using `join` incorrectly:
https://github.com/juju/python-libjuju/blob/5c4910b1c857abe811951539436748660a511457/juju/bundle.py#L1025

In 3.0.4 this results in:
```
Traceback (most recent call last):
  File "/home/ubuntu/code/cos-lite-bundle/tests/e2e/test_machine.py", line 52, in test_deploy_cos
    await k8s_mdl.deploy(str(rendered_bundle), trust=True)
  File "/home/ubuntu/code/cos-lite-bundle/.tox/e2e/lib/python3.10/site-packages/juju/model.py", line 1757, in deploy
    await handler.execute_plan()
  File "/home/ubuntu/code/cos-lite-bundle/.tox/e2e/lib/python3.10/site-packages/juju/bundle.py", line 402, in execute_plan
    log.info("Applying change: {}".format(change))
  File "/home/ubuntu/code/cos-lite-bundle/.tox/e2e/lib/python3.10/site-packages/juju/bundle.py", line 1063, in __str__
    endpoints = ":{}".format(self.endpoints.join(","))
AttributeError: 'list' object has no attribute 'join'
```

I'm guessing it should be
```python
endpoints = ":{}".format(",".join(self.endpoints)) 
```

The bundle file:
```yaml
---
bundle: kubernetes
name: cos-lite
description: >
  COS Lite is a light-weight, highly-integrated, observability stack running on Kubernetes
applications:
  traefik:
    charm: traefik-k8s
    scale: 1
    trust: true
    channel: edge
  alertmanager:
    charm: alertmanager-k8s
    scale: 1
    trust: true
    channel: edge
  prometheus:
    charm: prometheus-k8s
    scale: 1
    trust: true
    channel: edge
  grafana:
    charm: grafana-k8s
    scale: 1
    trust: true
    channel: edge
  catalogue:
    charm: catalogue-k8s
    scale: 1
    trust: true
    channel: edge
    options:
      title: Canonical Observability Stack
      tagline: Model-driven Observability Stack deployed with a single command.
      description: |
        Canonical Observability Stack Lite, or COS Lite, is a light-weight, highly-integrated, 
        Juju-based observability suite running on Kubernetes.
  loki:
    charm: loki-k8s
    scale: 1
    trust: true
    channel: edge
  avalanche:
    charm: avalanche-k8s
    scale: 2
    trust: true
    channel: edge
[...]

PR #817, opened 2023-03-15
Add an offer to the existing bundle test

sed-i:
#### Description

- [x] Demonstrate #816: reproduced [in CI](https://github.com/juju/python-libjuju/actions/runs/4431006367/jobs/7773478291#step:6:405).
- [x] Commit fix

#### QA Steps

```
tox -e integration -- tests/integration/bundle
```


sed-i:
@juanmanuel-tirado the original error is gone but there's [a new one](https://github.com/juju/python-libjuju/actions/runs/4433509759/jobs/7778631597#step:6:411) now:
```
   File "/home/runner/work/python-libjuju/python-libjuju/juju/offerendpoints.py", line 35, in parse_offer_endpoint
    raise ParseError("endpoint must conform to format \"<application-name>:<endpoint-name>\"")
juju.offerendpoints.ParseError
```
Not sure I'd be able to look into this in the next few days. Feel free to take over if you like.

sed-i:
@juanmanuel-tirado would you have a moment to take a look at the itest failure? Not sure why this fails.

sed-i:
Hey @cderici,
There's a very obvious bug this PR was trying to fix, but I haven't looked into why CI fails.
Would you be able to take a look?

cderici:
> Hey @cderici, There's a very obvious bug this PR was trying to fix, but I haven't looked into why CI fails. Would you be able to take a look?

Absolutely, I'll take a look today, thanks for the reminder, sorry I missed this before.

sed-i:
If CI now passes we may want to restore [the test that I dropped](https://github.com/juju/python-libjuju/pull/817/commits/735236811d442389600ad61a0fb609c51acfb5f1).

cderici:
The CI is failing for unrelated reasons, so I think we can safely land this.

But before that I'd like to see if that test actually reproduces this, in which case we should definitely keep 735236811d442389600ad61a0fb609c51acfb5f1. Unfortunately the logs for the run when you pushed it have expired so I can't see their outputs, but I may go ahead and try it myself manually.

But other than that, I'll give this a thumbsup, it's an obvious fix as you said.

sed-i:
Hi @cderici 
I restored the test. CI shouldn't fail as per issue because the fix is in place, but now it would be easier to confirm locally.

cderici:
@sed-i so the join is part of the problem here, looks like the main offense comes from that we just don't support endpoint syntax  without `:` in pylibjuju https://github.com/juju/python-libjuju/blob/ad62156b3bd18c949cd7f65a6b528cbaf023e121/juju/offerendpoints.py#L34

Fix is coming soon 👍

cderici:
#887 takes over, so closing this. (@sed-i your commits are not discarded, I just moved them to the new PR, feel free to push there as well 👍 )

PR #820, opened 2023-03-23
[JUJU-3348] Fix assumes parsing

### LINKED ISSUE #766: [Bug]: Error deploying Juju Bundles and charms with resources
### Description

As stated on [chat.charmhub.io](https://chat.charmhub.io/charmhub/pl/na6nde1os38b8ki19f5ek4b49w) i've found a scenario where bundle deployment succeeds from the client side (ie -- no exception is raised) but the deployment ultimately fails because the juju controller doesn't provided necessary resources to each charm. 

Here's a [link to the PR](https://github.com/charmed-kubernetes/charm-coredns/pull/34/files/e2c208349c6857476c051179646b9b77dade3fce#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56L49-R55) where i've made an adjustment to deploy with the **juju** snap rather than **python-libjuju**

### Urgency

### Python-libjuju version

3.0.4

### Juju version

2.9.35, 2.9.34

### Reproduce / Test

```python
cmd = f"juju deploy -m {model_full_name} kubernetes-core --channel=latest/edge"
    subprocess.check_call(*shlex.split(cmd))
    # the following leaves the model in a failed state missing charm resources
    # await model.deploy("kubernetes-core", channel="latest/edge")
```

juanmanuel-tirado:
#### Description

This fixes a problematic parsing of assumes in bundles with resources for K8s. This could Fix #766. 

Additionally we update the Charms facade to version 5.

#### QA Steps

Use the bundle mentioned in #766 
```
from juju.model import Model
m = Model()
await m.connect()
LCM_CHARM = "osm-lcm"
LCM_APP = "lcm"
LCM_CHANNEL = "latest/beta"
LCM_SERIES = "focal"
LCM_RESOURCES = {"lcm-image": "opensourcemano/lcm:testing-daily"}
await model.deploy(
    LCM_CHARM,
    resources=LCM_RESOURCES,
    application_name=LCM_APP,
    channel=LCM_CHANNEL,
    series=LCM_SERIES,
)

```

#### Notes & Discussion

This fix has to be ported to 3.* branches.

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -692,7 +692,7 @@ class Type:
                     cls.splitEntries(v, d)
             return cls(**d)
         return None
-    
+
     @classmethod
     def splitEntries(cls, entry, result_dict):
         '''
@@ -711,7 +711,6 @@ class Type:
         else:
             # something like k8s-api
             result_dict[entry] = ''
-        
 
     def serialize(self):
         d = {}
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -675,20 +675,43 @@ class Type:
                 raise
         if isinstance(data, list):
             # check: https://juju.is/docs/sdk/assumes
-            # assumes are in the form of a list
             d = {}
             for entry in data:
-                if '>' in entry or '>=' in entry:
-                    # something like juju >= 2.9.31
-                    i = entry.index('>')
-                    key = entry[:i].strip()
-                    value = entry[i:].strip()
-                    d[key] = value
+                if isinstance(entry, dict):
+                    # this could be
+                    # any-of
+                    # all-of
+                    for _, v in entry.items():
+                        if isinstance(v, list):
+                            for i in v:
+                                cls.splitEntries(i, d)
+                        else:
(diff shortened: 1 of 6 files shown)


juju/client/facade.py
```diff
@@ -675,21 +675,43 @@ def from_json(cls, data):
...
+                if isinstance(entry, dict):
+                    # this could be
+                    # any-of
+                    # all-of
+                    for _, v in entry.items():
+                        if isinstance(v, list):
+                            for i in v:
+                                cls.splitEntries(i, d)
+                        else:
+                            cls.splitEntries(v, d)
```
cderici, line 689: I think having a static method mutating the input in a complex control like this will give us long debugging sessions in the future :) my suggestion would be to avoid side-effects in this part of the code (I mean in the `facade.py` where we generate client codes).

cderici (approved):
Aside from my suggestion, this LGTM. QA went well. Maybe @addyess would like to confirm too that this solves #766 ?

PR #833, opened 2023-04-19
[JUJU-3565] Expect and handle exceptions from the AllWatcher task

### LINKED ISSUE #829: [Bug]: Model connection hangs when migration is in progress
### Description

When libjuju tries to establish connection to a model that is in the middle of a migration, it outputs the follow error messages and hangs without returning to the main program:
```
unknown facade EnvironUpgrader
unexpected facade EnvironUpgrader found, unable to decipher version to use
Task exception was never retrieved
future: <Task finished name='Task-32' coro=<Model._watch.<locals>._all_watcher() done, defined at /home/ubuntu/.venv/lib/python3.10/site-packages/juju/model.py:1058> exception=JujuAPIError('model migration in progress')>
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/model.py", line 1061, in _all_watcher
    model_config = await self.get_config()
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/model.py", line 2017, in get_config
    result = await config_facade.ModelGet()
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/client/facade.py", line 483, in wrapper
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/client/_client2.py", line 10431, in ModelGet
    reply = await self.rpc(msg)
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/client/facade.py", line 656, in rpc
  File "/home/ubuntu/.venv/lib/python3.10/site-packages/juju/client/connection.py", line 627, in rpc
juju.errors.JujuAPIError: model migration in progress
```

It seems that it is not communicating the exception from the controller to the main libjuju thread. An expected behavior would be to raise the exception back to the main program and either triggers exception handling blocks or exit with unclean code.

### Urgency

Blocker for our release

### Python-libjuju version

2.9.11, 2.9.42.1, 3.0.4

### Juju version

2.9.42

### Reproduce / Test

```python
import asyncio

async def juju_stats():
    controller = Controller(max_frame_size=2**27)
    await controller.connect()
    model = await controller.get_model(<model_uuid>)
    status = await model.get_status()
    print(status)
    await model.disconnect()

asyncio.run(juju_stats())
```

cderici:
#### Description

This one was a bit tricky. fixes #829

The `_all_watcher` task is a coroutine for the AllWatcher to run in the background all the time forever, and it involves a while loop that's being controlled manually through some flags (asyncio events), e.g. things like `_watch_stopping`, `watch_stopped`.

The problem is that when the `_all_watcher` raises an exception (or receives one from things like `get_config()` like in the case of #829, that exception is thrown in the background somewhere in ether in the event loop, not handled/or re-raised. This is because this coroutine is not `await`ed (for good reason), it can't be `await`ed because there won't ever be any results, this method is supposed to be working in the background forever getting the deltas for us. As a result of this, if `_all_watcher` fails, then external flags like `_watch_received` is never set, and whoever's calling `await self._watch_received.wait()` will block forever (in this case the `_after_connect()`). Similarly the `disconnect()` waits for the `_watch_stopped` flag, which won't be set either, so if we call disconnect when all_watcher failed then it'll hang forever.

This change fixes this problem by allowing (at the wait-for-flag spots) to wait for two things, 1) whichever flag we're waiting for, 2) `_all_watcher` task to be `"done()"`. In the latter case, we should expect to see an exception because that task is not supposed to be finished. More importantly, if we do see that the `_watcher_task.done()`, then we don't sit and wait forever for the _all_watcher event flags to be set, so we won't hang.

Also a nice side effect of this should be that we should be getting less number of extra exception outputs saying that the "Task exception is never handled", since we do call the `.exception()` on the `_all_watcher` task. Though we'll probably continue to get those from the tasks like `_pinger` and `_debug_log` etc. However, this is a good first example solution to handle them as well.

#### QA Steps

This should be rigorously tested, as it slightly changes a fundamental mechanism. We do need to make sure all the tests are passing for sure.

For the manual QA, what I did was that I changed the body of the `model.get_config()` with `raise JujuError("FOO")` (artificially inducing an error seemingly coming from the api outside of the all_watcher loop). This creates the exact condition happening in the #829. You can also get two controllers and use pylibjuju while running a migration in the background, getting a `migration is in progress` error. Whichever it is, with the error in place, run the following and it should print "Error handled" in the stdout:

```python
async def juju_stats():
    m = Model()
    await m.connect()

try:
    asyncio.run(juju_stats())
except JujuError:
    print("Error handled")
```

#### Notes & Discussion

We might wanna also get this onto the other branches after we carefully test and land this onto `2.9` as requested by the #829 .

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2562,7 +2562,7 @@ class Model:
                         units_ready.add(unit.name)
                         now = datetime.now()
                         idle_start = idle_times.setdefault(unit.name, now)
-                        print(f'unit {unit.name} is waiting since : {idle_start} -- now : {now} -- waiting for : {now - idle_start}')
+
                         if now - idle_start < idle_period:
                             busy.append("{} [{}] {}: {}".format(unit.name,
                                                                 unit.agent_status,
--- a/juju/model.py
+++ b/juju/model.py
@@ -2478,8 +2478,7 @@ class Model:
             warnings.warn("wait_for_active is deprecated; use status", DeprecationWarning)
             status = "active"
 
-        if wait_for_units is None:
-            _wait_for_units = 1
+        _wait_for_units = wait_for_units if wait_for_units is not None else 1
 
         timeout = timedelta(seconds=timeout) if timeout is not None else None
         idle_period = timedelta(seconds=idle_period)
--- a/juju/model.py
+++ b/juju/model.py
@@ -759,7 +759,26 @@ class Model:
         # we've received all the model data, which might be
         # a whole load of unneeded data if all the client wants
         # to do is make one RPC call.
-        await self._watch_received.wait()
+        async def watch_received_waiter():
+            await self._watch_received.wait()
+        waiter = jasyncio.create_task(watch_received_waiter())
+
+        # If we just wait for the _watch_received event and the _all_watcher task
+        # fails (e.g. because API fails like migration is in progress), then
+        # we'll hang because the _watch_received will never be set
+        # Instead, we watch for two things, 1) _watch_received, 2) _all_watcher done
+        # If _all_watcher is done before the _watch_received, then we should see
+        # (and raise) an exception coming from the _all_watcher
+        # Otherwise (i.e. _watch_received is set), then we're good to go
+        done, pending = await jasyncio.wait([waiter, self._watcher_task],
+                                            return_when=jasyncio.FIRST_COMPLETED)
+        if self._watcher_task in done:
+            # Cancel the _watch_received.wait
+            waiter.cancel()
+            # If there's no exception, then why did the _all_watcher broke its loop?
+            if not self._watcher_task.exception():
+                raise JujuError("AllWatcher task is finished abruptly without an exception.")
+            raise self._watcher_task.exception()
 
         await self.get_info()
         self.uuid = self.info.uuid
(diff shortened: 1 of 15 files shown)


juanmanuel-tirado:
We need to increase the testing timeout to 150. Interestingly, all the bundle-related tests are failing.

juju/model.py
```diff
@@ -759,7 +759,26 @@ async def _after_connect(self):
...
+        # Instead, we watch for two things, 1) _watch_received, 2) _all_watcher done
+        # If _all_watcher is done before the _watch_received, then we should see
+        # (and raise) an exception coming from the _all_watcher
+        # Otherwise (i.e. _watch_received is set), then we're good to go
+        done, pending = await jasyncio.wait([waiter, self._watcher_task],
+                                            return_when=jasyncio.FIRST_COMPLETED)
+        if self._watcher_task in done:
+            # Cancel the _watch_received.wait
+            waiter.cancel()
+            # If there's no exception, then why did the _all_watcher broke its loop?
```
SimonRichardson, line 778:
```suggestion
            # If there's no exception, then why did the _all_watcher break its loop?
```

juju/model.py
```diff
@@ -759,7 +759,26 @@ async def _after_connect(self):
...
+            if not self._watcher_task.exception():
+                raise JujuError("AllWatcher task is finished abruptly without an exception.")
```
SimonRichardson, line 780: I'd expect you'd see logs in juju about a connection loss, otherwise, I'd say there was a bug in the socket library.

cderici: Yeah we should never ever hit this case, if _all_watcher stops there has to be an exception because the other option is that it returns something which is not possible (it literally doesn't have a `return` in it :) )

cderici:
Cherry-picking and manually adding fixes for the integration tests from https://github.com/juju/python-libjuju/pull/774/

juanmanuel-tirado (approved):
LG2M

PR #825, opened 2023-04-05
[JUJU-3404] Pass series info into origin for ResolveCharm

cderici:
#### Description

Without the series the ResolveCharm will select the latest base for the charm. E.g. even if we want `focal`, the `ResolveCharms` will return `22.04` (jammy) for the base channel in the resulting origin if the charm supports both jammy and focal.

This communicates the series info with the `ResolveCharms` via the inputted origin so the resulting origin will have the correct base channel.

Fixes #822 

#### QA Steps

This is against 2.9, so get a juju 2.9 and bootstrap a `2.9` controller.

```sh
 $ sudo snap install juju --channel=2.9/stable --classic
 $ juju bootstrap localhost test
```
Then copy paste the following code from #822 deploy command into the `examples/deploy.py`

```python
LCM_CHARM = "osm-lcm"
LCM_APP = "lcm"
LCM_CHANNEL = "latest/beta"
LCM_SERIES = "focal"
LCM_RESOURCES = {"lcm-image": "opensourcemano/lcm:testing-daily"}

async def deploy():
    model = Model()
    await model.connect_current()
    try:
        await model.deploy(
            LCM_CHARM,
            application_name=LCM_APP,
            resources=LCM_RESOURCES,
            channel=LCM_CHANNEL,
            series=LCM_SERIES,
        )
    finally:
        await model.disconnect()
```

Then just run it from the top:

```sh
 $ python examples/deploy.py
```

After starting this (doesn't need to return), just check the `juju status` and you should see the revision 174 according to the `juju info osm-lcm`.

I suggest trying it with different channels and revisions, as well as different charms too. I did try with different channels and it seems to be working well. I haven't tried it with different charms.

Also we need to be careful about the CI tests, as this might change things, so we need to make sure all the charm related teests are passing.

#### Notes & Discussion

I haven't tried this in > 3.0, we might need to cherry pick this into the master as well.

PR #834, opened 2023-04-19
Fix _resolve_charm errors

cderici:
#### Description

This adds some changes to fix some errors about the `_resolve_charm` method that are accidentally introduced in #825.

* `_resolve_charm` force parameter is made optional
* fixed the `_resolve_charm` calls in the `bundle.py` to have less number of variables to unpack onto

#### QA Steps

The ones that are `FAILED` in [#825's CI run ](https://github.com/juju/python-libjuju/actions/runs/4622441500/jobs/8175154975) need to be passing.

In fact all of the CI tests need to pass.

### DIFF
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -358,7 +358,7 @@ class BundleHandler:
                                             architecture=architecture,
                                             risk=risk,
                                             track=track)
-                charm_url, charm_origin, _ = await self.model._resolve_charm(charm_url, origin)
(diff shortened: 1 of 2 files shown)


cderici:
Ok for this the only test that's failing is that ubuntu charm storage thing : `juju.errors.JujuAPIError: cannot upgrade application "ubuntu" to charm "local:focal/ubuntu-0": required storage "files" removed`

so I think we can merge this to at least unblock the tests for other PRs to be able to run their CI tests

PR #840, opened 2023-04-20
[JUJU-3583] wait_for_idle to not block when enough units are ready

### LINKED ISSUE #837: [Bug]: ops_test.model.wait_for_idle does not respect argument
### Description

wait_for_units has following definition:
```
param wait_for_units (int): The least number of units to be expected before
            going into the idle state.
            The default is 1 unit.
```

in our scenario juju for some reason scaled up the app, so, we have 2/1. But I do not care about the second unit, since we can work only with leader

now all the tests hang, since second unit is in waiting status and waits for the leadership.

if I say `wait_for_units=1`, then logically it should say we are good even if only 1 unit is in idle/active

image

### Urgency

### Python-libjuju version

2.9.42.1

### Juju version

2.9.42

### Reproduce / Test

```python
    await ops_test.model.wait_for_idle(
        apps=apps,
        status="active",
        raise_on_blocked=False,
        raise_on_error=False,
        timeout=from_minutes(minutes=100),
    )
```

cderici:
#### Description

`wait_for_idle` will keep waiting (until timeout) if there are less number of units available than requested (via the `wait_for_units` flag). This is a correct behavior. However, if there are already a number of units in a desired status ready to go, more than (or equal to) `wait_for_units`, then it shouldn't block until the other not-yet-available units to get into that desired state as well. This is because when we first introduced that flag we respected the former case, however, we did not consider the latter case where the `wait_for_idle` to return when that number of units is reached (without further blocking for whatever reason).

fixes #837 

#### QA Steps

An integration test is added for this particular case. That needs to pass. Also the other `wait_for_idle` related integration tests should be passing too.

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_more_units_than_needed
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2457,6 +2457,7 @@ class Model:
         start_time = datetime.now()
         apps = apps or self.applications
         idle_times = {}
+        units_ready = set()  # The units that are in the desired state
         last_log_time = None
         log_interval = timedelta(seconds=30)
 
@@ -2478,6 +2479,9 @@ class Model:
                 ))
 
         while True:
+            # The list 'busy' is what keeps this loop going,
+            # i.e. it'll stop when busy is empty after all the
+            # units are scanned
             busy = []
             errors = {}
             blocks = {}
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado:
Build requested after #834 was landed.

beliaev-maksim:
@cderici @juanmanuel-tirado 
not good

you break the outer loop. If any of the units from the entire deployment goes into "ready", then loop is broken and all code exits providing false positive.

Issue #839, opened 2023-04-20
[Bug]: Local Proxy Charm in LXD fails to be deployed


gcalvinos:
### Description

When trying to deploy a local charm in LXD python libjuju returns the following error:

2023-04-20T07:44:50 ERROR lcm.ns ns.py:4829 Task ns=8eb21604-9e55-4e08-ae17-c2d260750f32 instantiate=0beca26a-4c7b-44da-8dd2-eb22b13d9747 Deploying VCA vnf2.: Install configuration Software Error deploying charm into ee=8eb21604-9e55-4e08-ae17-c2d260750f32.app-vnf-ea396e3481-z0-uce1s.1 : **('Unknown version : %s', '')**

I checked that this error happens in the method _deploy_ in model.py, when it calls the following line:

`base = utils.get_local_charm_base(
         charm_series, channel, metadata, charm_dir, client.Base)`

### Urgency

Blocker for our release

### Python-libjuju version

2.9.42.1

### Juju version

2.9.42

### Reproduce / Test

```python
To reproduce this error I used the following Charm:
https://osm.etsi.org/gitlab/vnf-onboarding/osm-packages/-/tree/master/charm-packages/nopasswd_proxy_charm_vnf/charms/simple

I already commented this issue with @juanmanuel-tirado
```

gcalvinos:
After testing the 2.9 branch, I still see the same issue:

```
+-------------------------+--------------------------------------+---------------------+----------+-------------------+------------------------------------------+
| ns instance name        | id                                   | date                | ns state | current operation | error details                            |
+-------------------------+--------------------------------------+---------------------+----------+-------------------+------------------------------------------+
| nopasswd_proxy_charm-ns | e8d239e0-2638-43fd-bae8-8a121c090180 | 2023-05-09T08:32:20 | BROKEN   | IDLE (None)       | Operation: INSTANTIATING.aa5f365f-ac45-4 |
|                         |                                      |                     |          |                   | 2a0-a5b3-9d138b83424d, Stage 2/5:        |
|                         |                                      |                     |          |                   | deployment of KDUs, VMs and execution    |
|                         |                                      |                     |          |                   | environments.                            |
|                         |                                      |                     |          |                   | Detail: Deploying VCA vnf2.: Install     |
|                         |                                      |                     |          |                   | configuration Software Error deploying   |
|                         |                                      |                     |          |                   | charm into ee=e8d239e0-2638-43fd-        |
[...]

PR #846, opened 2023-05-09
[JUJU-3641] Fix local charm base channel discovery

cderici:
#### Description

`utils.get_local_charm_base()` was incorrectly using the `--channel` argument (the charm's channel) for discovering the channel part of the base. (we should stop using the word 'channel' for two different things).

This fixes that by taking out the incorrect part of the code.

Should fix https://github.com/juju/python-libjuju/issues/839

#### QA Steps

So a trivial way to reproduce the #839 is to deploy a local charm with a `--channel='stable'` argument. You may try to use one of the examples to validate this. Additionally an integration test is added (see below), so passing that should be enough. I also suggest getting a confirmation from [@gcalvinos](https://github.com/gcalvinos), just in case.

```
tox -e integration -- tests/integration/test_model.py::test_deploy_local_charm_channel
```

All CI tests need to pass. We might have a couple of time-outs from previously known CI test failures.

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -1682,13 +1682,11 @@ class Model:
                 # We have a local charm dir that needs to be uploaded
                 charm_dir = os.path.abspath(os.path.expanduser(identifier))
                 charm_origin = res.origin
-                base = None
 
                 metadata = utils.get_local_charm_metadata(charm_dir)
                 charm_series = charm_series or await get_charm_series(metadata,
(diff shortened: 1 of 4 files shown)


juanmanuel-tirado:
I have added some fixes to the CI covering the latest changes in the postgresl charm.

juanmanuel-tirado (approved):
LG2M

PR #849, opened 2023-05-12
[JUJU-1467] Application status from API

cderici:
#### Description

Determining the status of an application is a convoluted and logically a bit tricky process in Juju. Some applications are reported as in the `error` state not because of its units' `workload_status` or the `agent_status`, but for instance in caas models because of the `operator_status`, and through the API this can only be observed in the `DetailedStatus` of an `Application` coming from the `FullStatus` (and not through our conventional way of getting a delta about it via the AllWatcher). Therefore pylibjuju is unable to see that the application is in the error state, while in the `juju status` we clearly see that it is, in fact, in the error state.

[This LP bug](https://bugs.launchpad.net/juju/+bug/1981833) is an example of this scenario (see the QA steps below to reproduce it).

This change remedies this by introducing `application.get_status()` that calls the `FullStatus` to get the most up to date status info from the API and include that in our `derive_status` process.

#### QA Steps

This requires a caas model, as it's not observed in machine models. (this is because the `admission-webhook` charm transitions into the error state for different reasons in machine models than on the caas models). Additionally, for the same reason, it's not possible to make an integration test as our tests are set up to run on the lxd. Though maybe devising an elaborate setup to simulate that behavior through the API might be possible.

Bootstrap a k8s model on microk8s.
```
 $ juju bootstrap microk8s removeme
```

Deploy the `admission-webhook` charm.

```
 $ juju deploy admission-webhook --channel 1.4/stable
```

It will go into the `error` state. Confirm this with `juju status`. It may take a couple seconds to get there.

Now run `wait_for_idle` as follows. I modified one of the examples in the `examples` folder.

```python
await model.wait_for_idle(
        apps=["admission-webhook"],
        status="active",
        raise_on_blocked=True,
        raise_on_error=True,
        timeout=1500,
    )
```

As also said in the reported [LP bug](https://bugs.launchpad.net/juju/+bug/1981833), without this PR this `wait_for_idle` call returns, because the application status is set as `active` (because all the unit workload_statuses are active and the agent_status is idle). With this change, it will correctly raise the `JujuAppError` because the application's status is correctly set to `error`, coming through the API.

#### Notes & Discussion

This should probably be cherry-picked onto the main branch later on.

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -91,10 +91,15 @@ class Application(model.ModelEntity):
         """
         status = self.safe_data['status']['current']
         if status == "unset":
-            unit_status = []
+            known_statuses = []
             for unit in self.units:
-                unit_status.append(unit.workload_status)
-            return derive_status(unit_status)
+                known_statuses.append(unit.workload_status)
+            # If the self.get_status() is called (i.e. the status
+            # is received by FullStatus from the API) then add
+            # that into this computation as it might be more up
+            # to date (and more severe).
+            known_statuses.append(self._status)
+            return derive_status(known_statuses)
         return status
 
     @property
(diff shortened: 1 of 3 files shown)


juanmanuel-tirado (approved):
LG2M

I ran some manual QA and it works. Do not forget to set bug to `fix committed` in Launchpad. Good job!

Issue #851, opened 2023-05-15
[Bug]: local variable referenced before assignment


beliaev-maksim:
### Description

```
 Traceback (most recent call last):
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/tests/integration/test_charm.py", line 45, in test_relations
    await ops_test.model.deploy(ISTIO_PILOT, channel="1.16/stable", trust=True)
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1664, in deploy
    resources = await self._add_charmhub_resources(res.app_name,
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/model.py", line 1797, in _add_charmhub_resources
    res = await charm_facade.CharmInfo(entity_url)
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 502, in wrapper
    result = cls.from_json(reply['response'])
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 673, in from_json
    return cls(**d)
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/_definitions.py", line 4890, in __init__
    meta_ = CharmMeta.from_json(meta) if meta else None
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 673, in from_json
    return cls(**d)
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/_definitions.py", line 5402, in __init__
    assumes_expr_ = ExpressionTree.from_json(assumes_expr) if assumes_expr else None
  File "/home/runner/work/oidc-gatekeeper-operator/oidc-gatekeeper-operator/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 692, in from_json
    cls.splitEntries(v, d)
[...]

PR #854, opened 2023-05-16
[JUJU-3750] Fix bug in Type.from_json() parsing simple entries

### LINKED ISSUE #850: [Bug]: Libjuju errors out if simple entries are present in a charmhub charm's metadata
### Description

In my case the error seems to be caused if assumes juju or assumes k8s-api are present in the charm's metadata

I think the root cause is the changes made to facade.py in https://github.com/juju/python-libjuju/commit/4eb26c7dd46508dcbc7d85950407ebcd465dccbf

Errors in Github CI:
- [Assumes k8s-api](https://github.com/canonical/pgbouncer-k8s-operator/actions/runs/4954060987/jobs/8909145210)
- [Assumes juju](https://github.com/canonical/pgbouncer-operator/actions/runs/4979717669/jobs/8911712013)

Traceback:
```
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/model.py", line 1664, in deploy
    resources = await self._add_charmhub_resources(res.app_name,
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/model.py", line 1797, in _add_charmhub_resources
    res = await charm_facade.CharmInfo(entity_url)
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/facade.py", line 502, in wrapper
    result = cls.from_json(reply['response'])
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/facade.py", line 673, in from_json
    return cls(**d)
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/_definitions.py", line 4890, in __init__
    meta_ = CharmMeta.from_json(meta) if meta else None
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/facade.py", line 673, in from_json
    return cls(**d)
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/_definitions.py", line 5402, in __init__
    assumes_expr_ = ExpressionTree.from_json(assumes_expr) if assumes_expr else None
  File "/home/ubuntu/workspace/pgbouncer-operator/.tox/backend-integration/lib/python3.10/site-packages/juju/client/facade.py", line 692, in from_json
    cls.splitEntries(v, d)
UnboundLocalError: local variable 'v' referenced before assignment
```

### Urgency

### Python-libjuju version

2.9.42.2

### Juju version

2.9.42

### Reproduce / Test

```python
import asyncio
from juju.model import Model
model = Model()
await model.connect_current()

await model.deploy("postgresql", channel="14/edge")
```

cderici:
#### Description

This fixes a bug found in the `facade.py` parsing simple entries in the `Type.from_json()` function.

Should fix #850 & #851

#### QA Steps

To go with the same way of #850 and #851 to reproduce this, using juju `2.9.42`, bootstrap a k8s model (I did it on microk8s).

```sh
 $ juju bootstrap microk8s removeme
```

Then get a libjuju console on the toplevel dir:

```sh
 $ python -m asyncio
```

Then the following should work just fine:

```sh
from juju import model;m = model.Model(); await m.connect(); await m.deploy("postgresql", channel="14/edge")
```

Also added an integration test that should work (fails without this change):

```tox
 tox -e integration -- tests/integration/test_model.py::test_deploy_charm_assumes
```

which just runs `await model.deploy('postgresql', channel='14/edge')`

Note that this is not the ideal test because it depends on that the upstream charm having the simple

assumes:
  -juju

expression. However, simulating the bug in the facade.py for parsing such expressions is non-trivial as we need a test to call something like the CharmsFacade.CharmInfo() to trigger parsing the metadata (which is actually where it fails in the reported bug). Creating a local charm wouldn't work because we locally handle the metadata (wihtout going through anything in the facade.py where the bug is located). Maybe we can manually call AddCharm in the test for a local charm and then manually call the CharmInfo with the returned url.

All CI tests need to pass.

#### Notes & Discussion

This will need to be forward ported into the master.

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -689,7 +689,7 @@ class Type:
                             cls.splitEntries(v, d)
                 else:
                     # this is a simple entry
-                    cls.splitEntries(v, d)
+                    cls.splitEntries(entry, d)
             return cls(**d)
         return None
 
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -151,6 +151,13 @@ async def test_deploy_local_charm(event_loop):
 
 @base.bootstrapped
 @pytest.mark.asyncio
+async def test_deploy_charm_assumes(event_loop):
+    async with base.CleanModel() as model:
+        await model.deploy('postgresql', channel='14/edge')
+
+
+@base.bootstrapped
+@pytest.mark.asyncio
 async def test_deploy_local_charm_channel(event_loop):
     charm_path = TESTS_DIR / 'charm'
 


SimonRichardson (commented):
Can we add a test for this?

cderici:
> Can we add a test for this?

It's a little weird, since the bug only surfaces with the `assumes` directive, but I suppose I could just make a test charm that has a similar dummy expression that'd trigger this, I'll give it a shot, thanks! 👍

SimonRichardson (approved):

PR #858, opened 2023-05-19
Revisit the secret backend integration test

cderici:
#### Description

Both changing the postgresql series to jammy (we should change all of these to use `--base` at some point) as per
https://github.com/juju/juju/pull/15616, and also increasing the timeout appears to be turning this test from timeout to green in local setup, we should see a similar effect on the CI as well.

#### QA Steps

The `test_secrets_backend_lifecycle` should be passing in the CI without timing out🤞 

It can be run manually for local setup on lxd as  well:

```
tox -e integration -- tests/integration/test_controller.py::test_secrets_backend_lifecycle
```

Other tests using the postgresql are also revisited for consistency:

```
tox -e integration -- tests/integration/test_model.py::test_deploy_from_ch_channel_revision_success
```

```
 tox -e integration -- tests/integration/test_model.py::test_add_and_list_storage
```


juanmanuel-tirado:
For consistency purposes, we should revisit all the `postgresql` charms we're using. There is another one in the `test_add_and_list_storage` test.

Again it looks like there are some issues with persistency and GitHub machinery. The failing tests works in my local environment. I will rerun just in case.

juanmanuel-tirado (approved):
R2GO

PR #873, opened 2023-06-05
Update readme and add some docstrings for functions

cderici:
#### Description

Small change that updates the example in the readme and adds some docstrings for `model.py` and `bundle.py`.

#### QA Steps

No change in any functionality, so no QA is needed. Just some proof read, maybe.


juanmanuel-tirado (approved):
LG2M

Issue #886, opened 2023-06-20
`wait_for_idle` returns prematurely when `wait_for_units` is provided 


sed-i:
When `wait_for_units` is provided, we exit the loop before checking the `waiting_for_a_particular_status` logic.
https://github.com/juju/python-libjuju/blob/ad62156b3bd18c949cd7f65a6b528cbaf023e121/juju/model.py#L2693-L2696

As a result, when we use something like:
```python
    await ops_test.model.wait_for_idle(
        apps=[grafana_app_name, tester_app_name], status="active", wait_for_units=1, timeout=300
    )
```
the await returns before status is active.

cderici:
We had some iterations for this on the `wait_for_idle` code, but it was on the `2.9` branch, we brought it forward into the master in https://github.com/juju/python-libjuju/pull/870, but evidently we missed something, I'll take another look. 

@sed-i Just to make sure, this is against 3.x controller using libjuju >2.9, right?

sed-i:
Yep,
- juju (3.1/edge) 3.1.4-8fbf7ef
- pylibjuju 3.2.0.0

cderici:
I'm not sure if this is a bug, @sed-i. Problem is that this use case is clashing with the use case in https://github.com/juju/python-libjuju/issues/837. 

In your particular case, you're trying to wait for `grafana_app_name` and `tester_app_name`, what was your reasoning/expectation in passing `wait_for_units=1`? Cause right now it's set to "the least number of units to be expected to be in the status" [here](https://github.com/juju/python-libjuju/blob/ad62156b3bd18c949cd7f65a6b528cbaf023e121/juju/model.py#L2613), which is why it's returning as soon as 1 unit is in the given status. We have another parameter called the [`wait_for_exact_units`](https://github.com/juju/python-libjuju/blob/ad62156b3bd18c949cd7f65a6b528cbaf023e121/juju/model.py#LL2617C16-L2617C37) to wait for exact number of units. Note that it's across multiple applications if you provide multiple apps.

The main problem is that `wait_for_idle` is super overloaded, and it's trying to work across multiple applications at the same time. I need to find a good way for users to express `I want this number of units from applications A and B to be in <status-1> (don't care which app they belong)` and `I don't care about application C's units (they can be in error state) as long as application D has at least 1 unit in <status-2>` at the same time, which is a bit complicated, but doable.

sed-i:
> The main problem is that `wait_for_idle` is super overloaded

Agreed.

[...]

PR #890, opened 2023-06-26
[JUJU-4076] Rename `wait_for_units` and make semantics clearer

cderici:
#### Description

This is to clear a bit the semantics of `wait_for_units` parameter in `model.wait_for_idle`, which is to wait for a certain amount of units across all the given applications and return when those are available, regardless of the state of the other units (they can be in an error state or whatever just don't care and return)

Fixes https://github.com/juju/python-libjuju/issues/886

#### QA Steps

We got a couple of tests specific to `wait_for_idle` and `wait_for_units`, so all of them should pass, some important ones are the following:

```
tox -e py3 -- tests/unit/test_model.py::TestModelWaitForIdle
```

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_enough_units
```

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_exact_units
```

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_exact_units_scale_down
```

```
tox -e integration -- tests/integration/test_model.py::test_wait_for_idle_with_exact_units_scale_down_zero
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -2622,7 +2622,7 @@ class Model:
             warnings.warn("wait_for_active is deprecated; use status", DeprecationWarning)
             status = "active"
 
-        _wait_for_units = wait_for_units if wait_for_units is not None else 1
+        _wait_for_units = wait_for_at_least_units if wait_for_at_least_units is not None else 1
 
         timeout = timedelta(seconds=timeout) if timeout is not None else None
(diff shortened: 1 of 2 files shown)


jack-w-shaw (approved):
LGTM

PR #907, opened 2023-07-12
[JUJU-4048] Use GetChangesMapArgs for bundle changes

jack-w-shaw:
Use GetChangesMapArgs for bundle changes instead of GetChanges

Also, drop support for GetChanges when parsing ChangeInfos

GetChanges returns a list of params in a magical order. This makes it brittle and adds lots of special knowledge boiler plate. GetChangesMapArgs was added a few years ago to replace it.

 Wer already support map arg params, we just never do so

#### QA Steps

```
tox -e py3
```

```
tox -e integration -- -m bundle
```

### DIFF
--- a/juju/bundle.py
+++ b/juju/bundle.py
@@ -12,7 +12,7 @@ import yaml
 from toposort import toposort_flatten
 
 from .client import client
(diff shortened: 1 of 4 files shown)


cderici (approved):
This LGTM 👍 Thanks for marking the tests!

PR #922, opened 2023-07-20
Repository Maintenance Improvements

cderici:
#### Description

This adds a couple of upgrades/fixes for repository maintenance:

- Brings back the issue/feature request templates
- Adds a github action/stale bot to automatically decay the issues/PRs that are not updated for a long time. In the current configuration, every issue/PR that hasn't been updated for 30 days will be marked with an `incomplete` label, and will be closed with a nice message if it's still not updated for 5 days after that mark.
- Adds a document for contributing guidelines for outside contributions. The page will look like this: https://github.com/cderici/cderici.github.io/blob/master/docs/CONTRIBUTING.md

#### QA Steps

No QA needed.


anvial (approved):
btw, am I right that if we remove the `incomplete` label from issue that will drop the timer. 

If not,  I think that we should add to `exempt labels`  branch labels (2.9/3.0/...)m because for us that means triage.

cderici:
Just for future reference I'll note my answers here:

> btw, am I right that if we remove the `incomplete` label from issue that will drop the timer.

This is correct.

> If not, I think that we should add to `exempt labels` branch labels (2.9/3.0/...)m because for us that means triage.

Already added some exempt labels to express these cases (awaiting approval/work in progress) without worrying about the stale bot.
 
```yaml
exempt-issue-labels: 'awaiting-approval,work-in-progress'
exempt-pr-labels: 'awaiting-approval,work-in-progress'
```

PR #930, opened 2023-08-07
Fix asyncio on README

marceloneppel:
#### Description

The following error is thrown if we run the example code from README.
```sh
  File "/home/ubuntu/main.py", line 6, in <module>
    from juju import asyncio
ImportError: cannot import name 'asyncio' from 'juju' (/home/linuxbrew/.linuxbrew/Cellar/python@3.11/3.11.4_1/lib/python3.11/site-packages/juju/__init__.py)
```

It fails because it should be using `jasyncio`, which is the correct module to import. The the other examples from the `examples` folder are using the correct module.

#### QA Steps

Add the code from README to a file named `main.py` and run it with `python3 main.py`. The ubuntu charm should be correctly deployed.

### DIFF
--- a/docs/readme.rst
+++ b/docs/readme.rst
@@ -51,7 +51,7 @@ Pylibjuju requires an already bootstrapped Juju controller to connect to.
   import logging
   import sys
 
-  from juju import asyncio
+  from juju import jasyncio
   from juju.model import Model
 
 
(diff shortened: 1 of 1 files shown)


cderici (approved):
LGTM. Thanks for the fix!

PR #931, opened 2023-08-07
Update 3.2 facade clients

cderici:
#### Description

This does 3 things:

1. Adds a tiny fix to suppress redundant warning messages about facade versions to be emitted in case the facade in question is not yet known to the pylibjuju client.
2. Adds the `AgentLifeFlag` to known facades ([reported in a nightly build](https://github.com/cderici/python-libjuju/actions/runs/5779670269/job/15662283086#step:8:383))
3. While we're at the facades, this adds the latest schema json from juju 3.2 branch and rebuilds the clients to keep the compatibility.

#### QA Steps

All CI tests need to pass.


cderici:
`test_action` problem is being fixed by #932

PR #932, opened 2023-08-07
Fix integration/test_application.test_action

cderici:
#### Description

This fixes an issue (we see [in the nightly build](https://github.com/cderici/python-libjuju/actions/runs/5779670269/job/15662283086#step:8:334)) with the `test_action` integration test that's seemingly caused by the percona-cluster cham (no code about the `get_actions` has changed in the pylibjuju side). In this PR we just change that to the `mysql` charm to have this test pass again.

#### QA Steps

```
tox -e integration -- tests/integration/test_application.py::test_action
```

### DIFF
--- a/tests/integration/test_application.py
+++ b/tests/integration/test_application.py
@@ -16,39 +16,10 @@ logger = logging.getLogger(__name__)
 @pytest.mark.asyncio
 async def test_action(event_loop):
     async with base.CleanModel() as model:
-        mysql_app = await model.deploy(
-            'mysql',
-            application_name='mysql',
-            channel='8.0/stable',
-            config={
-                'cluster-name': 'cluster1',
-            },
-            constraints={
-                'arch': 'amd64',
-                'mem': 256 * MB,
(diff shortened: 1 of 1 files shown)


tests/integration/test_application.py
```diff
@@ -16,13 +16,12 @@
 @pytest.mark.asyncio
 async def test_action(event_loop):
     async with base.CleanModel() as model:
-        ubuntu_app = await model.deploy(
-            'percona-cluster',
+        mysql_app = await model.deploy(
+            'mysql',
```
jack-w-shaw, line 20: Perhaps we should deploy one of our `juju-qa-.*` test charms?

jack-w-shaw (approved):
LGTM. One suggestion to consider

cderici:
> LGTM. One suggestion to consider

Oh sorry I didn't refresh the page before assigning you again. Thanks for the review, and the suggestion, I'll see if we can use one of our test charms 👍

PR #935, opened 2023-08-23
Update async tests to use builtin python suite

DanielArndt:
#### Description

As of Python 3.8, `AsyncMock` and many other async testing tools are available as standard libraries. This PR removes the third-party library `asynctest` and replaces it with those available in the standard python library.

As part of this, the tests will now run on Python 3.11, which they previously did not.

Additionally, warnings during tests are significantly reduced (from hundreds, to tens).

#### QA Steps

Running the unit and integration tests is hopefully sufficient

#### Notes & Discussion

I went down a rabbit hole here while investigating something totally unrelated. I understand if this change needs a more thorough review which there may or may not be time for, but I thought I'd put up the change since I was having some fun digging in.


cderici:
@DanielArndt thanks for the change! Let's get this in a working order and I'd be happy to give it a thorough review. For that, please fix the linter and let's see what's going on in the CI and I'll take it from there 👍

DanielArndt:
> @DanielArndt thanks for the change! Let's get this in a working order and I'd be happy to give it a thorough review. For that, please fix the linter and let's see what's going on in the CI and I'll take it from there 👍

Sure. There's some pre-existing issues that I've separated here #936

cderici (approved):
CI test failures are known intermittent errors unrelated to this change. They will be fixed soon in a subsequent PR.

Thanks again for the change! Good cleanup for sure 👍

PR #937, opened 2023-08-25
Handle FileNotFoundError on current_controller()

DanielArndt:
#### Description

When Juju is not bootstrapped, `controllers.yaml` does not yet exist. This throws a `FileNotFoundError` which is hard to decipher the root cause.

This is a bit of a contract change, but it appears the only place this is used (at least internally) already guards against `None`. I added a hint to the error message in this case to indicate that Juju may not be bootstrapped.

See Notes & Discussion for alternatives.

#### QA Steps

- Try calling `jujudata.current_controller()` on a machine that is not yet bootstrapped.

#### Notes & Discussion

An alternative to this might be raising a `JujuError` directly from `FileJujuData.current_controller()`. 

This should also probably be applied to `FileJujuData.controllers()`, `FileJujuData.models()`, `FileJujuData.accounts()`, and `FileJujuData.credentials()` calls.

Happy to make changes here if there's a preferred approach, but I thought I'd open the discussion.

### DIFF
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -5,12 +5,14 @@ import copy
 import logging
 
 import macaroonbakery.httpbakery as httpbakery
+
+from juju.client import client
 from juju.client.connection import Connection
 from juju.client.gocookies import GoCookieJar, go_to_py_cookie
-from juju.client.jujudata import FileJujuData, API_ENDPOINTS_KEY
+from juju.client.jujudata import API_ENDPOINTS_KEY, FileJujuData
 from juju.client.proxy.factory import proxy_from_config
-from juju.errors import JujuConnectionError, JujuError, PylibjujuProgrammingError
-from juju.client import client
+from juju.errors import (JujuConnectionError, JujuError,
+                         PylibjujuProgrammingError)
 from juju.version import SUPPORTED_MAJOR_VERSION, TARGET_JUJU_VERSION
 
 log = logging.getLogger('connector')
@@ -113,8 +115,6 @@ class Connector:
         """
         if not controller_name:
             controller_name = self.jujudata.current_controller()
-        if not controller_name:
-            raise JujuConnectionError('No current controller')
 
         controller = self.jujudata.controllers()[controller_name]
         endpoints = controller[API_ENDPOINTS_KEY]
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -12,30 +12,31 @@ from juju.client.gocookies import GoCookieJar, go_to_py_cookie
 from juju.client.jujudata import API_ENDPOINTS_KEY, FileJujuData
 from juju.client.proxy.factory import proxy_from_config
 from juju.errors import JujuConnectionError, JujuError
-from juju.client import client
 from juju.version import SUPPORTED_MAJOR_VERSION, TARGET_JUJU_VERSION
 
-log = logging.getLogger('connector')
+log = logging.getLogger("connector")
 
 
 class NoConnectionException(Exception):
-    '''Raised by Connector when the connection method is called
-    and there is no current connection.'''
+    """Raised by Connector when the connection method is called
+    and there is no current connection."""
+
     pass
 
 
 class Connector:
-    '''This class abstracts out a reconnectable client that can connect
+    """This class abstracts out a reconnectable client that can connect
     to controllers and models found in the Juju data files.
-    '''
+    """
+
     def __init__(
         self,
         max_frame_size=None,
         bakery_client=None,
         jujudata=None,
     ):
-        '''Initialize a connector that will use the given parameters
-        by default when making a new connection'''
+        """Initialize a connector that will use the given parameters
+        by default when making a new connection"""
         self.max_frame_size = max_frame_size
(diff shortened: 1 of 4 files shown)


cderici:
@DanielArndt thanks for the PR! Looks like the `test_jujudata.py` needs a copyright header. If you could fix that, I'd be happy to take over and review/land etc 👍

juju/client/connector.py
```diff
@@ -103,7 +103,7 @@ async def connect_controller(self, controller_name=None, specified_facades=None)
         if not controller_name:
             controller_name = self.jujudata.current_controller()
         if not controller_name:
-            raise JujuConnectionError('No current controller')
+            raise JujuConnectionError('No current controller. Is Juju bootstrapped?')
```
cderici, line 106: I'm not a fan of `current_controller()` returning `None` and us interpreting it here. I think we should just raise this error at the spot and have it bubble up.

juju/client/jujudata.py
```diff
@@ -75,7 +75,10 @@ def refresh(self):
 
     def current_controller(self):
         '''Return the current controller name'''
-        return self._load_yaml('controllers.yaml', 'current-controller')
+        try:
+            return self._load_yaml('controllers.yaml', 'current-controller')
+        except FileNotFoundError:
+            return None
```
cderici, line 81: This is the spot at which we should raise the `JujuConnectionError` ☝️

cderici (changes requested):

juju/client/connector.py
DanielArndt: I did mention the exception approach in the "Notes & Discussion", but here is why I leaned toward `None`:

The class we're calling on (`FileJujuData`) is an interface to the Juju files. We're asking it "what do the files say the current controller is?", and I think the correct answer is actually "the files say there isn't one".

It's not really a connection error in `FileJujuData`, but it does become a connection error in the client since there is no controller for us to connect to.

There might be a better argument for raising a `JujuError`, but is that even true? Is it really an error if this file doesn't exist, or does it just mean some things haven't been set up yet, and therefore some operations are restricted? The distinction is important, because this only becomes an error once you decide to do something that requires Juju to have a controller set.

DanielArndt: Well, I managed to beat your late response with an extremely delayed response! I was caught up in something that was a little more urgent.

> I'm glad you said that we're using an interface, the thing is the FileJujuData is not the interface, but the implementation of the interface

Yes, sorry, I didn't mean interface from a language contruct perspective, just that `FileJujuData`'s job is to be the intermediary between code and the file.

I think you've convinced me it's reasonable to classify this an error. I'll add a new error type and we can see how it looks. If you have more comments I'll do my best to iterate more quickly.

juju/client/jujudata.py
DanielArndt: Fixed in https://github.com/juju/python-libjuju/pull/937/commits/747dc771710bd0295c4480c0927d07a144e39b17

tests/unit/test_jujudata.py
```diff
@@ -0,0 +1,16 @@
...
+
+from juju.client.jujudata import FileJujuData
+
+
+class TestJujuData(unittest.IsolatedAsyncioTestCase):
+    @mock.patch("io.open")
+    async def test_verify_controller_uninitialized(self, mock_io_open):
+        mock_io_open.side_effect = FileNotFoundError()
+        jujudata = FileJujuData()
+        assert jujudata.current_controller() is None
```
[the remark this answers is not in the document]
DanielArndt: Fixed in https://github.com/juju/python-libjuju/pull/937/commits/747dc771710bd0295c4480c0927d07a144e39b17

juju/client/connector.py
DanielArndt: Does https://github.com/juju/python-libjuju/pull/937/commits/747dc771710bd0295c4480c0927d07a144e39b17 make more sense?

juju/client/jujudata.py
```diff
@@ -77,7 +79,10 @@ def refresh(self):
 
     def current_controller(self):
         '''Return the current controller name'''
-        return self._load_yaml('controllers.yaml', 'current-controller')
+        try:
+            return self._load_yaml('controllers.yaml', 'current-controller')
+        except FileNotFoundError:
+            raise JujuControllerNotFoundError('No controllers.yaml file found. Did you forget to bootstrap Juju?')
```
cderici, line 85: I'd change the wording here just a little bit, as people often deploy workloads that sets up the environment automatically (e.g. GH actions), so the part that sets up the juju env might fail even though they didn't forget to bootstrap themselves.

```suggestion
            raise JujuControllerNotFoundError('No controllers.yaml file found. python-libjuju requires a bootstrapped Juju controller.')
```

cderici:
@DanielArndt the merge is failing because there's a commit in there that's not signed, can you please `rebase -i` and make sure all the commits are signed? Thanks!

DanielArndt:
@cderici Done.

cderici:
@DanielArndt linter fix needed. Feel free to add me as a collaborator on your fork too if you want me to take care of small things like these 👍

PR #939, opened 2023-08-28
Green CI cleanup for python-libjuju

cderici:
#### Description

There's a couple of known intermittent failures in CI, some excessive warning logs and asyncio outputs. This PR attempts to clear up some of these issues for the new release.

Some major ones of the known CI failures are as follows (will be updated on the go) in the QA steps below. All the tests should be green before this lands.

#### QA Steps

```
tox -e integration --tests/integration/test_model.py::test_add_and_list_storage
```

```
tox -e integration -- tests/integration/test_controller.py::test_destroy_model_by_name
```

```
tox -e integration -- tests/integration/test_controller.py::test_secrets_backend_lifecycle
```

```
tox -e integration -- tests/integration/test_crossmodel.py::test_relate_with_offer
```

```
tox -e integration -- tests/integration/test_model.py::test_deploy_bundle_local_charms
```

```
tox -e integration -- tests/integration/test_model.py::test_deploy_bundle_local_charm_series_manifest
```

```
tox -e integration -- tests/integration/test_connection.py::test_reconnect
```

All CI tests need to pass.

Following examples should complete without any noise:
```sh
 $ python examples/connect_current_model.py
```

Also try to deliberately have it error. The error should only include the relevant parts, i.e. no clutter or noise from pending tasks etc.

```sh
 $ juju deploy ubuntu
 $ python examples/deploy.py
Connecting to model
Deploying ubuntu
Disconnecting from model
  ...........
  File "/home/caner/work/python-libjuju/juju/client/_client17.py", line 1051, in Deploy
    reply = await self.rpc(msg)
  File "/home/caner/work/python-libjuju/juju/client/facade.py", line 659, in rpc
  File "/home/caner/work/python-libjuju/juju/client/connection.py", line 693, in rpc
    raise errors.JujuError(err_results)
juju.errors.JujuError: ['cannot add application "ubuntu": application already exists']
```

#### Notes & Discussion

JUJU-4549

### DIFF
--- a/juju/utils.py
+++ b/juju/utils.py
@@ -128,7 +128,7 @@ async def block_until(*conditions, timeout=None, wait_period=0.5):
     async def _block():
         while not all(c() for c in conditions):
             await jasyncio.sleep(wait_period)
-    await jasyncio.wait_for(_block(), timeout)
+    await jasyncio.shield(jasyncio.wait_for(_block(), timeout))
 
 
 async def block_until_with_coroutine(condition_coroutine, timeout=None, wait_period=0.5):
@@ -139,7 +139,7 @@ async def block_until_with_coroutine(condition_coroutine, timeout=None, wait_per
     async def _block():
         while not await condition_coroutine():
             await jasyncio.sleep(wait_period)
-    await jasyncio.wait_for(_block(), timeout=timeout)
+    await jasyncio.shield(jasyncio.wait_for(_block(), timeout=timeout))
 
 
 async def wait_for_bundle(model, bundle, **kwargs):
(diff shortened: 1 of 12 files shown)


juju/utils.py
```diff
@@ -181,6 +181,10 @@ async def run_with_interrupt(task, *events, log=None):
                                         return_when=jasyncio.FIRST_COMPLETED)
     for f in pending:
         f.cancel()  # cancel unfinished tasks
+        try:
+            await f
```
jameinel, line 185: Shouldn't this be a loop around cancel, and *then* a loop around await? That way everything has time to start shutting down before we wait on them? Otherwise if the first is slow, the second never gets started, etc.

cderici: Yeah that's a better way of doing it. I'm trying to understand why the test keeps getting stuck (only fails on CI, locally it seems to be working). I'm not sure, but it might have something to do with this, actually, I'll see if fixing this will change the test behavior https://github.com/juju/python-libjuju/actions/runs/6029145138/job/16358032056?pr=939

jameinel (approved):

anvial (approved):

PR #940, opened 2023-08-29
Parse nested assumes expressions

### LINKED ISSUE #938: Support nested assumes
### Description

There is an error when using juju lib with a charm that has nested [`assumes`](https://juju.is/docs/sdk/metadata-yaml#heading--assumes) statement.

Currently, the library seems to only [support 1 level of nesting](https://github.com/juju/python-libjuju/blob/3edef16a7e1d2d1fb8b303c71c03a8ed46f022cf/juju/client/facade.py#L683-L695) whereas by [definition](https://github.com/juju/charm/blob/master/assumes/parser.go) the n-depth nesting statements are supported. For example, the [mysql-k8s charm](https://github.com/canonical/mysql-k8s-operator/blob/main/metadata.yaml#L74-L82) uses a nesting of depth 2.

### Urgency

### Python-libjuju version

all versions

### Juju version

all versions

### Reproduce / Test

```python
import pytest
import OpsTest
from juju.model import Model

@pytest.fixture(scope="module")
def model(ops_test: OpsTest) -> Model:
    model = ops_test.model
    assert model
    return model

def test_deploy_nested_assumes(model: Model):
    await model.deploy("mysql-k8s", channel="8.0/edge", trust=True)
    await model.wait_for_idle(status="active", apps=["mysql-k8s"])
```
```

### PR DESCRIPTION
**Pull request by cderici**
#### Description

This adds support for nested `assumes` expression. Without this change, pylibjuju could only parse 1-level of `assumes`, with this change, we can support assumes expression that look like:

```
assumes:
  - juju
  - any-of:
      - all-of:
          - juju >= 2.9
          - juju < 3
      - all-of:
          - juju >= 3.1
          - juju < 4
```

Fixes #938

#### QA Steps

Unfortunately it's not trivial to write an integration test for this, as the code that's been changed is part of the facade api. So we'll use the `mysql-k8s` charm that's been used in #938.

```
 $ juju version
2.9.x
 $ juju bootstrap microk8s micro28
 $ juju add-model assumes-test
```

```python
 $ python -m asyncio
asyncio REPL 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
>>> import asyncio
>>> from juju import model;m=model.Model();await m.connect();await m.deploy("mysql-k8s", channel="8.0/edge", trust=True)
<Application entity_id="mysql-k8s">
>>>
exiting asyncio REPL...
```

#### Notes & Discussion

JUJU-4562

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -661,6 +661,18 @@ class Type:
 
     @classmethod
     def from_json(cls, data):
+        def _parse_nested_list_entry(expr, result_dict):
+            if isinstance(expr, str):
+                cls.splitEntries(expr, result_dict)
+            elif isinstance(expr, dict):
+                for _, v in expr.items():
+                    _parse_nested_list_entry(v, result_dict)
+            elif isinstance(expr, list):
+                for v in expr:
+                    _parse_nested_list_entry(v, result_dict)
+            else:
+                raise TypeError(f"Unexpected type of entry in assumes expression: {expr}")
+
         if isinstance(data, cls):
             return data
         if isinstance(data, str):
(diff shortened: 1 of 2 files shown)


anvial (approved):
lg2m

PR #952, opened 2023-09-21
Optimize connection teardown

cderici:
#### Description

This solves a couple of issues about the connection teardown. The main observation is that we have a rogue connection that we open to the controller during model connection to get some model info but never close. That connection has its own pinger/receiver tasks and an internal websocket object (and that object has its own managing tasks etc), and because this connection is totally forgotten, these tasks etc are not handled properly. For a long time I thought the task errors we were intermittently receiving were about the tasks belonging to the main connection. This fixes it by just using the `ConnectedController` context manager to properly disconnect after used, which wraps up the tasks gracefully.

Another change is that this reduces the websocket close timeout to 1 second (default is 10). With the first change, we essentially won't need this, but in any case, we might wanna experiment with this timeout to make sure we don't get false closes.

#### QA Steps

I used @jameinel 's annotate script to reproduce the issues and investigate this, so I think it would make a good use for QAing this as well:

https://github.com/jameinel/test-annotate/blob/main/annotate.py

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -450,6 +450,7 @@ class Connection:
             self._debug_log_task.cancel()
 
         if self._ws and not self._ws.closed:
+            log.debug('close: calling websocket.close()')
             await self._ws.close()
 
         if not to_reconnect:
@@ -463,6 +464,7 @@ class Connection:
                 pass
             except websockets.exceptions.ConnectionClosed:
                 pass
+        log.debug('close: all tasks are done')
 
         self._pinger_task = None
         self._receiver_task = None
@@ -595,6 +597,7 @@ class Connection:
             raise
 
     async def _pinger(self):
+        log.warning('Pinger: Starting')
         '''
         A Controller can time us out if we are silent for too long. This
         is especially true in JaaS, which has a fairly strict timeout.
@@ -604,6 +607,7 @@ class Connection:
         '''
         async def _do_ping():
             try:
+                log.debug(f'Pinger {self._pinger_task}: pinging')
                 await pinger_facade.Ping()
             except jasyncio.CancelledError:
                 raise
@@ -643,7 +647,7 @@ class Connection:
         if "version" not in msg:
             msg['version'] = self.facades[msg['type']]
         outgoing = json.dumps(msg, indent=2, cls=encoder)
-        log.debug('connection {} -> {}'.format(id(self), outgoing))
+        log.debug('connection id: {} -- sending {}'.format(id(self), outgoing))
         for attempt in range(3):
             if self.monitor.status == Monitor.DISCONNECTED:
                 # closed cleanly; shouldn't try to reconnect
@@ -666,7 +670,7 @@ class Connection:
                     log.error('RPC: Automatic reconnect failed')
                     raise
         result = await self._recv(msg['request-id'])
-        log.debug('connection {} <- {}'.format(id(self), result))
+        log.debug('connection id : {} -- receiving {}'.format(id(self), result))
 
         if not result:
             return result
@@ -797,6 +801,7 @@ class Connection:
             if not self.is_debug_log_connection:
                 self._build_facades(res.get('facades', {}))
                 if not self._pinger_task:
+                    log.debug('reconnect: creating pinger task')
                     self._pinger_task = jasyncio.create_task(self._pinger(), name="Task_Pinger")
 
     async def _connect(self, endpoints):
@@ -853,6 +858,7 @@ class Connection:
         #  If this is regular connection, and we dont have a
         #  receiver_task yet, then schedule a _receiver_task
         elif not self.is_debug_log_connection and not self._receiver_task:
+            log.debug('_connect: creating receiver task')
             self._receiver_task = jasyncio.create_task(self._receiver(), name="Task_Receiver")
 
         log.debug("Driver connected to juju %s", self.addr)
@@ -908,6 +914,7 @@ class Connection:
             login_result = await self._connect_with_login(e.endpoints)
         self._build_facades(login_result.get('facades', {}))
         if not self._pinger_task:
+            log.debug('_connect_with_redirect: creating pinger task')
             self._pinger_task = jasyncio.create_task(self._pinger(), name="Task_Pinger")
 
     # _build_facades takes the facade list that comes from the connection with the controller,
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -433,6 +433,7 @@ class Connection:
             max_size=self.max_frame_size,
             server_hostname=server_hostname,
             sock=sock,
+            close_timeout=1,
         )), url, endpoint, cacert
 
     async def close(self, to_reconnect=False):
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -454,7 +454,6 @@ class Connection:
             self._debug_log_task.cancel()
 
         if self._ws and not self._ws.closed:
-            log.debug('close: calling websocket.close()')
             await self._ws.close()
 
         if not to_reconnect:
@@ -465,7 +464,6 @@ class Connection:
                 pass
             except websockets.exceptions.ConnectionClosed:
                 pass
-        log.debug('close: all tasks are done')
 
         self._pinger_task = None
         self._receiver_task = None
@@ -598,7 +596,6 @@ class Connection:
             raise
 
     async def _pinger(self):
-        log.warning('Pinger: Starting')
         '''
         A Controller can time us out if we are silent for too long. This
         is especially true in JaaS, which has a fairly strict timeout.
@@ -802,7 +799,7 @@ class Connection:
             if not self.is_debug_log_connection:
                 self._build_facades(res.get('facades', {}))
                 if not self._pinger_task:
-                    log.debug('reconnect: creating pinger task')
+                    log.debug('reconnect: scheduling a pinger task')
                     self._pinger_task = jasyncio.create_task(self._pinger(), name="Task_Pinger")
 
     async def _connect(self, endpoints):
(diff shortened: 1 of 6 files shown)


jameinel:
So I did a quick prototype with just 2 changes in it:
```
-                tasks_need_to_be_gathered = [task for task in jasyncio.all_tasks() if task != jasyncio.current_task()]
+                tasks_need_to_be_gathered = [self._receiver_task, self._pinger_task]
```
and
```
            async with ConnectedController(self.connection()) as contr:
                self._info = await contr.get_model_info(model_name, model_uuid)
                log.debug('Got ModelInfo: %s', vars(self.info)
```

that made:
```
$ time ./annotate.py get
baz: 'bing'

real    0m0.474s
```

so really I think the fix is just that the second connection needs to be closed, and then everything works like we want.

I don't know if we want some of the other logging, etc changes. mostly I don't think we need them. But that bit is a huge win.

juju/client/connection.py
```diff
@@ -433,6 +433,7 @@ def _exit_tasks():
             max_size=self.max_frame_size,
             server_hostname=server_hostname,
             sock=sock,
+            close_timeout=1,
```
jameinel, line 436: I actually don't think we want to change this. If it is something that we want to expose to clients we can, but I'd rather use default timeouts unless we have a real need to change it. (This seems sane, except when you might have poor connectivity, etc.)
Since we know we don't need to change it when things are 'happy' (in my testing I can do the annotation round trip in 400ms and exit success).

juju/client/connection.py
```diff
@@ -642,7 +645,7 @@ async def rpc(self, msg, encoder=None):
         if "version" not in msg:
             msg['version'] = self.facades[msg['type']]
         outgoing = json.dumps(msg, indent=2, cls=encoder)
-        log.debug('connection {} -> {}'.format(id(self), outgoing))
+        log.debug('connection id: {} -- sending {}'.format(id(self), outgoing))
```
jameinel, line 648: I actually prefer the "->" vs "<-" to 'sending' and 'receiving' because it means the lines line up in debug output and the arrows aren't too hard to interpret.
we could go with 'send' and 'recv', I suppose

juju/client/connector.py
```diff
@@ -73,6 +73,12 @@ async def connect(self, **kwargs):
             assert self._connection
             self._log_connection = await Connection.connect(**kwargs)
         else:
+            # TODO (cderici): we need to investigate how to reuse/share
+            # connections between Model & Controller objects
+            # At least either avoid overwriting self._connection with
+            # different types of connections (model, controller), or
+            # have an indication of which type of entity this is
+            # connected to.
```
jameinel, line 81: Do we want to do something like refcounted connections that only close when the last reference goes away?

Does the comment about "have an indication" still apply, given that you changed disconnect() to take `entity`?

juju/model.py
```diff
@@ -772,9 +772,13 @@ async def watch_received_waiter():
             raise self._watcher_task.exception()
 
         if self.info is None:
```
jameinel, line 774: It seems odd to me that we check "self.info is None" but we then set "self._info". I realize that info is a property that just returns self._info, but it seems much better to check the same variable that you are setting. (it becomes much clearer that you are using a cache variable)

Otherwise I would put the caching into the info property itself, so that only when we need self.info do we write to self._info.
And then this code becomes:
```
if self.uuid is None:
  self.uuid = self.info.uuid
```

and the property becomes:
```
@property
def info(self):
    if self._info is None:
            async with ConnectedController(self.connection()) as contr:
                self._info = await contr.get_model_info(model_name, model_uuid)
                log.debug('Got ModelInfo: %s', vars(self.info))
   return self._info
```

Then we cache whenever anyone accesses self.info, but not until then, and we don't need it if we already know the uuid.

Thoughts?

juju/controller.py
```diff
@@ -193,7 +193,7 @@ async def disconnect(self):
         """Shut down the watcher task and close websockets.
 
         """
-        await self._connector.disconnect()
+        await self._connector.disconnect(entity='controller')
```
jameinel, line 196: Here you use 'controller', but later you use 'Model'. We should at least be consistent about the capitalization

juju/client/connection.py
cderici: Yeah I agree about this, let's not change it if we don't have to.

juju/client/connection.py
cderici: Yeah, I think the `->` and `<-` *looks* better, but I got frustrated when I was manually reading the debug output that I was misinterpreting some of them, that's why I thought the change might make it clearer. But I haven't considered that the lines line up in the output, so *that* might be a good reason to revert this 👍

cderici: I think I'll make them `--->` and `<---`, which might make it harder to misinterpret, and would still line up

juju/client/connector.py
cderici: > Does the comment about "have an indication" still apply, given that you changed disconnect() to take `entity`?

No, because you might wanna discover the kind of connection in other places too, depends on what you wanna do, for example we'd get different facades from a model connection then a controller connection. 

So potentially I'd have `self.model_connection` and `self.controller_connection` separately, where we could have a better control which would enable us to avoid doing awkward things like `Connector.connect_model(_model_name)` [first connects to the controller](https://github.com/juju/python-libjuju/blob/54c49e2b704fb67a5881d3d18a935066f78a730a/juju/client/connector.py#L152) to get some info like model uuid etc, [then throws away its connection](https://github.com/juju/python-libjuju/blob/54c49e2b704fb67a5881d3d18a935066f78a730a/juju/client/connector.py#L77) to connect to the model.

> Do we want to do something like refcounted connections that only close when the last reference goes away?

I'd want to move on the opposite direction, actually, to cache the connections and keep it alive in a nice way as long as the process is still going on (and the entity is still alive on the other end). Because a lot of pylibjuju users are still doing the practice in their scripts where they do "connect -> do something -> disconnect -> connect again -> do something else -> disconnect ...". Another solution to this would be to expose nice contexts ([we actually use some internally](https://github.com/juju/python-libjuju/blob/54c49e2b704fb67a5881d3d18a935066f78a730a/tests/base.py#L66)) where they could do:

```python
with model.Model() as m:
  await m.deploy()
```

rather than the current way of doing it like:

```python
m = model.Model()
await m.connect()
await m.deploy()
await m.disconnect()
```

juju/model.py
cderici: Yeah I agree, but I'd probably go with just checking the `self._info`, because caching can't be done in the property method (you can't do async stuff in a non-async method).

juju/controller.py
cderici: Oh yeah good catch, thanks

juju/model.py
```diff
@@ -771,10 +771,14 @@ async def watch_received_waiter():
...
-            contr = await self.get_controller()
-            self._info = await contr.get_model_info(model_name, model_uuid)
-            log.debug('Got ModelInfo: %s', vars(self.info))
+        if self._info is None:
+            # TODO (cderici): See if this can be optimized away, or at least
+            # be done lazily (i.e. not everytime after_connect, but whenever
+            # self.info is needed -- which here can be bypassed if model_uuid
+            # is known)
+            async with ConnectedController(self.connection()) as contr:
+                self._info = await contr.get_model_info(model_name, model_uuid)
```
jameinel, line 780: https://pypi.org/project/async-property/

Supposedly has things like:
`@async_cached_property`
Which will call the function only once and otherwise cache the result.

That said, I don't think async and properties act nice together, because then you need to 'foo = await obj.foo' which really doesn't feel good (I generally don't think accessing a property should trigger an HTTP call.)

Even if their doc, this doesn't feel good:
```
>>> instance.value
<AwaitableOnly "Foo.value">

>>> await instance.value
loading value
123
>>> await instance.value
123
>>> instance.value
123
```
The fact that you don't always have to await, but sometimes you do is really not a great UX.

So I was wrong to ask you to do so as part of the property, and checking _info is ok.

juju/client/connector.py
jameinel: I think having a pattern like this could be ok:
```
active_connection = ...
  with model.Model(active_connection, name) as m:
    do stuff

  with model.Model(active_connection, other_name) as m:
    do other stuff
```
or
```
   with active_connection.Model() as m:
     ...
```

But I wouldn't want to have a global (module/package level) cache of active connections that get reused behind your back. (what happens when one breaks, and who was using it, how do you force a reconnection, etc.)

tests/integration/test_model.py
```diff
@@ -223,8 +223,6 @@ async def test_deploy_local_charm_channel(event_loop):
     async with base.CleanModel() as model:
         await model.deploy(str(charm_path), channel='stable')
         assert 'charm' in model.applications
-        await model.wait_for_idle(status="active")
-        assert model.units['charm/0'].workload_status == 'active'
```
jameinel, line 226: shouldn't we be waiting for the unit to be idle? Or is this the bugfix you did that said "a model isn't idle until all units in the model are idle' ?

I believe that I'm ok with all the changes, except this one.

juju/client/connection.py
```diff
@@ -603,6 +604,7 @@ async def _pinger(self):
         '''
         async def _do_ping():
             try:
+                log.debug(f'Pinger {self._pinger_task}: pinging')
```
jameinel, line 607: I have the feeling that this is too verbose even at Debug level.

cderici:
> But I wouldn't want to have a global (module/package level) cache of active connections that get reused behind your back. (what happens when one breaks, and who was using it, how do you force a reconnection, etc.)

@jameinel That's exactly what happens right now and what I'm trying to fix. The  `_connection` within a `Connector` is kept being overwritten and repurposed all the time internally (this excerpt is from  `Connector.connect()` https://github.com/juju/python-libjuju/blob/48570bb8d51d38c430d12b86e39706ffd6969fcc/juju/client/connector.py#L76-L77)

jameinel (approved):

PR #949, opened 2023-09-12
Use new DeployFromRepository endpoint for deploy

cderici:
#### Description

This allows pylibjuju to use the new `DeployFromRepository` endpoint, introduced in the `Application` facade `v19`.

`PendingUploadResources` will be handled in a separate PR.

This also fixes what https://github.com/juju/juju/commit/22e3fc0ef46b718005c25b9dec6c3e2bfdee6204#diff-cd1b6b10813a1b0ebe7fe9a04f11c401dfeec0574ffb00058bfec98b6bb1d255 seems to be breaking by returning a non-empty error list `(Pdb) self.plan.errors
['', '', '', '', '', '', '', '', '']`  from `GetChangesMapArgs`.

#### QA Steps

There's two parts for QAing this. Making sure deploy still works for everything `< 3.3`. CI tests will cover most of that part of the QA, as we use deploy in almost every single integration test, and it's being tested against `latest/stable` (which as of today is `3.1.5`). So it is advisable to do the manual QA (see below) against juju `3.2`, just to be sure.

Second part is obviously making sure that the deploy works for `3.3` with the new endpoint. To do that, I simply 

```sh
 $ juju version
3.3-beta2-ubuntu-amd64
 $ juju bootstrap localhost lxd33 && juju add-model test
```

Then just manually deployed the `ubuntu` charm using repl.

```python
python -m asyncio
asyncio REPL 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from juju import model;m=model.Model();await m.connect();await m.deploy('ubuntu')
<Application entity_id="ubuntu">
>>>
exiting asyncio REPL...
```

All CI tests need to pass.

#### Notes & Discussion

JUJU-3637

PR #953, opened 2023-09-22
Handle pending upload resources deployfromrepository

cderici:
#### Description

This is the continuation of https://github.com/juju/python-libjuju/pull/949, that implements handling of the local resources that need to be uploaded after a (server-side) deploy.

In particular, this splits out the second part of the `add_local_resources` into a separate `_upload` function to use after the `DeployFromRepository` call which reports the pending file uploads if there's any.

#### QA Steps

Following the https://github.com/juju/python-libjuju/pull/949, this needs the server side deploy support from the controller (i.e. `<= 3.3`). So, 

```sh
 $ juju version
3.3-beta2-ubuntu-amd64
 $ juju bootstrap localhost lxd33 && juju add-model test
```

Now let's make a local resource to use:

```sh
 $ cd python-libjuju
 $ echo "jujurulez" > ./foo.txt
 $ cat ./foo.txt
jujurulez
```

Then just manually deploy the `juju-qa-test` charm with the local resource `foo.txt`.

```python
python -m asyncio
asyncio REPL 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
>>> import asyncio
>>> from juju import model;m=model.Model();await m.connect();await m.deploy('juju-qa-test', application_name='j1', resources={'foo-file':'./foo.txt'}))
<Application entity_id="j1">
>>>
exiting asyncio REPL...
```

Now confirm that the resource is uploaded:

```sh
 $ juju resources j1
Resource  Supplied by  Revision
foo-file  admin        2023-09-19T22:55
```

All CI tests need to pass.

#### Notes & Discussion

JUJU-3638

Note that, this is targeting the `origin:deploy-from-repository-new-endpoint` branch (of my own fork), which will be merged to `juju/python-libjuju:master`. After the merge, this will be pointing to there.


cderici:
Failing test is a timeout, unrelated to the PR, should be fixed by #952

jack-w-shaw (approved):
LGTM

Issue #963, opened 2023-10-13
python-libjuju cannot use `app.refresh(switch=...)` to switch to a local charm


jack-w-shaw:
### Description

This is a capability of the juju cli

### Urgency

Casually reporting

### Python-libjuju version

all

### Juju version

all

### Reproduce / Test

```python
$ juju deploy ubuntu
$ cat test.py
from juju import jasyncio
from juju.model import Model

async def main():
    model = Model()
    print('Connecting to model')
    await model.connect()

    try:
        app = model.applications["ubuntu"]
        print("refreshing", app)
        await app.refresh(switch="local:/home/jack/charms/ubuntu")

    finally:
        print('Disconnecting from model')
        await model.disconnect()

if __name__ == '__main__':
    jasyncio.run(main())

$ python test.py
Connecting to model
refreshing <Application entity_id="ubuntu">
ERROR:root:Error during request: https://api.jujucharms.com/charmstore/v5/local:/home/jack/charms/ubuntu/meta/any?include=bundle-machine-count&include=bundle-metadata&include=bundle-unit-count&include=charm-actions&include=charm-config&include=charm-metadata&include=common-info&include=extra-info&include=owner&include=published&include=resources&include=supported-series&include=terms&include=id&include=stats status code:(400) message: {"Code":"bad request","Message":"Charm ID not recognized"}

ERROR:root:Error during request: https://api.jujucharms.com/charmstore/v5/local:/home/jack/charms/ubuntu/meta/any?include=bundle-machine-count&include=bundle-metadata&include=bundle-unit-count&include=charm-actions&include=charm-config&include=charm-metadata&include=common-info&include=extra-info&include=owner&include=published&include=resources&include=supported-series&include=terms&include=id&include=stats status code:(400) message: {"Code":"bad request","Message":"Charm ID not recognized"}

ERROR:root:Error during request: https://api.jujucharms.com/charmstore/v5/local:/home/jack/charms/ubuntu/meta/any?include=bundle-machine-count&include=bundle-metadata&include=bundle-unit-count&include=charm-actions&include=charm-config&include=charm-metadata&include=common-info&include=extra-info&include=owner&include=published&include=resources&include=supported-series&include=terms&include=id&include=stats status code:(400) message: {"Code":"bad request","Message":"Charm ID not recognized"}

Disconnecting from model
Traceback (most recent call last):
  File "/home/jack/.local/lib/python3.10/site-packages/theblues/charmstore.py", line 78, in _get
[...]

PR #971, opened 2023-10-24
Allow switch kwarg in refresh to switch to local charms

jack-w-shaw:
Port forward https://github.com/juju/python-libjuju/pull/968

Check if the switch param is pointing to a local charm. If it is, run local_refresh as if it were provided

This is how the Juju CLI works, which was missing from pylibjuju. Replicate this.

Also push url parsing of switch kwarg behind a check if it's local charm, meaning we avoid the possibility where we attempt to parse a path as if it were a url

Resolves https://github.com/juju/python-libjuju/issues/963

### QA Steps

```
tox -e integration -- tests/integration/test_application.py::test_refresh_charmhub_to_local
```

All CI tests need to pass.

Verify the following example script runs successfully:
```
from juju import jasyncio
from juju.model import Model


async def main():
    model = Model()
    print('Connecting to model')
    await model.connect()
    try:
        print('path="/home/jack/charms/ubuntu"')
        await depl(model, path="/home/jack/charms/ubuntu")
        print('switch="/home/jack/charms/ubuntu"')
        await depl(model, switch="/home/jack/charms/ubuntu")
        print('switch="local:/home/jack/charms/ubuntu"')
        await depl(model, switch="local:/home/jack/charms/ubuntu")
    finally:
        print('Disconnecting from model')
        await model.disconnect()

async def depl(model, **kwargs):
    try:
        app = await model.deploy("ubuntu")
        await model.block_until(lambda: all(u[0].workload_status == 'active' for u in app.units))
        await app.refresh(**kwargs)
    finally:
        await app.remove()
        await model.block_until(lambda: not len(model.applications))

if __name__ == '__main__':
    jasyncio.run(main())
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -4,7 +4,7 @@
 import hashlib
 import json
 import logging
-import pathlib
+from pathlib import Path
 
 from . import model, tag, utils, jasyncio
 from .url import URL
@@ -12,7 +12,7 @@ from .status import derive_status
 from .annotationhelper import _get_annotations, _set_annotations
 from .client import client
 from .errors import JujuError, JujuApplicationConfigError
-from .bundle import get_charm_series
+from .bundle import get_charm_series, is_local_charm
 from .placement import parse as parse_placement
 from .origin import Channel, Source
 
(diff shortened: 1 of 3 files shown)


cderici (approved):
LGTM 👍

jack-w-shaw:
NOTE: The uni integration test failure is a timeout, which passes when I try locally. Merging anyway

Issue #961, opened 2023-10-13
Local charm urls are not parsed correctly


jack-w-shaw:
### Description

The way we parse local charm urls does not line up with Juju itself.

Often we deploy local charms with a charm-url-like string such as `local:/path/to/charm`. Accordingly pylibjuju parses this charm like {schema: "local", name: "/path/to/charm"}

However, this is not actually a valid charm url. See:
https://go.dev/play/p/PWib_PWD0KB?v=goprev

A real local charm url looks something like `local:focal/ubuntu-12` i.e. the same as charmhub charm urls. Sometimes pylibjuju recieves a local charm url like this, in the response to a facade call (charm upload, refresh, etc.) or to deploy a pre-deployed local charm.

Pylibjuju does not parse these correctly, so it is impossible to, say, parse the revision or series from this.

I'm guessing just out of coincidence we don't need to do so? But this could lead to difficult to track down bugs

### Urgency

Casually reporting

### Python-libjuju version

all

### Juju version

all

### Reproduce / Test

```python
from juju.url import URL
u = URL.parse("local:focal/ubuntu-12")
print(u.name)
print(u.revision)
```

PR #974, opened 2023-10-25
Parse charm URLs consistantly for local charms

jack-w-shaw:
Port forward https://github.com/juju/python-libjuju/pull/969

Parse charm URLs consistantly for local charms

Local charm URLs are confused in pylibjuju because often a charm
url-like string is passed into deploy, to explicitly specify a local
charm. These 'urls' were of the form 'local:/path/to/charm'

Local URLs were parsed accordingly.

However, the above is in no sense a url really so should be treated as
such. This also means that pylibjuju is unable to parse real local charm
urls returned to the client from the server when, say, uploading a local
charm.

Drop support for local charms like this. Instead parse local charm urls the
same way we parse charmhub charm url.

This has the side-effect, however, that we can no longer treat all
objects to deploy as URLs. Now, they are either URLs (ch or cs charms)
or paths or paths prepended with 'local:' (local charms)

As such, in deploy code and refresh code, we often need to branch on
is_local_charm. This is to ensure we don't attempt to parse paths as
URLs. Everywhere we URL.parse user input, we ensure the entity is not a
path

This resolves https://github.com/juju/python-libjuju/issues/961

#### QA Steps

```
tox -e unit -- tests/unit/test_url.py
```

```
tox -e integration -- tests/integration/test_model.py::test_deploy_local_bundle_dir
tox -e integration -- tests/integration/test_model.py::test_deploy_local_bundle_file
tox -e integration -- tests/integration/test_model.py::test_deploy_bundle_local_charms
tox -e integration -- tests/integration/test_model.py::test_deploy_local_charm
```

All CI tests need to pass.

(In a python interpreter)
```
>>> from juju.url import URL
>>> u = URL.parse("local:focal/ubuntu-12")
>>> print(u.name)
ubuntu
>>> print(u.revision)
12
>>> URL.parse("local:/path/to/charm")
  File "/home/jack/juju/python-libjuju/juju/url.py", line 50, in parse
    c = parse_v2_url(u, s, default_store)
  File "/home/jack/juju/python-libjuju/juju/url.py", line 138, in parse_v2_url
    raise JujuError("charm or bundle URL {} malformed".format(s))
juju.errors.JujuError: charm or bundle URL local:/path/to/charm malformed
```

Ensure the example scripts still function
```
$ python examples/deploy_bundle.py
$ python examples/deploy_local_bundle_with_resources.py
```

Verify the following example script runs successfully:
```
from juju import jasyncio
from juju.model import Model


async def main():
    model = Model()
    print('Connecting to model')
    await model.connect()
    try:
        print('path="/home/jack/charms/ubuntu"')
        await depl(model, path="/home/jack/charms/ubuntu")
        print('switch="/home/jack/charms/ubuntu"')
        await depl(model, switch="/home/jack/charms/ubuntu")
        print('switch="local:/home/jack/charms/ubuntu"')
        await depl(model, switch="local:/home/jack/charms/ubuntu")
    finally:
        print('Disconnecting from model')
        await model.disconnect()

async def depl(model, **kwargs):
    try:
        app = await model.deploy("ubuntu")
        await model.block_until(lambda: all(u[0].workload_status == 'active' for u in app.units))
        await app.refresh(**kwargs)
    finally:
        await app.remove()
        await model.block_until(lambda: not len(model.applications))

if __name__ == '__main__':
    jasyncio.run(main())
```

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -450,7 +450,7 @@ class LocalDeployType:
     """LocalDeployType deals with local only deployments.
     """
 
-    async def resolve(self, url, architecture,
+    async def resolve(self, charm_path, architecture,
                       app_name=None, channel=None, series=None,
                       revision=None, entity_url=None, force=False,
                       model_conf=None):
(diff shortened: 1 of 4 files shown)


juju/model.py
```diff
@@ -541,7 +532,7 @@ async def resolve(self, url, architecture,
             raise JujuError('revision and channel are mutually exclusive when deploying a bundle. Please choose one.')
 
         if app_name is None:
-            app_name = url.name
+            app_name = charm_url.name
```
cderici, line 535: Looks like the `charm_url` is returned as a string and blowing up here, and needs to be fixed before further review. See https://github.com/juju/python-libjuju/actions/runs/6642890817/job/18048743466?pr=974#step:6:1553

cderici (changes requested):

juju/model.py
jack-w-shaw: Good spot. Thanks, this is a fairly easy fix, _resolve_charm is only called a few times, and when we do we don't assume it's a string mostly

jack-w-shaw: So I've decided to change it to return a charm url

cderici (approved):
CI fail is unrelated, this can land 👍

cderici:
Might need a rebase

PR #975, opened 2023-10-26
Fix juju config directory location pointed by XDG_DATA_HOME

cderici:
#### Description

XDG_DATA_HOME is supposed to point to the parent directory for applications. We were incorrectly setting that as Juju config dir.

#### QA Steps

No operational logic changes. All CI tests need to pass.

#### Notes & Discussions

Thanks @jameinel for pointing this out!

### DIFF
--- a/juju/utils.py
+++ b/juju/utils.py
@@ -46,19 +46,17 @@ def juju_config_dir():
 
     """
     # Check $JUJU_DATA first
-    config_dir = os.environ.get('JUJU_DATA', None)
+    config_dir = Path(os.environ.get('JUJU_DATA', None))
 
     # Second option: $XDG_DATA_HOME for ~/.local/share
-    if not config_dir:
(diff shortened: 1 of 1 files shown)


juju/utils.py
```diff
@@ -50,7 +50,9 @@ def juju_config_dir():
 
     # Second option: $XDG_DATA_HOME for ~/.local/share
     if not config_dir:
-        config_dir = os.environ.get('XDG_DATA_HOME', None)
+        base_dir = os.environ.get('XDG_DATA_HOME', None)
+        if base_dir is not None:
+            config_dir = base_dir + '/juju'
```
jack-w-shaw, line 55: What about the case `$XDG_DATA_HOME` has a trailing `/`?

jack-w-shaw (approved):
LGTM

Issue #983, opened 2023-10-31
Juju connect fails because it can't find controllers.yaml


gboutry:
### Description

Controller.connect(...) fails because it can't find controllers.yaml.

This stacktrace is coming from the inside of the sunbeam snap.

```
[Errno 2] No such file or directory: '/home/ubuntu/snap/openstack/x1/.local/share/juju/controllers.yaml'                                                                                                                 utils.py:282
Traceback (most recent call last):                                                                                                                                                                                                   
    File "/snap/openstack/x1/lib/python3.10/site-packages/sunbeam/utils.py", line 276, in __call__                                                                                                                                     
    return self.main(*args, **kwargs)                                                                                                                                                                                                
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/core.py", line 1055, in main                                                                                                                                           
    rv = self.invoke(ctx)                                                                                                                                                                                                            
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/core.py", line 1657, in invoke                                                                                                                                         
    return _process_result(sub_ctx.command.invoke(sub_ctx))                                                                                                                                                                          
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/core.py", line 1635, in invoke                                                                                                                                         
    rv = super().invoke(ctx)                                                                                                                                                                                                         
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/core.py", line 1404, in invoke                                                                                                                                         
    return ctx.invoke(self.callback, **ctx.params)                                                                                                                                                                                   
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/core.py", line 760, in invoke                                                                                                                                          
    return __callback(*args, **kwargs)                                                                                                                                                                                               
    File "/snap/openstack/x1/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func                                                                                                                                   
    return f(get_current_context(), *args, **kwargs)                                                                                                                                                                                 
    File "/snap/openstack/x1/lib/python3.10/site-packages/sunbeam/commands/configure.py", line 818, in configure                                                                                                                       
    _configure(openrc, preseed, accept_defaults)                                                                                                                                                                                     
    File "/snap/openstack/x1/lib/python3.10/site-packages/sunbeam/commands/configure.py", line 745, in _configure                                                                                                                      
    run_sync(jhelper.get_model(OPENSTACK_MODEL))                                                                                                                                                                                     
[...]

PR #984, opened 2023-11-01
Free pyblijuju from relying on juju client when connecting to a controller

cderici:
#### Description

This moves the computation of the `controller_name` from the `connector` to the `Controller` class, to be lazily evaluated from `jujudata` (and cached). We can move it out of the connector safely because:

1 - `controller_name` is not used within the connector.
2 - the only location that the `controller_name` is returned to, ignores it.

This has the nice consequence of freeing pylibjuju from relying on things like the `controllers.yaml` file (in turn the juju client to be installed) to connect to a controller.

Fixes https://github.com/juju/python-libjuju/issues/983

#### QA Steps

Added a test for lazy computation of `controller_name`. 

```sh
 $ tox -e integration -- tests/integration/test_controller.py::test_connection_lazy_jujudata
```

But the scenario from the #983 can be tried easily as follows:

```python
# make a new controller and connect it in a normal way:
from juju import controller
c = controller.Controller()
await c.connect()

# get the connection
conn = c.connection()
```

Now go move the `.local/share/juju/controllers.yaml` file temporarily.

```python
# make a new controller
c2 = controller.Controller()

# connect it using the information within the connection we got above
await c2.connect(endpoint=conn.endpoints[0][0], cacert=conn.cacert, username=conn.usertag, password=conn.password)
```

Should succeed without any errors.

Don't forget to move back your `controllers.yaml` back so nothing else on your machine freaks out.

All CI tests need to pass.

#### Notes & Discussion

juju-4891

### DIFF
--- a/juju/client/connector.py
+++ b/juju/client/connector.py
@@ -9,7 +9,7 @@ from juju.client.connection import Connection
 from juju.client.gocookies import GoCookieJar, go_to_py_cookie
 from juju.client.jujudata import FileJujuData, API_ENDPOINTS_KEY
 from juju.client.proxy.factory import proxy_from_config
-from juju.errors import JujuConnectionError, JujuError, PylibjujuProgrammingError
+from juju.errors import JujuConnectionError, JujuError
 from juju.client import client
 from juju.version import SUPPORTED_MAJOR_VERSION, TARGET_JUJU_VERSION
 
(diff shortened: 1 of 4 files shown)


jack-w-shaw (approved):
LGTM

cderici:
CI fails are unrelated, @jack-w-shaw if you wanna tackle a bitesize thing, only a charm change would fix [this one](https://github.com/juju/python-libjuju/actions/runs/6725820058/job/18712236551?pr=984#step:6:392).

PR #986, opened 2023-11-29
Add support for adding user secrets

cderici:
#### Description

This is the first PR in exposing user secrets for python-libjuju. I decided to split this work into multiple PRs (rest is/are coming right after this) because this PR was getting really big with the updated schema and the clients etc.

The main changes are as follows:

- Adds a new `secrets.py` module that contains the client logic for producing the data bag for secrets to be added (and updated), including reading values from yaml/json files etc.
- Updates the schema with the latest 3.3 schema from juju and re-generates the clients to get the `Secrets` v2 facade.
- Adds the `model.add_secret()` function that corresponds to `juju add-secret` cli command.

#### QA Steps

Some unit tests are added for the functions in the new `secrets.py` module.

```
tox -e py3 -- tests/unit/test_secrets.py::TestCreateSecretData
```

```
tox -e py3 -- tests/unit/test_secrets.py::TestReadSecretData
```

Also added an integration test for `add_secret`:

```
tox -e integration -- tests/integration/test_secrets.py::test_add_secret
```

This can also be tested manually by using Python repl:

```python
 $ python -m asyncio 
>>> from juju import model
>>> m=model.Model();await m.connect()
>>> await model.add_secret(name='my-apitoken', dataArgs=['token=34ae35facd4'])
secret:........
```

After this, check the `juju secrets`, you should see the newly added secret.

#### Notes & Discussion

JUJU-4926


cderici:
The integration tests in the gh action is running against 3.2 (which means no user secrets support in the controller), the GH workflow needs to be updated (I'll do that first thing tomorrow 👍).

anvial (approved):
I'm with the code.

So, We need only to fix the Integration tests.

PR #990, opened 2023-12-04
Pass websocket Close frame into ConnectionClosed

### LINKED ISSUE #989: (Bug in client/connection.py) "raise websockets.exceptions.ConnectionClosed" needs a 'Close' object, not the code and reason directly.
As per https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosed The parameters need to be 'Close' object, not a number code and reason string as seen on line 483.

https://github.com/juju/python-libjuju/blob/2581b0ced1df6201c6b7fd8cc0b20dcfa9d97c51/juju/client/connection.py#L483

cderici:
#### Description

This fixes a bug in connection where we pass close code and reason directly into the `websockets.exception.ConnectionClosed`, where it needs a Close frame that contains those.

Fixes #989

#### QA Steps

This doesn't affect any normal behavior, as this is sort of a safeguard against races. The only two times we raise this explicit exception is when the `MONITOR` is closed but either the receiver task is running still or someone makes an rpc call, which shouldn't happen anyways. So manually checking the arguments by the [doc](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosed) is sufficient.

All CI tests need to pass.

#### Notes & Discussion

Needs to be forward ported into 3.x.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -436,7 +436,9 @@ class Connection:
 
     async def _recv(self, request_id):
         if not self.is_open:
-            raise websockets.exceptions.ConnectionClosed(0, 'websocket closed')
+            raise websockets.exceptions.ConnectionClosed(
+                websockets.frames.Close(websockets.frames.CloseCode.NORMAL_CLOSURE,
+                                        'websocket closed'))
         try:
             return await self.messages.get(request_id)
         except GeneratorExit:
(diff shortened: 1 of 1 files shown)


jack-w-shaw (approved):
LGTM

PR #993, opened 2023-12-14
Drop use of walrus operator

freyes:
#### Description

The walrus operator (`:=`) was introduced in Python 3.8, the use of it in the `2.9` branch breaks python 3.6 compatibility which is the version used in Ubuntu 18.04 (Bionic).

https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions

#### QA Steps

1. Install python-libjuju in python 3.6
2. python3 -c "import juju.utils"

```
virtualenv -p /usr/bin/python3.6 /tmp/myvenv
source /tmp/myvenv/bin/activate
python3 setup.py install
python3 -c "import juju.utils"
```

All CI tests need to pass.

#### Notes & Discussion

This regression was introduced by https://github.com/juju/python-libjuju/pull/975

### DIFF
--- a/juju/utils.py
+++ b/juju/utils.py
@@ -49,11 +49,11 @@ def juju_config_dir():
     config_dir = Path('~/.local/share/juju')
 
     # Check $JUJU_DATA
-    if juju_data := os.environ.get('JUJU_DATA'):
-        config_dir = Path(juju_data)
+    if os.environ.get('JUJU_DATA'):
+        config_dir = Path(os.environ.get('JUJU_DATA'))
     # Secondly check: $XDG_DATA_HOME for ~/.local/share
-    elif xdg_data_home := os.environ.get('XDG_DATA_HOME'):
-        config_dir = Path(xdg_data_home) / 'juju'
+    elif os.environ.get('XDG_DATA_HOME'):
+        config_dir = Path(os.environ.get('XDG_DATA_HOME')) / 'juju'
 
     return str(config_dir.expanduser().resolve())
 


cderici:
If this is the only breaking point for 3.6, I won't be against removing this particular use of the walrus, just to unblock you. 

However, it's important to note that we dropped support for python < 3.8 a long while ago, and will soon be moving to support 3.11-3.12 and 3.13. So you should expect more things to fail if you continue using python < 3.8.

juju/utils.py
```diff
@@ -49,10 +49,12 @@ def juju_config_dir():
     config_dir = Path('~/.local/share/juju')
 
     # Check $JUJU_DATA
-    if juju_data := os.environ.get('JUJU_DATA'):
-        config_dir = Path(juju_data)
+    juju_data = os.environ.get('JUJU_DATA')
     # Secondly check: $XDG_DATA_HOME for ~/.local/share
-    elif xdg_data_home := os.environ.get('XDG_DATA_HOME'):
+    xdg_data_home = os.environ.get('XDG_DATA_HOME')
```
cderici, line 54: No need to get this if the `juju_data` is available. Please put it in the else case down below.

cderici (changes requested):

freyes:
> If this is the only breaking point for 3.6, I won't be against removing this particular use of the walrus, just to unblock you.
> 
> However, it's important to note that we dropped support for python < 3.8 a long while ago, and will soon be moving to support 3.11-3.12 and 3.13. So you should expect more things to fail if you continue using python < 3.8.

I understand the desire (and need) to drop support for old versions of python, although this is the `2.9` branch, where the expectation to support Ubuntu 18.04 sounds reasonable to me.

juju/utils.py
freyes: ack. Addressed by https://github.com/juju/python-libjuju/pull/993/commits/f05162e3d361d36b297afa3900e081bef1774b95

cderici (approved):
LGTM 👍

cderici:
@freyes All the commits need to be signed to be able to land this, please rebase with `git config --global commit.gpgSign true`

freyes:
> @freyes All the commits need to be signed to be able to land this, please rebase with `git config --global commit.gpgSign true`

since I will need to do a "push --force", do you mind if I squash those two commits into a single one?, it would make history cleaner.

cderici (approved):
> since I will need to do a "push --force", do you mind if I squash those two commits into a single one?, it would make history cleaner.

I'd even prefer it that way, thanks a lot for taking care of this, this is good to go 👍

PR #1005, opened 2024-01-09
Remove paramiko upper-bound

gboutry:
#### Description

This change removes paramiko upper bound, allow using paramiko >3.0

Fixes: #1004

#### QA Steps

*\<Commands / tests / steps to run to verify that the change works:\>*

```
tox -e py3 -- tests/unit/...
```

```
tox -e integration -- tests/integration/test_model.py::test_add_manual_machine_ssh
tox -e integration -- tests/integration/test_model.py::test_add_manual_machine_ssh_root
```

All CI tests need to pass.

#### Notes & Discussion

I've had to pin `urllib3` to `urllib3==1.25.7` to have the test pass. but the error was from pylxd failing with `TypeError: HTTPConnection.request() got an unexpected keyword argument 'chunked'`. Not sure this is related to paramiko change, since paramiko does not depend on urllib3.

```
File "/home/gboutry/python-libjuju/tests/integration/test_model.py", line 596, in test_add_manual_machine_ssh
    await add_manual_machine_ssh(event_loop, is_root=False)
  File "/home/gboutry/python-libjuju/tests/integration/test_model.py", line 463, in add_manual_machine_ssh
    client = pylxd.Client()
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/pylxd/client.py", line 383, in __init__
    response = self.api.get()
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/pylxd/client.py", line 192, in get
    response = self.session.get(self._api_endpoint, *args, **kwargs)
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/requests/sessions.py", line 602, in get
    return self.request("GET", url, **kwargs)
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/requests/adapters.py", line 486, in send
    resp = conn.urlopen(
  File "/home/gboutry/python-libjuju/.tox/py3/lib/python3.10/site-packages/urllib3/connectionpool.py", line 790, in urlopen
    response = self._make_request(
[...]

PR #1006, opened 2024-01-09
Remove explicit passing of event_loop into tests

cderici:
#### Description

This clears up from the test output the flood of warnings from pytest that looks like (e.g. [example](https://github.com/juju/python-libjuju/actions/runs/7463235016/job/20307443804#step:5:982)):

```
tests/unit/test_bundle.py:738
  tests/unit/test_bundle.py:738: PytestDeprecationWarning: test_run is asynchronous and explicitly requests the "event_loop" fixture. Asynchronous fixtures and test functions should use "asyncio.get_running_loop()" instead.
    @pytest.mark.asyncio
```

#### QA Steps

No functionality changes. Though there were a couple of tests that I needed to manually get the running loop (where the test actually was using the `event_loop`), so we need to make sure those are still passing.

#### Notes & Discussions

Maybe need to be back-ported? I'm not sure yet.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -11,20 +11,6 @@ import pytest
 from juju.client.jujudata import FileJujuData
 from juju.controller import Controller
 
-from juju.jasyncio import SingletonEventLoop
-
-
-@pytest.fixture(scope="session")
-def event_loop():
-    """
-    This fixture forces all the asyncio tests
-    to use the same events loop
-    """
-
-    loop = SingletonEventLoop().loop
-    yield loop
-    loop.close()
-
 
 def is_bootstrapped():
     try:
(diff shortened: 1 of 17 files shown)


cderici:
Failures in CI are non-related. 

`TypeError: HTTPConnection.request() got an unexpected keyword argument 'chunked'` on is being fixed in #1005, the rest it known intermittent issues that will be addressed later.

jack-w-shaw (approved):
LGTM presuming integration tests pass

jack-w-shaw:
Should this be targeting 2.9, and be merged forward?

cderici:
> Should this be targeting 2.9, and be merged forward?

I didn't think that it was happening in 2.9, but [looks like it is](https://github.com/juju/python-libjuju/actions/runs/7454120253/job/20280863629?pr=1002#step:6:817). Well we can backport this no problem I think because there's no functionality change, just removing an argument wouldn't result it any conflicts.

Issue #1010, opened 2024-01-23
Can't run python-libjuju async loop on background thread in py3.8, py3.9  and py3.10? due to "set_wakeup_fd() only works in main thread"


ajkavanagh:
### Description

For zaza, for various reasons, we need/have to run python-libjuju's async loop in a background thread.  This is because the zaza  is written as a non-blocking test framework on top of unitest.  However, the async thread needs to keep running to get status updates and so to make it all work, python-libjuju runs on the background thread.

However, due to https://github.com/python/cpython/issues/87173 and a [recent-ish change](https://github.com/juju/python-libjuju/commit/62a988fad7c574914c08a8ee9723f642ae7d5b80) this now blows up on py3.8, py3.9 and py3.10.  This is a major problem for our CI as it runs py3.8 test runners (a limitation we do eventually need to solve).

The traceback is here:

 ```
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/controller.py", line 108, in connect
      await self._connector.connect_controller(controller_name, **kwargs)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connector.py", line 127, in connect_controller
      await self.connect(
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connector.py", line 86, in connect
      self._connection = await Connection.connect(**kwargs)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 356, in connect
      await self._connect_with_redirect([_ep])
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 908, in _connect_with_redirect
      login_result = await self._connect_with_login(endpoints)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 875, in _connect_with_login
      await self._connect(endpoints)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 824, in _connect
      result = await task
    File "/usr/lib/python3.10/asyncio/tasks.py", line 571, in _wait_for_one
      return f.result()  # May raise f.exception().
[...]

PR #1011, opened 2024-01-23
Add a try: except: around loop.add_signal_handler()

ajkavanagh:
For py 3.8, 3.9 and 3.10, the set_wakeup_fd() call fails on a background thread, due to [1], which has been fixed in py3.11.  This patch simply catches the error if the python version can't do the adding of a signal handler so that the existing/previous behaviour continues.

[1] https://github.com/python/cpython/issues/87173

Closes-Bug: #1010

#### Description

Closes-Bug: #1010

#### QA Steps

```
tox -e py3 -- tests/unit/...
```

```
```

All CI tests need to pass.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -425,9 +425,14 @@ class Connection:
             for task in jasyncio.all_tasks():
                 task.cancel()
 
-        loop = jasyncio.get_running_loop()
-        for sig in (signal.SIGINT, signal.SIGTERM):
-            loop.add_signal_handler(sig, _exit_tasks)
+        try:
+            loop = jasyncio.get_running_loop()
+            for sig in (signal.SIGINT, signal.SIGTERM):
+                loop.add_signal_handler(sig, _exit_tasks)
+        except (ValueError, OSError, RuntimeError) as e:
+            # add_signal_handler doesn't work in a thread
+            if 'main thread' not in str(e):
+                raise
 
         return (await websockets.connect(
             url,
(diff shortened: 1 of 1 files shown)


cderici:
Hey @ajkavanagh, yeah I can see why this can be a problem off the main thread.

In general, I like this solution (except maybe just catch the `RuntimeError` (instead of `(ValueError, OSError, RuntimeError)`) and match the entire message instead of doing `if 'main thread' not in str(e):`, so that we'd catch all the problems other than this particular issue.

However, this makes me wonder whether pylibjuju installing signal handlers is correct to begin with. So, pylibjuju has two notions built-in to make it useful in general; a library, and a client (I'd like to eventually split these two in the future). Adding signal handlers makes sense from a client perspective, however, it doesn't makes sense at all from a library perspective.

I'm beginning to think that it's a mistake for pylibjuju to enforce certain way of handling a signal (for projects using pylibjuju as a library). So it makes slightly more sense to remove the code that adds (and removes) the signal handlers in a connection. That'll solve your problem as well. WDYT?

ajkavanagh:
> In general, I like this solution (except maybe just catch the `RuntimeError` (instead of `(ValueError, OSError, RuntimeError)`) and match the entire message instead of doing `if 'main thread' not in str(e):`, so that we'd catch all the problems other than this particular issue.

Yes, that's a reasonable point; I was essentially just copying the existing model from juju/jasyncio.py to try to keep things similar.

> However, this makes me wonder whether pylibjuju installing signal handlers is correct to begin with. So, pylibjuju has two notions built-in to make it useful in general; a library, and a client (I'd like to eventually split these two in the future). Adding signal handlers makes sense from a client perspective, however, it doesn't makes sense at all from a library perspective.

Well that was my first thought too.  Libs generally shouldn't, I guess, be handling signals, as it ought to be something that the client-of-the-library ought to do.

> I'm beginning to think that it's a mistake for pylibjuju to enforce certain way of handling a signal (for projects using pylibjuju as a library). So it makes slightly more sense to remove the code that adds (and removes) the signal handlers in a connection. That'll solve your problem as well. WDYT?

Yes, I agree with you.  It would probably be good to give an example in the lib docs on how to catch signals so that the client can clean up properly as well, but I'd agree that the onus should be on the user-of-the-library to be responsible for cleaning up the code.

As such, please fee free to reject/close this change if you'd prefer a different approach.  Thanks!

ajkavanagh:
Based on #1014 merging, this is no longer needed. Thanks for the fix!

PR #1013, opened 2024-01-24
[WIP] Support both 2.9 and 3.x at the same time

cderici:
#### Description

Uses both schemas to generate clients that have code to support both 2.9 and 3.x facade functionality at the same time (no need to keep separate sets of client files and dynamically check -- like we did in 3.0.4 to do this).

Adds a separate integration tests against 2.9, let's see which functionalities are gonna blow up, then we'll continue from there.


jneo8:
Hi @cderici 

I would like to know the status of this PR, is this one still on the schedule?

cderici:
> I would like to know the status of this PR, is this one still on the schedule?

Hey @jneo8, not really tbh, we have way more things having higher priority right now in Juju.

If you're having trouble migrating from 2.9 to 3.x feel free to reach out at Juju channel on Matrix.

jneo8:
We just want to know if python-libjuju can support this in a short time(6 months) to make the technical decision.

Thanks for reply.

Aflynn50:
> We just want to know if python-libjuju can support this in a short time(6 months) to make the technical decision.
> 
> Thanks for reply.

Hey @jneo8,
There are no plans right now to continue this work.

PR #1015, opened 2024-01-27
chore: migrate types & type hint machine module

yanksyoon:
#### Description

This migrates all the known status types from juju https://github.com/juju/juju/blob/f8e67c7f7c16f03312b96a0731a2b3fdbef1645e/core/status/status.go#L69
to the python-libjuju's `status.py` module.

*\<Fixes: \>*

#### QA Steps

```
tox -e py3 -- tests/unit/...
```

```
```

All CI tests need to pass.

#### Notes & Discussion


tlm:
@yanksyoon thank you for the pull request. I have asked @cderici to have a look as he understands a lot better what is going on. Would you mind filling out your why a bit more for the PR? Are you able to describe a potential problem this is solving for you. Doesn't have to be much just helps any future developers to understand why something change a bit better.

yanksyoon:
@tlm Sure! I thought this would be a good step for making dev-exp with Juju better. The charmers rely heavily on the `ops` library and many of the `ops` library refers to the python-libjuju components.
Updated the description!

yanksyoon:
@cderici Thanks for the explanation, I feel like that Juju having a compatibility matrix like Kubernetes does for server - client version would help a lot.
Having to run the tests and see what's being returned from juju takes quite a lot of time as a charm developer and having type hints would really reduce the amount of time spent - since Juju docs also is not versioned as of yet.

cderici:
> I feel like that Juju having a compatibility matrix like Kubernetes does for server - client version would help a lot. 

Are you talking about pylibjuju compatibility with Juju controllers? If that's the case it's quite simple, actually, currently the latest 2.9.x pylibjuju is compatible with latest Juju 2.9, and latest 3.x pylibjuju is compatible with any Juju 3.y. I might add a section to the Readme about that. Let me know if this is different than what you were talking about 👍 

> Having to run the tests and see what's being returned from juju takes quite a lot of time as a charm developer and having type hints would really reduce the amount of time spent - since Juju docs also is not versioned as of yet.

This is fair. Improving dev experience might be a good reason to add the types. My concern is, that keeping backwards compatibility in pylibjuju will require adding union types everywhere. I think a better solution would be to improve the documentation for that. WDYT? Would it make it equally easier if pylibjuju documentation is improved so that it includes type signatures?

Issue #997, opened 2023-12-27
Machine does not come online


yanksyoon:
### Description

When adding a machine and using `machine.ssh` commands, the ssh address is None (i.e. ubuntu@None). A bit of investigation shows that the machine address is None, even if the `juju status` reports the machine as `started`.

### Urgency

Annoying bug in our test suite

### Python-libjuju version

3.3

### Juju version

3.1.6

### Reproduce / Test

```python
import inspect
import time
import typing

import pytest_asyncio
from juju.machine import Machine
from juju.model import Model


async def wait_for(
    func: typing.Callable[[], typing.Union[typing.Awaitable, typing.Any]],
    timeout: int = 300,
    check_interval: int = 10,
) -> typing.Any:
    """Wait for function execution to become truthy.

    Args:
        func: A callback function to wait to return a truthy value.
        timeout: Time in seconds to wait for function result to become truthy.
        check_interval: Time in seconds to wait between ready checks.

    Raises:
        TimeoutError: if the callback function did not return a truthy value within timeout.

    Returns:
        The result of the function if any.
    """
    deadline = time.time() + timeout
    is_awaitable = inspect.iscoroutinefunction(func)
    while time.time() < deadline:
        if is_awaitable:
            if result := await func():
                return result
        else:
            if result := func():
                return result
        time.sleep(check_interval)

    # final check before raising TimeoutError.
    if is_awaitable:
        if result := await func():
            return result
    else:
        if result := func():
            return result
    raise TimeoutError()


@pytest_asyncio.fixture(scope="module", name="ssh_machine")
async def ssh_machine_fixture(model: Model):
    """A machine to test tmate ssh connection"""
    machine: Machine = await model.add_machine()

    def wait_machine_ip():
        wait_machine: Machine = model.machines[machine.entity_id]
        print(wait_machine.dns_name, wait_machine.agent_status)
        return wait_machine.dns_name and wait_machine.agent_status == "started"

    await wait_for(wait_machine_ip, timeout=60 * 10)
    print(machine.data)

    await machine.ssh("sudo apt update -y")
    await machine.ssh("sudo apt install -y tmate")

async def test_machine(ssh_machine: Machine):
    ssh_machine.ssh("echo hello")
```

PR #1016, opened 2024-01-27
fix: machine scp/ssh operations

yanksyoon:
#### Description

This is a follow up PR to [#1015](https://github.com/juju/python-libjuju/pull/1015)
This addresses the issue #997 

The current machine.ssh or machine.scp will fail due to machine address not being ready.
This makes the machine ssh command become `ssh ubuntu@None`.

*\<Fixes: \>*
Added a function `machine.wait()` which waits for machine agent to become ready and the machine address to exist.
The scp/ssh functions now will raise if `await machine.wait()` hasn't been called.

#### QA Steps

```
tox -e integration -- -k test_machine_ssh
```

All CI tests need to pass.

#### Notes & Discussion


tlm:
@cderici mind taking a look? Or happy to do a dual review if you want to shadow me through this to make sure my assumptions are correct?

cderici (commented):
This is nice, but instead of failing in `machine.ssh` and `machine.scp`, I'd add an optional `wait_until_active=False` argument and inline the body of `wait()` in where you're currently raising an error. Note that `block_until` is defined in the model, so you're gonna need to call it like `self.model.block_until()` in `Machine` methods.

Also this in it's current implementation requires heavy testing, as the `status` is a bit finicky in pylibjuju, as the status we get through the api is not exactly the same status you'd see via juju-cli (i.e. `juju status`). For machines it shouldn't be too much of a problem, however, I would refrain from checking the status directly (like `self.agent_status == "started"`), as there might be a race where in between the block_until check intervals the status might change and you'd lose the time to catch it, so I wouldn't want my code to rely on the status itself. Instead I'd check just the addresses, as that's what you essentially need.

However, `safe_data` shouldn't be used outside the class, as it's internal, so let's add a `machine.addresses()` method and use that (also in the `machine.dns_name()` as well). WDYT?

yanksyoon:
Thank you for the review @cderici !

I've applied the changes - but since this was branched off of #1015 , shall I wait until the conversation there is resolved? Or I could cherry-pick the changes for this PR and branch if off of main again :)

yanksyoon:
@cderici I've created #1020 with cherry-picked commits, I'll close this one! Thank you for the feedback so far! :D

Issue #781, opened 2023-01-04
[Feature]: Add `block_until_done`  arg to `remove_relation`


sed-i:
### Description

#665 added a `block_until_done` arg to `remove_application`.
It would be super handy to have the same for `remove_relation`.

https://github.com/juju/python-libjuju/blob/88eb6c4b85d7d960fdc65b89ac0d5e79ee4cc6bc/juju/application.py#L195

### Urgency

It would save from a weird workaround

### Code I'd Like to Run

```python
await ops_test.model.applications[trfk.name].remove_relation(f"{ipr.name}:ingress", trfk.name, block_until_done=True)
# and then, without any additional `block_until`:
await ops_test.model.applications[trfk.name].add_relation(f"{ipr.name}:ingress", trfk.name)
```

hmlanigan:
Closed prematurely by the bot.

PR #1017, opened 2024-01-27
feat: remove app block until done

yanksyoon:
#### Description

Addresses #781  to add `block_until_done` to `remove_relation`.

*\<Fixes: \>*

#### QA Steps

```
tox -e integration -- -k test_app_relation_destroy_block_until_done
```

All CI tests need to pass.

#### Notes & Discussion

### DIFF
--- a/tests/integration/test_application.py
+++ b/tests/integration/test_application.py
@@ -3,6 +3,7 @@
 
 import logging
 from pathlib import Path
+import asyncio
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tlm:
@cderici same again. Happy to do the review if you want to triple check it for me at the end.

tests/integration/test_application.py
```diff
@@ -331,3 +333,14 @@ async def test_app_charm_name():
...
+
+@base.bootstrapped
+async def test_app_relation_destroy_block_until_done():
+    async with base.CleanModel() as model:
+        app: Application = await model.deploy('docker-registry')
+        rsa: Application = await model.deploy("easyrsa")
+        await app.relate("cert-provider", rsa.name)
+        await model.wait_for_idle(status="active")
+        await app.destroy_relation("cert-provider", rsa.name, block_until_done=True)
+        await app.relate("cert-provider", rsa.name)
```
cderici, line 346: Instead of relating again, let's call the `relations` here and make sure we don't see the relation there.

cderici (changes requested):
Thanks for tackling this ❤️ this mostly LGTM, with one little nit in the test 👍

tests/integration/test_application.py
yanksyoon: Done, thanks for the feedback!

cderici (approved):

Issue #855, opened 2023-05-16
[Feature]: remove_application support timeout and status output


beliaev-maksim:
### Description

We have a test where we want to remove the app and deploy another one

However, juju for unrelated to this lib reason cannot effectively remove the app and apps stays in terminated state (I assume)
and our CI test hangs for 6h until GitHub kills the runner.

would be great if remove_application would support timeout argument (similar to `wait_for_idle`)
and if it will print into STDOUT status of the app/unit that was requested to terminate (similar to `wait_for_idle` but only for this app)

### Urgency

It would save from a weird workaround

### Code I'd Like to Run

```python
await ops_test.model.remove_application(APP_NAME, block_until_done=True, timeout=600)
```

beliaev-maksim:
@cderici not sure that feature requests should be autoclosed in 30 days

beliaev-maksim:
@cderici reopen this one or we decide as no go for this feature?

cderici:
@beliaev-maksim yeah feel free to reopen any issues/PRs you want 👍

hmlanigan:
Closed prematurely by the bot.

PR #1018, opened 2024-01-27
feat: remove app timeout

yanksyoon:
#### Description

Adds timeout parameter to `model.remove_application()`
Addresses #855 

*\<Fixes: \>*

#### QA Steps

```
tox -e integration -- -k test_app_remove_timeout
```

All CI tests need to pass.

#### Notes & Discussion

### DIFF
--- a/tests/integration/test_application.py
+++ b/tests/integration/test_application.py
@@ -2,6 +2,7 @@
 # Licensed under the Apache V2, see LICENCE file for details.
 
 from pathlib import Path
+import asyncio
 
 import pytest
 import logging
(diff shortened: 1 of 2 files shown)


tlm:
@cderici same again :)

tests/integration/test_application.py
```diff
@@ -324,6 +325,16 @@ async def test_app_remove_wait_flag():
         assert a_name not in model.applications
 
 
+@base.bootstrapped
+async def test_app_remove_timeout():
+    async with base.CleanModel() as model:
+        app = await model.deploy('ubuntu')
```
cderici, line 331:
```suggestion
        app = await model.deploy('juju-qa-test')
```

cderici (approved):
This LGTM. Thanks for the change ❤️The test might fail in the CI (because we're not doing the best job of handling exceptions in the asyncio event loop - and for some reason the GH actions exposes that), however, it passes for me locally. One little change in the test and we can land this safely.

PR #1020, opened 2024-02-01
fix: machine scp & ssh

yanksyoon:
#### Description

This is a follow-up PR to #1016 , which contains cherry-picked commits without type imports.

Fİxes #997 

#### QA Steps

```
tox -e py3 -- tests/unit/...
```

```
tox -e integration -- tests/integration/test_machine.py
```

All CI tests need to pass.

#### Notes & Discussion


tests/integration/test_machine.py
```diff
@@ -36,3 +37,39 @@ async def test_status():
                          machine.status_message.lower() == 'running' and
                          machine.agent_status == 'started')),
             timeout=480)
+
+
+@base.bootstrapped
+async def test_scp(event_loop):
```
cderici, line 43: This test seems redundant to me, as we already have it in the [intergration/test_unit.py](https://github.com/juju/python-libjuju/blob/16149dbef02160a7f4170e1630bfa069ba275479/tests/integration/test_unit.py#L160), and `unit.scp_to` is actually a wrapper around `machine.scp_to`, so it's not doing anything new or different (aside from the `add_machine` call I suppose).

tests/integration/test_machine.py
```diff
@@ -36,3 +37,39 @@ async def test_status():
...
+    async with base.CleanModel() as model:
+        await model.add_machine()
+        await asyncio.wait_for(
+            model.block_until(lambda: model.machines),
+            timeout=240)
+        machine = model.machines['0']
+        await asyncio.wait_for(
+            model.block_until(lambda: (machine.status == 'running' and
+                                       machine.agent_status == 'started')),
+            timeout=480)
```
cderici, line 58: Instead of doing these, let's modify this to actually use the newly added `wait_for_active` flag and test that it's waiting correctly👍

tests/integration/test_machine.py
```diff
@@ -36,3 +37,39 @@ async def test_status():
...
+
+        with NamedTemporaryFile() as f:
+            await machine.scp_from('testfile', f.name, scp_opts='-p')
+            assert f.read() == b'testcontents'
+
+
+async def test_machine_ssh():
+    async with base.CleanModel() as model:
+        machine: Machine = await model.add_machine()
+        out = await machine.ssh("echo hello world!")
```
cderici, line 73: Same redundancy here, see the `test_ssh` test in [integration/test_unit.py](https://github.com/juju/python-libjuju/blob/16149dbef02160a7f4170e1630bfa069ba275479/tests/integration/test_unit.py#L193), `unit.ssh` is just a wrapper around the `machine.ssh`.

cderici (changes requested):
Thanks for opening this ❤️ with just a couple changes in the tests and then we can land this 👍

cderici:
Hi @yanksyoon, just pinging for this to let you know we're planning on making a release soon, so I'd love to roll this in if it's ready as well 👍

yanksyoon:
@cderici Sorry! I was on my break, i'll get on this asap!

cderici:
> @cderici Sorry! I was on my break, i'll get on this asap!

@yanksyoon np, hope you had a good break!

Feel free to ping me whenever this is ready for review again 👍

tests/integration/test_machine.py
yanksyoon: Removed the test - it is redundant indeed :) Thank you for the review!

tests/integration/test_machine.py
yanksyoon: This test has now been removed.

tests/integration/test_machine.py
yanksyoon: Added `wait_for_active` parameter here - this should no longer be redundant :)

yanksyoon:
@cderici The tests seem to have failed due to changes in how juju handles old charm-store (`cs:`) urls. Looking into it!

cderici (changes requested):
Yeah the CI failures are unrelated to this change. This can land. 

One final request though, the `wait_for_active` should be an opt-in feature, so let's have `False` as the default value, as this would currently block any code that uses this without this parameter if the machine is not ready. Granted, it would be weird for them to rely on failure, but still, I'd steer away from changing default semantics on this. Hope this makes sense.

Thanks for working on this! <3

yanksyoon:
@cderici Changed! Thanks for the review! :D

cderici:
@yanksyoon looks like the `test_machine_ssh` was failing in the CI, indicating a race that happened between the time the machine gets the IP and the ssh service to be up and running, so I just added a retry to get around that. I'll approve and land this after the tests run 👍 thanks again for working on this!

cderici (approved):

PR #1026, opened 2024-02-14
Resolve build problems in 3.3.1.0

### LINKED ISSUE #1025: Libjuju 3.3.1.0 is unusable
### Description

# From inside a snap
When trying to import with `from juju.controller import Controller` from inside a snap, we get the following failure:

```
  File "/snap/openstack/x2/bin/sunbeam", line 6, in <module>
    from sunbeam.main import main
  File "/snap/openstack/x2/lib/python3.10/site-packages/sunbeam/main.py", line 23, in <module>
    from sunbeam.commands import configure as configure_cmds
  File "/snap/openstack/x2/lib/python3.10/site-packages/sunbeam/commands/configure.py", line 26, in <module>
    from sunbeam import utils
  File "/snap/openstack/x2/lib/python3.10/site-packages/sunbeam/utils.py", line 33, in <module>
    from sunbeam.plugins.interface.v1.base import PluginError
  File "/snap/openstack/x2/lib/python3.10/site-packages/sunbeam/plugins/interface/v1/base.py", line 30, in <module>
    from sunbeam.jobs.deployment import Deployment
  File "/snap/openstack/x2/lib/python3.10/site-packages/sunbeam/jobs/deployment.py", line 20, in <module>
  File "/snap/openstack/x2/lib/python3.10/site-packages/juju/controller.py", line 12, in <module>
    from .client import client, connector
  File "/snap/openstack/x2/lib/python3.10/site-packages/juju/client/connector.py", line 12, in <module>
    from juju.client.connection import Connection
  File "/snap/openstack/x2/lib/python3.10/site-packages/juju/client/connection.py", line 19, in <module>
    from juju.version import CLIENT_VERSION
  File "/snap/openstack/x2/lib/python3.10/site-packages/juju/version.py", line 19, in <module>
    CLIENT_VERSION = re.search(r'\d+\.\d+\.\d+', open(VERSION_FILE_PATH).read().strip()).group()
FileNotFoundError: [Errno 2] No such file or directory: '/snap/openstack/x2/lib/python3.10/site-packages/VERSION'
```

# From inside a venv
When trying to import `from juju.controller import Controller` from inside a virtualenv, we get the following failure:

```
python3 -c "from juju.controller import Controller"
  File "<string>", line 1, in <module>
  File "/home/ubuntu/testos/lib/python3.10/site-packages/juju/controller.py", line 12, in <module>
    from .client import client, connector
  File "/home/ubuntu/testos/lib/python3.10/site-packages/juju/client/connector.py", line 7, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```

Side notes:
The snap has other dependencies that to include `packaging`, so no failure on packaging import
Everything works fine with Juju `3.3.0.0`

### Urgency

### Python-libjuju version

3.3.1.0

### Juju version

Unrelated

### Reproduce / Test

```python
How to reproduce both failure in a venv:


sudo apt update
sudo apt install python3.10-venv
python3 -m venv venv
source venv/bin/activate
pip install 'juju==3.3.1.0'
python3 -c "from juju.controller import Controller" <-- packaging import missing
pip install packaging
python3 -c "from juju.controller import Controller" <-- VERSION file not found error
```

cderici:
#### Description

This solves 2 problems with the latest release (also described in [this comment](https://github.com/juju/python-libjuju/issues/1025#issuecomment-1944086744)):

1. We depend on the `packaging` module, but didn't include it in the `setup.py` (because our testing and dev environments have tox installed and tox satisfies that requirement for itself, so we didn't notice).
2. We [accidentally depend](https://github.com/juju/python-libjuju/blob/905de781cb92041dcae158412861464d87a8bb77/juju/version.py#L18) on the static [VERSION ](https://github.com/juju/python-libjuju/blob/master/VERSION) file where we keep track of the version of pylibjuju.

This change fixes #1025, solving these issues by:

1. Adding the `packaging` to the required modules.
2. Removing the file `VERSION`, and moving the static version information into the `version.py` module (just like in Juju) and use that information from there. The places that dynamically use that information (outside of the juju modules) are i) `setup.py` to get the version info into the build, ii) `Makefile` to use it in the release target iii) `docs/config.py` for building the docs. All of these spots are updated to use the info within the `version.py`.

#### QA Steps

We'll use the steps described in #1025. After pulling the branch in this PR;

- Create a new virtual env:

```
python -m venv venv
```

- Activate it:

```
source venv/bin/activate
```

- Build pylibjuju at this spot:

```
python setup.py sdist
```

- The sdist build will create a `dist/juju-3.3.1.0.tar.gz`, so install that:

```
pip install dist/juju-3.3.1.0.tar.gz
```

- Now try to import a `Controller`.

```
python3 -c "from juju.controller import Controller"
```

Without the (1)st change in this PR, this command would result in:

```
ModuleNotFoundError: No module named 'packaging'
```

And without the (2)nd change, this command would error with 

```
FileNotFoundError: [Errno 2] No such file or directory: '......./venv/lib/python3.10/site-packages/VERSION'
```

But with this change, we shouldn't see either (or any other) error. I.e., the command should succeed with no problems.

#### Notes & Discussion

Note that this changes the release process. In particular where we need to manually change the version information is moved into the `version.py` module (the release process document will be updated right before this PR is merged).

Moreover, we need to make a patch release (`3.3.1.1`) after this for these changes to take effect (i.e. for pylibjuju to be usable in the environments where it's installed as a dependency).

JUJU-5484


docs/conf.py
```diff
@@ -21,13 +21,16 @@
...
 here = Path(__file__).absolute().parent
-version = (here.parent / 'VERSION').read_text().strip()
+
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
+from juju.version import CLIENT_VERSION
```
jack-w-shaw, line 31: Imports should be at the top no?

setup.py
```diff
@@ -13,11 +15,10 @@
     changelog.read_text()
 )
 long_description_content_type = 'text/x-rst'
-version = here / 'VERSION'
 
 setup(
     name='juju',
-    version=version.read_text().strip(),
+    version=CLIENT_VERSION.strip(),
```
jack-w-shaw, line 21: Do we still need to strip?

jack-w-shaw (approved):

docs/conf.py
cderici: This one should stay there because it won't work before the `sys.path.insert(0, os.path.abspath('..'))` line above

setup.py
cderici: We currently don't need to, but the line that's coming from will be manually edited by people who're making releases (replaces the file VERSION) in the future, so I thought this would cleanup any silly mistakes like "4.2.0 " (whitespace in the end).

Issue #1028, opened 2024-02-20
AttributeError: 'NoneType' object has no attribute 'exception'


freyes:
### Description

libjuju crashes when the shared state watcher was stopped, libjuju tries to handle and fails with the following error: `AttributeError: 'NoneType' object has no attribute 'exception'`

We are seeing this issue in our CI - https://github.com/openstack-charmers/zaza/actions/runs/7907308186/job/21769045812?pr=608

The full stacktrace is:

```
  Error: -20 13:19:12 [ERROR] Exception in callback create_task_with_handler.<locals>._task_result_exp_handler(task_name='tmp', logger=None)(<Task finishe...was stopped')>) at /home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/jasyncio.py:53
  handle: <Handle create_task_with_handler.<locals>._task_result_exp_handler(task_name='tmp', logger=None)(<Task finishe...was stopped')>) at /home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/jasyncio.py:53>
  Traceback (most recent call last):
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/jasyncio.py", line 55, in _task_result_exp_handler
      task.result()
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/facade.py", line 486, in wrapper
      reply = await f(*args, **kwargs)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/_client1.py", line 315, in Next
      reply = await self.rpc(msg)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/overrides.py", line 113, in rpc
      result = await self.connection.rpc(msg, encoder=TypeEncoder)
    File "/home/runner/work/zaza/zaza/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 635, in rpc
      raise errors.JujuAPIError(result)
  juju.errors.JujuAPIError: shared state watcher was stopped
  During handling of the above exception, another exception occurred:
  Traceback (most recent call last):
    File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
      self._context.run(self._callback, *self._args)
[...]

PR #1035, opened 2024-03-20
Bugfix none type

Aflynn50:
#### Description

Fix for #1028 
The logger is not passed to calls of `utils.run_with_interupt` from model.py and controller.py. The default value is `None`, so when it is later used we get the error. 
*\<Fixes:  \>*

#### QA Steps
Added a `raise Exception("my random exception")` to the `Next` method of `ModelSummeryWatcherFacade`. This is hit by the `run_with_interupt` in controller and we can observe that when we run the file `modelsummaries.py`, if the log arguments added in this PR are not there then we get the None Type error and if they are there, we do not.

Jira: JUJU-5610

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ pytestdebug.log
 .vscode/
 deb_dist
 *.tar.gz
+.idea/
--- a/juju/controller.py
+++ b/juju/controller.py
@@ -828,7 +828,8 @@ class Controller:
                     try:
                         results = await utils.run_with_interrupt(
                             watcher.Next(),
-                            stop_event)
+                            stop_event,
+                            log=log)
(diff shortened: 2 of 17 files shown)


cderici (approved):
Yeah the CI is unrelated, and will be fixed soon by the recent changes.

Can we also take care of [these warnings](https://github.com/juju/python-libjuju/actions/runs/8367029700/job/22958419133?pr=1035#step:6:978) as a drive-by, [I fixed it in 3.x](https://github.com/juju/python-libjuju/pull/1006), and I thought I back-ported it, but looks like it fell through the cracks. Feel free to just cherry-pick those into this one or open another PR either is fine 👍 

Thanks for the fix!

Issue #1031, opened 2024-03-07
Model.consume() from a different controller doesn't work


sudeephb:
Calling `model.consume` as shown below causes an error.
```
await model.consume(
               f"admin/{k8s_model.name}.grafana-dashboards",
               application_alias="grafana",
               controller_name=k8s_ctl.controller_name,
        )
```
The error looks like this:
```
FAILED ... - juju.errors.JujuError: ['application offer "admin/current-model.grafana-dashboards" not found'] 
```

where, `current-model` is the model name, and `grafana-dashboards` offer is exposed.

As a workaround, using juju CLI works ie. the following works fine
```
        cmd = [
            "juju",
            "consume",
            "--model",
            f"current-controller:current-model",
            f"{k8s_ctl.controller_name}:admin/{k8s_model.name}.grafana-dashboards",
        ]
        subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```

The name of the model is the same in both controllers.
python-libjuju version is `3.3.1.1`

Aflynn50:
This works if you also put the controller name in the endpoint url, for example:
```
await model.consume(
               f"k8s_ctl.controller_name:admin/{k8s_model.name}.grafana-dashboards",
               application_alias="grafana",
               controller_name=k8s_ctl.controller_name,
      )
```

The name of the argument is pretty confusing though, I will bring a PR to allow you to just specify the controller name in the endpoint rather than having to do it separately as well.

PR #1038, opened 2024-03-27
Make consume respect the controller name in the url

Aflynn50:
#### Description

`consume` takes a `controller_name` argument. This is not needed and should be automatically generated from the offer url. A deprecation warning is logged and the variable is not used.

Also update offer and consume integration tests to no longer be skipped.

This helps avoid #1031 

#### QA Steps
```
juju bootstrap lxd c1
juju add-model offerer
juju deploy juju-qa-dummy-source dummy-source
juju offer dummy-source:sink
juju bootstrap lxd c2
python -m asyncio

# In REPL
from juju import model
m = model.Model()
await m.connect()
await m.consume("c1:admin/offerer.dummy-source")

# Exit REPL
juju status
# Confirm SAAS exists
juju remove-saas dummy-source
# Confirm removal
python -m asyncio

# In REPL
from juju import model
m = model.Model()
await m.connect()
await m.consume("admin/offerer.dummy-source", controller_name="c1")

# Exit REPL
juju status
# Confirm SAAS exists
juju remove-saas dummy-source
# Confirm removal
python -m asyncio

# In REPL
from juju import model
m = model.Model()
await m.connect()
await m.consume("c1:admin/offerer.dummy-source", controller_name="c1")

```
All CI tests need to pass.


tests/integration/test_crossmodel.py
```diff
@@ -188,4 +188,4 @@ async def test_add_bundle():
                 await model_2.deploy('local:{}'.format(tmp_path))
                 await model_2.wait_for_idle(status="active")
 
-            await model_1.remove_offer("admin/{}.influxdb".format(model_1.name), force=True)
+            await model_1.remove_offer("admin/{}.influxdb".format(model_1.name), force=True)
```
cderici, line 191: linter must be not happy with this change :)

cderici (changes requested):
Code looks good. Could you please check if there's a test with a controller name in the endpoint url? If not we should add that (for `parse_offer_url`, or `_get_source_api`) I'd be ok if it's a mocked test within `tests/unit` if coming up with a sensible integration test is hard. We just need to ensure the capability is there for future. Hope that makes sense.

Also looks like we need a rebase.

Aflynn50:
> Code looks good. Could you please check if there's a test with a controller name in the endpoint url? If not we should add that (for `parse_offer_url`, or `_get_source_api`) I'd be ok if it's a mocked test within `tests/unit` if coming up with a sensible integration test is hard. We just need to ensure the capability is there for future. Hope that makes sense.
> 
> Also looks like we need a rebase.

I've added a unit test which checks the consume behavior, though its uses a lot of mocks.

I've also changed it so you can use the controller_name argument as expected, just to ensure that we don't break anything with this.

tests/unit/test_offerendpoint.py
```diff
@@ -90,3 +93,59 @@ def test_parse_local_endpoint_failures(self):
...
+
+
+class TestConsume(unittest.IsolatedAsyncioTestCase):
+    @mock.patch.object(Model, 'connection')
+    @mock.patch.object(Controller, 'disconnect')
+    @mock.patch('juju.model._create_consume_args')
+    @mock.patch('juju.client.client.ApplicationFacade.from_connection')
+    async def test_external_controller_consume(self, mock_from_connection,
+                                               mock_controller, mock_connection, mock_create_consume_args):
+        """ Test consuming an offer an external controller. This would be better
```
cderici, line 105:
```suggestion
        """ Test consuming an offer by an external controller. This would be better
```

cderici (approved):
Yeah this looks good 👍 mocking is fine as we can't bootstrap, the purpose of the test is just to protect the api in the future.

PR #1041, opened 2024-04-04
Fix multiline description in textarea in bug template GH workflow

cderici:
#### Description

Fixes the bug in the bug template yml. Apparently the indentation + the quotes in the message caused the yml parser to interpret the line as a new field and barf.

#### QA Steps

No functional changes. Just updating the bug template. 

I tried to put the yaml in an [online validator](https://jsonformatter.org/yaml-validator/) and was able to see the same parsing issue. The issue went away after I applied the changes here and revalidated. Maybe that would help QA it as well.


Aflynn50 (approved):
LGTM

Issue #1052, opened 2024-05-17
Incorrect parsing of storage constraints


luissimas:
### Description

When deploying a bundle with an application that contains storage constraints the deployment fails with the following message:

```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### Urgency

Annoying bug in our test suite

### Python-libjuju version

3.4

### Juju version

3.5.0

### Reproduce / Test

```python
import asyncio
from juju.model import Model

bundle_file = "./bundle.yaml"

bundle = """
name: sample-bundle

series: jammy

machines:
  "0":
    constraints: instance-type=type1

applications:
  swift-storage:
    charm: swift-storage
    channel: yoga/stable
    num_units: 1
    to:
      - "0"
    storage:
      block-devices: cinder,1,5G
"""

async def main():
    with open(bundle_file, "w") as f:
        f.write(bundle)

    model = Model()
    await model.connect()
    await model.deploy(bundle_file)

asyncio.run(main())
```

PR #1053, opened 2024-05-17
fix parsing of storage constraints

luissimas:
#### Description

Fixes: #1052. Parses the storage constraints when deploying applications. Before this change the storage constraints where not parsing, resulting in an error when deploying bundles that contained applications with storage definitions.

#### QA Steps

The following python script can be used to verify both the bug in the current version as well as the fix implemented:

```python
import asyncio
from juju.model import Model

bundle_file = "./bundle.yaml"

bundle = """
name: sample-bundle

series: jammy

machines:
  "0":
    constraints: instance-type=type1

applications:
  swift-storage:
    charm: swift-storage
    channel: yoga/stable
    num_units: 1
    to:
      - "0"
    storage:
      block-devices: cinder,1,5G
"""

async def main():
    with open(bundle_file, "w") as f:
        f.write(bundle)

    model = Model()
    await model.connect()
    await model.deploy(bundle_file)

asyncio.run(main())
```

All CI tests need to pass.

#### Notes & Discussion

I wasn't able to add a regression test for this fix since the `juju-qa-test` charm does not support the addition of storage devices. With that said, I'm open to suggestions on how to test this behavior to make sure it's correct and that it stays that way on future changes.

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -29,6 +29,7 @@ from .charmhub import CharmHub
 from .client import client, connector
 from .client.overrides import Caveat, Macaroon
 from .constraints import parse as parse_constraints
+from .constraints import parse_storage_constraint
 from .controller import Controller, ConnectedController
 from .delta import get_entity_class, get_entity_delta
 from .errors import JujuAPIError, JujuError, JujuModelConfigError, JujuBackupError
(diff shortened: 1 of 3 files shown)


kwmonroe (approved):
this bit me as well on `juju==3.4.0.0`.  confirmed this PR fixes this on that release.

luissimas:
@cderici I've added a integration test in 132f96c29b01. I had to make some changes to the bundle to be able to deploy it on an LXD cloud, which I assume is the expected environment for the integration tests. With that said, I was able to reproduce both the bug and the fix using this bundle.

cderici (approved):

Issue #1062, opened 2024-06-19
Add support for noble


mthaddon:
### Description

Pylibjuju doesn't support noble yet, per https://github.com/juju/python-libjuju/blob/master/juju/utils.py#L305.

Failing test run available [here](https://github.com/canonical/penpot-operator/actions/runs/9585018551/job/26430073174?pr=2). Traceback for posterity in case the run expires:

```
____________________________ test_build_and_deploy _____________________________
Traceback (most recent call last):
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 241, in <lambda>
    lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 139, in _multicall
    raise exception.with_traceback(exception.__traceback__)
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 122, in _multicall
    teardown.throw(exception)  # type: ignore[union-attr]
  File "/home/ubuntu/github-runner/_work/penpot-operator/penpot-operator/.tox/integration/lib/python3.10/site-packages/_pytest/threadexception.py", line 87, in pytest_runtest_call
    yield from thread_exception_runtest_hook()
[...]

PR #1063, opened 2024-06-26
fix(series): add noble support

jack-w-shaw:
The pylibjuju client hardcodes a mapping between series codename and the base channel.

Noble was absent from this mapping, meaning pylibjuju doesn't know know to deploy to noble.

Add noble to this mapping.

After noble, we will only support deploying by base. Noble will be the last series added to this map

Fixes: https://github.com/juju/python-libjuju/issues/1062

#### QA Steps

Download a local charm `juju download ubuntu`, unzip, and amend the `manifest,yaml` to support ubuntu@24.04

Place the charm dir at `./ubuntu`

You will see 
```
$ cat ./ubuntu/manifest.yaml
bases:
- architectures:
  - amd64
  channel: '20.04'
  name: ubuntu
- architectures:
  - amd64
  channel: '22.04'
  name: ubuntu
- architectures:
  - amd64
  channel: '24.04'
  name: ubuntu
```

Then:
```
$ juju bootstrap lxd lxd
$ juju add-model
$ python -m asyncio
>>> m = Model()
>>> await m.connect()
>>> await m.deploy("./ubuntu", series="noble")
<Application entity_id="ubuntu">
>>> exit
(wait)
$ juju status
 Model  Controller  Cloud/Region         Version      SLA          Timestamp
m      lxd         localhost/localhost  3.6-beta1.1  unsupported  12:03:15+01:00

App     Version  Status  Scale  Charm   Channel  Rev  Exposed  Message
ubuntu  24.04    active      1  ubuntu             0  no       

Unit       Workload  Agent  Machine  Public address  Ports  Message
ubuntu/0*  active    idle   0        10.219.211.77          

Machine  State    Address        Inst id        Base          AZ  Message
0        started  10.219.211.77  juju-742671-0  ubuntu@24.04      Running
```

All CI tests need to pass.

### DIFF
--- a/juju/utils.py
+++ b/juju/utils.py
@@ -301,6 +301,7 @@ JAMMY = "jammy"
 KINETIC = "kinetic"
 LUNAR = "lunar"
 MANTIC = "mantic"
+NOBLE = "noble"
 
 UBUNTU_SERIES = {
     PRECISE: "12.04",
@@ -327,6 +328,7 @@ UBUNTU_SERIES = {
     KINETIC: "22.10",
     LUNAR: "23.04",
     MANTIC: "23.10",
+    NOBLE: "24.04",
 }
 
 KUBERNETES = "kubernetes"
--- a/tests/integration/charm-manifest/manifest.yaml
+++ b/tests/integration/charm-manifest/manifest.yaml
@@ -9,5 +9,9 @@ bases:
   - amd64
   channel: '20.04'
   name: ubuntu
+- architectures:
+  - amd64
+  channel: '24.04'
+  name: ubuntu
 charmcraft-started-at: '2021-08-20T08:09:00.639806Z'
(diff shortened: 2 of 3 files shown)


cderici (approved):
Yeah this makes sense, let's also add an integration test that deploys a workload on noble, just to catch future regressions about that, because compatibility across versions will be relevant when the 4.0 comes, we don't wanna miss anything there.

jack-w-shaw:
> Yeah this makes sense, let's also add an integration test that deploys a workload on noble, just to catch future regressions about that, because compatibility across versions will be relevant when the 4.0 comes, we don't wanna miss anything there.

Integration test added

UtkarshBhatthere:
Hey guys! is there any ETA for this patch to land ?

sabaini:
Hm, it looks like merging is still blocked?

Issue #1057, opened 2024-05-30
[Refresh] revision option does not work


phvalguima:
### Description

I am currently deploying OpenSearch operator: https://charmhub.io/opensearch

Writing an integration test that must start on rev. 90 (latest currently is rev. 95) and then jump:
90 > 91 > 95

However, the first jump above fails. Does not matter how I configure it, it always jumps: `90 > 95`.

I have tried:
```
        await application.refresh(
            channel=OPENSEARCH_CHANNEL,
            rev=91,
        )
```

Also:
```
        await application.refresh(
            channel=OPENSEARCH_CHANNEL,
            switch="ch:amd64/jammy/opensearch-91",
        )
```

And still seeing the same issue.

One thing I've noticed, we are using "ResolveCharms" facade version 6. There are newer versions.

I also built juju for debug and inspected both what libjuju is sending and what `juju refresh` command sends.

Here are the comparison:
```
$ juju refresh opensearch --revision 91

...

# In my controller, I can see with GDB:
345 // ResolveCharms resolves the given charm URLs with an optionally specified
346 // preferred channel.  Channel provided via CharmOrigin.
347 func (a *API) ResolveCharms(args params.ResolveCharmsWithChannel) (params.ResolveCharmWithChannelResults, error) {
348 logger.Tracef("ResolveCharms %+v", args)
349 if err := a.checkCanRead(); err != nil {
350 return params.ResolveCharmWithChannelResults{}, errors.Trace(err)
351 }
(gdb) p args
$1 = {Resolve = []github.com/juju/juju/rpc/params.ResolveCharmWithChannel = {{Reference = "ch:amd64/jammy/opensearch-91", Origin = {Source = "charm-hub",
        Type = "", ID = "", Hash = "", Risk = "edge", Revision = 0xc004975a60, Track = 0xc00390c760, Branch = 0x0, Architecture = "amd64", Base = {
          Name = "ubuntu", Channel = "22.04/stable"}, InstanceKey = ""}, SwitchCharm = true}}, Macaroon = 0x0}
(gdb) c
```

With libjuju, where my code is set with:
```
        await application.refresh(
            channel=OPENSEARCH_CHANNEL,
            rev=91,
        )
```
I can see:
```
425 // ResolveCharms resolves the given charm URLs with an optionally specified
426 // preferred channel.  Channel provided via CharmOrigin.
427 // We need to include SupportedSeries in facade version 6
[...]

PR #1067, opened 2024-06-28
fix(refresh): bug with revisions

jack-w-shaw:
#### Description

We are incorrectly refreshing charms. The current method queries the controller for the charm's current origin and charm url

In pylibjuju, we then just overwrite the appropriate fields. However, this leaves behind id_ and hash_

This means when we later refresh, we are always returned the most recent revision by charmhub, no matter what revision we specify.

The way we handle this in the Juju CLI is by reconstructing the charm origin using MakeOrigin:
https://github.com/juju/juju/blob/a03ade4d358b1b0f135374b15b1bcd6b46d0ba0f/cmd/juju/application/utils/origin.go#L20-L75

This ensures only the fields we expect are present.

Do something similar in pylibjuju, ensuring we behave the same as the CLI.

Fixes: https://github.com/juju/python-libjuju/issues/1057

#### QA Steps

Unit tests
```
tox -e unit
```

Integration tests
```
tox -e integration -- tests/integration/test_application.py::test_refresh_revision
```

Manual tests
```
$ juju bootstrap lxd lxd
$ juju add-model m
$ juju deploy opensearch --revision 90 --channel=2/edge
$ python -m asyncio
>>> m = Model()
>>> await m.connect()
>>> app = m.applicatyions["opensearch"]
>>> await app.refresh(revision=91)
>>> exit()
$ juju status
Model  Controller  Cloud/Region         Version    SLA          Timestamp
m      lxd1        localhost/localhost  3.6-beta1  unsupported  13:32:44+01:00

App         Version  Status       Scale  Charm       Channel  Rev  Exposed  Message
opensearch           maintenance      1  opensearch  2/edge    91  no       Installing OpenSearch...

Unit           Workload     Agent      Machine  Public address  Ports  Message
opensearch/0*  maintenance  executing  0        10.219.211.254         (install) Installing OpenSearch...

Machine  State    Address         Inst id        Base          AZ  Message
0        started  10.219.211.254  juju-56bc09-0  ubuntu@22.04      Running
```

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -12,12 +12,13 @@ from .annotationhelper import _get_annotations, _set_annotations
 from .bundle import get_charm_series, is_local_charm
 from .client import client
 from .errors import JujuApplicationConfigError, JujuError
-from .origin import Channel, Source
+from .origin import Channel
 from .placement import parse as parse_placement
 from .relation import Relation
 from .status import derive_status
 from .url import URL
 from .utils import block_until
+from .version import DEFAULT_ARCHITECTURE
 
 log = logging.getLogger(__name__)
 
@@ -691,12 +692,14 @@ class Application(model.ModelEntity):
         if charm_url_origin_result.error is not None:
             err = charm_url_origin_result.error
             raise JujuError(f'{err.code} : {err.message}')
-        origin = charm_url_origin_result.charm_origin
 
+        current_origin = charm_url_origin_result.charm_origin
         if path is not None or (switch is not None and is_local_charm(switch)):
-            await self.local_refresh(origin, force, force_series,
+            await self.local_refresh(current_origin, force, force_series,
                                      force_units, path or switch, resources)
             return
+
+        origin = _refresh_origin(current_origin, channel, revision)
 
         # If switch is not None at this point, that means it's a switch to a store charm
         charm_url = switch or charm_url_origin_result.url
@@ -705,20 +708,6 @@ class Application(model.ModelEntity):
 
         if parsed_url.schema is None:
             raise JujuError(f'A ch: or cs: schema is required for application refresh, given : {str(parsed_url)}')
-
-        if revision is not None:
-            origin.revision = revision
-
-        # Make the source-specific changes to the origin/channel/url
-        # (and also get the resources necessary to deploy the (destination) charm -- for later)
-        origin.source = Source.CHARM_HUB.value
-        if channel:
-            ch = Channel.parse(channel).normalize()
-            origin.risk = ch.risk
-            origin.track = ch.track
-
-        charmhub = self.model.charmhub
-        charm_resources = await charmhub.list_resources(charm_name)
 
         # Resolve the given charm URLs with an optionally specified preferred channel.
         # Channel provided via CharmOrigin.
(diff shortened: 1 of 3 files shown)


juju/application.py
```diff
@@ -706,19 +701,23 @@ async def refresh(
         if parsed_url.schema is None:
             raise JujuError(f'A ch: or cs: schema is required for application refresh, given : {str(parsed_url)}')
 
-        if revision is not None:
-            origin.revision = revision
+        current_origin = charm_url_origin_result.charm_origin
```
cderici, line 704: because of this move, the parse above might fail for the local urls. I believe that's the reason for [this failure](https://github.com/juju/python-libjuju/actions/runs/9778130375/job/26994335382?pr=1067#step:6:462).

juju/application.py
```diff
@@ -706,19 +701,23 @@ async def refresh(
...
-        if channel:
-            ch = Channel.parse(channel).normalize()
-            origin.risk = ch.risk
-            origin.track = ch.track
+        if channel is not None:
+            channel = Channel.parse(channel).normalize()
 
-        charmhub = self.model.charmhub
-        charm_resources = await charmhub.list_resources(charm_name)
+        origin = client.CharmOrigin(
```
cderici, line 713: let's move this out into a helper function and have some unit tests against this to protect the logic for any regressions.

tests/integration/test_application.py
```diff
@@ -286,6 +286,18 @@ async def test_local_refresh():
                                             base=client.Base("20.04", "ubuntu"))
 
 
+@base.bootstrapped
+@pytest.mark.asyncio
+async def test_refresh_revision():
+    async with base.CleanModel() as model:
+        app = await model.deploy('juju-qa-test', channel="latest/stable", revision=23)
+        # NOTE: juju-qa-test revision 26 has been releases to this channel
```
cderici, line 294: typo

juju/application.py
```diff
@@ -706,19 +701,23 @@ async def refresh(
...
+            source=str(Source.CHARM_HUB),
+            risk=channel.risk if channel else current_origin.risk,
+            track=channel.track if channel else current_origin.track,
+            revision=revision or current_origin.revision,
+            base=current_origin.base,
+            architecture=current_origin.architecture or DEFAULT_ARCHITECTURE,
```
cderici, line 719: I'm not sure if the architecture is gonna be set every time, so you might wanna 

```suggestion
            architecture=current_origin.get('architecture', DEFAULT_ARCHITECTURE),
```

cderici (changes requested):
Looks good, couple of changes needed, and all the tests related to this need to pass.

tests/integration/test_application.py
jack-w-shaw: Fixed

juju/application.py
jack-w-shaw: Fixed

juju/application.py
jack-w-shaw: What logic exactly are you referring to here?

juju/application.py
jack-w-shaw: Good spot. The order of things seems to lack any real order here. I'll move things around a little

juju/application.py
cderici: More like decisions I meant to say, we make a bunch of decisions here, it would be nice if we have unit tests that assert those decisions so if any of them are violated in the future we'd catch it before it breaks.

```
            source=str(Source.CHARM_HUB),
            risk=channel.risk if channel else current_origin.risk,
            track=channel.track if channel else current_origin.track,
            revision=revision or current_origin.revision,
            base=current_origin.base,
            architecture=current_origin.get('architecture', DEFAULT_ARCHITECTURE),
```

jack-w-shaw: Done

cderici (approved):
Looks very good, thanks for adding the extra unit tests 👍 

We'll land it as soon as the check job is fixed.

Issue #1061, opened 2024-06-15
Cant' connect using macaroons or bakery_client


Maharacha:
### Description

I'm trying to connect to a controller for which I'm using Ubuntu One authentication. I'm following [this](https://pythonlibjuju.readthedocs.io/en/latest/narrative/model.html#connecting-with-authentication), trying both with macaroons and with `bakery_client`. In both cases I get:
`ValueError: Some authentication parameters are required : password`

Using macaroons
```
    await model.connect(macaroons=[
        {
            "Name": "macaroon-identity",
            "Value": "abc123",
            "Domain": "api.jujucharms.com",
            "Path": "/identity/",
            "Secure": False,
            "HostOnly": True,
            "Expires": "2024-06-22T13:46:43Z",
        },
    ])
```

or bakery
```
    bakery_client = Client()
    bakery_client.cookies = FileCookieJar('/home/foobar/.local/share/juju/cookies/mycontroller.json')
    model = Model()
    await model.connect(bakery_client=bakery_client)
```

What am I doing wrong?

### Urgency

Casually reporting

### Python-libjuju version

3.3.1.1

### Juju version

3.3.4

### Reproduce / Test

```python
Using macaroons

    await model.connect(macaroons=[
        {
            "Name": "macaroon-identity",
            "Value": "abc123",
            "Domain": "api.jujucharms.com",
            "Path": "/identity/",
            "Secure": False,
            "HostOnly": True,
            "Expires": "2024-06-22T13:46:43Z",
        },
    ])
```

or bakery
```
    bakery_client = Client()
    bakery_client.cookies = FileCookieJar('/home/foobar/.local/share/juju/cookies/mycontroller.json')
    model = Model()
    await model.connect(bakery_client=bakery_client)
```
```

erik78se:
We're still hit by this. Also, normal connect isn't working. See: https://github.com/juju/python-libjuju/issues/1081

PR #1072, opened 2024-07-12
fix: macaroon connection issue

Aflynn50:
#### Description

This fixes issue #1061 which fails because we check for a username/password on a macaroon connection. The code is changed to no longer do this for macaroons.

#### QA Steps

QA is a little difficult for this PR. I have not been able to http macaroon bakery to work with python3.12 which I have installed. In fact it seems connecting to a model in the regular way is broken with python3.12. I will take a look at this soon.


benhoyt:
@Aflynn50 Can you please get review from the Juju team on this and merge (or close if not needed)?

PR #1091, opened 2024-08-23
Fixing cookie based authentication

### LINKED ISSUE #1061: Cant' connect using macaroons or bakery_client

awnns:
#### Description

At present, cookie based authentication does not work in pylibjuju. This is due to a number of bugs. This PR intends to fix many of these issues so that pylibjuju can use the juju client's cookies for authentication.

Fixes:
* The library now detects the appropriate cookie domain from jujudata.
* The library now adds correct user tags for external and local users.
* The library now correctly detects when a discharge is required due to third party caveats.
* The library now attempts to use cookies if no password has been provided and cookies are available.
* A subtle implementation difference between juju's cookie jar and `http.cookiejar.DefaultPolicy` was fixed with monkey patching.

I have also cherry picked a commit from @Aflynn50 from a different ongoing PR.

Fixes: #1061 #1081

#### QA Steps

I have had some difficulty with running the tests locally. I am not sure what sort of setup they want to run in.

What I have tested manually is the following:

* Creating connections and logging into controllers with username and password
* Creating connections and logging into controllers with an external user and identity provider using cookies from the juju client.
* Creating connections and logging into controllers with a local user using cookies from the juju client.
* Various commands such as listing models and checking model statuses.

Things I have been unable to test that are affected by this PR:

* Anything to do with proxied controllers.
* Anything to do with controllers with proper certificates (all I have access to are self signed). There is potential for issues with the cookie domain for these since I have not tested whether it works. It might require trivial modifications.
* Juju 2.X

#### Notes & Discussion

The monkey patches in `gocookies.py` are not a good look and they are somewhat brittle, but I believe it will take a lot more work to fix the `CookiePolicy` if we wish to retain all the other sane behaviors from `http.cookiejar.DefaultCookiePolicy`. An alternative is to use an empty `CookiePolicy` and subclass it to only check for domain equality and path. The problem the monkey patch fixes is caused by incompatible cookiejars. The juju client uses https://github.com/juju/persistent-cookiejar which permits domain fields without dots under the condition that the domain matches exactly with the request's host. Python's implementation instead rewrites the request's host by appending ".local", which makes the comparison fail. 

There are still issues with CA certificates when requiring discharges from the juju controller. This is due to intricacies of the HTTP handling in pymacaroonbakery. It may require substantial changes in pymacaroonbakery, but due to a number of bugfixes in this PR it is not essential for most uses (self signed external identity providers are the exception). The lack of a fix does mean that if a user is not logged in on a local user, and tries to connect to the controller, then they will be met with the wrong exception. The intended outcome is to require an interaction to enter a password, but this is not implemented in pymacaroonbakery. In any case, allowing self signed certificates requires either a hack where pylibjuju sets an environment variable and writes a cacert to a temporary file, or making changes in pymacaroonbakery.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -256,7 +256,8 @@ class Connection:
             specified_facades=None,
             proxy=None,
             debug_log_conn=None,
-            debug_log_params={}
+            debug_log_params={},
+            cookie_domain=None,
     ):
         """Connect to the websocket.
 
@@ -285,8 +286,10 @@ class Connection:
             to prevent using the conservative client pinning with in the client.
         :param TextIOWrapper debug_log_conn: target if this is a debug log connection
         :param dict debug_log_params: filtering parameters for the debug-log output
+        :param str cookie_domain: Which domain the controller uses for cookies.
         """
         self = cls()
+        self.cookie_domain = cookie_domain
         if endpoint is None:
             raise ValueError('no endpoint provided')
         if not isinstance(endpoint, str) and not isinstance(endpoint, list):
@@ -301,7 +304,7 @@ class Connection:
             if password is not None:
                 raise errors.JujuAuthError('cannot log in as external '
                                            'user with a password')
-            username = None
+
         self.usertag = tag.user(username)
         self.password = password
 
@@ -762,6 +765,7 @@ class Connection:
             'bakery_client': self.bakery_client,
             'max_frame_size': self.max_frame_size,
             'proxy': self.proxy,
+            'cookie_domain': self.cookie_domain,
         }
 
     async def controller(self):
@@ -774,6 +778,7 @@ class Connection:
             cacert=self.cacert,
             bakery_client=self.bakery_client,
             max_frame_size=self.max_frame_size,
+            cookie_domain=self.cookie_domain,
         )
 
     async def reconnect(self):
@@ -871,7 +876,7 @@ class Connection:
             # a few times.
             for i in range(0, 2):
                 result = (await self.login())['response']
-                macaroonJSON = result.get('discharge-required')
+                macaroonJSON = result.get('bakery-discharge-required')
                 if macaroonJSON is None:
                     self.info = result
                     success = True
@@ -962,7 +967,7 @@ class Connection:
             params['credentials'] = self.password
         else:
             macaroons = _macaroons_for_domain(self.bakery_client.cookies,
-                                              self.endpoint)
+                                              self.cookie_domain)
             params['macaroons'] = [[bakery.macaroon_to_dict(m) for m in ms]
                                    for ms in macaroons]
 
--- a/juju/client/gocookies.py
+++ b/juju/client/gocookies.py
@@ -5,8 +5,44 @@ import datetime
 import http.cookiejar as cookiejar
 import json
 import time
-
 import pyrfc3339
+
+
+# There are certain subtle differences between the Go and Python
+# cookie implementations that make them incompatible. Particularly,
+# Juju relies on domains being permitted to be arbitrary strings,
+# while Python will try to convert these internally to local domains.
+# The only reasonably economical solution to this problem is to monkey
+# patch the affected class in anticipation of a good solution.
+
+def better_return_ok_domain():
+    old_method = cookiejar.DefaultCookiePolicy.return_ok_domain
+
+    def inner(self, cookie, request):
+        '''This is a monkey patch for the DefaultCookiePolicy class
+        that allows arbitrary domains to be used in cookies.'''
+        if cookie.domain == request.host:
+            return True
+        else:
+            return old_method(self, cookie, request)
+    return inner
+
+
+def better_domain_return_ok():
+    old_method = cookiejar.DefaultCookiePolicy.domain_return_ok
+
+    def inner(self, domain, request):
+        '''This is a monkey patch for the DefaultCookiePolicy class
+        that allows arbitrary domains to be used in cookies.'''
+        if domain == request.host:
+            return True
+        else:
+            return old_method(self, domain, request)
+    return inner
+
+
+cookiejar.DefaultCookiePolicy.return_ok_domain = better_return_ok_domain()
+cookiejar.DefaultCookiePolicy.domain_return_ok = better_domain_return_ok()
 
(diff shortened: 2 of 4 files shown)


dimaqq (commented):
Code looks sensible.

Unit tests would be great, unless the existing unit test actually cover this change.

I wonder how to validate this change against real Juju.

juju/client/gocookies.py
```diff
@@ -5,10 +5,46 @@
 import http.cookiejar as cookiejar
 import json
 import time
-
 import pyrfc3339
 
 
+# There are certain subtle differences between the Go and Python
```
Aflynn50: Rather than monkey patching here, would it be possible to get the inherit from the [DefaultCookiePolicy](https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.DefaultCookiePolicy), override the relevent methods then pass it as a [policy kwarg](https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.FileCookieJar) to `cookiejar.FileCookieJar`?

I believe this should achive the same effect without monkey patching the code.

juju/client/connection.py
```diff
@@ -962,7 +967,7 @@ async def login(self):
             params['credentials'] = self.password
         else:
             macaroons = _macaroons_for_domain(self.bakery_client.cookies,
-                                              self.endpoint)
```
Aflynn50, line 965: I'm not sure I fully get the intent behind this from inspection. What is the difference here between the endpoint and the cookie domain?

Aflynn50 (commented):
Thank you for the contribution! It looks good and it'll be great to get this working again.

As @dimaqq mentioned, it would be good to get some tests working for this. There are some disabled tests in `/tests/integration/test_macaroon_auth.py` that should work again with this fix. Would you be able to get the relevant ones from there passing and remove the skip decorators on them?

juju/client/connection.py
awnns: This is because when you are using macaroons from the bakery client, you are actually using cookies that were originally put in the cookie jar by the juju client. The juju client does not use the endpoint as the cookie domain. Instead it uses either the controller's UUID (if the controller is self signed) or its SNIHostName. If we don't pick the correct cookie domain then pylibjuju will not find the cookie and it will fail to authenticate with the controller. I believe this is because if you have a self signed HA controller (for example) then there will be three different potential endpoints, so if you use that for your cookie domain, then your coookie will only work with one of the endpoints. As an example: you might have an endpoint which is 192.168.3.14, but the cookie in the jar has the domain "791fa163-16bd-4726-9c58-7c360cc133b2".

juju/client/gocookies.py
awnns: There was some complication which I have forgotten. I can look into it again later today.

juju/client/connection.py
Aflynn50: Ah fair, good work on the digging! Seems like a bug that this wasn't implemented the same way then.

SimonRichardson (commented):
@wallyworld can you look over this PR, because you did a lot of work around macroons in the go side.

juju/client/connection.py
```diff
@@ -871,7 +876,7 @@ async def _connect_with_login(self, endpoints):
             # a few times.
             for i in range(0, 2):
                 result = (await self.login())['response']
-                macaroonJSON = result.get('discharge-required')
+                macaroonJSON = result.get('bakery-discharge-required')
```
wallyworld, line 879: For the casual observer - newer Juju controllers (2.9 onwards) use the "bakery-discharge-required" attribute.

juju/client/connection.py
wallyworld: Correct, the juju code is 
```
cookieURL := api.CookieURLFromHost(api.PerferredHost(info))
info.Macaroons = httpbakery.MacaroonsForURL(cookieJar, cookieURL)
```
where PreferredHost is either the SNIHostName or controller UUID.

wallyworld (commented):
Code looks ok. Left a couple of comments to provide a little bit of extra context from the Juju side. I'll leave it to others to agree/disagree with the trade-offs around the cookie jar behaviour, suffice to say it's progress and if it works from the perspective of "don't let perfect be the enemy of good", then it seems ok as a step forward, especially since upstream Juju changes are out of scope here.

dimaqq:
This PR needs to be rebased since #1186 has been merged.

dimaqq:
Please rebase 🙇🏻

benhoyt:
We'd be a lot more comfortable merging this if there were unit and/or integration tests. We'll close for now, but feel free to reopen if revisiting and adding tests.

PR #1103, opened 2024-09-18
fix(connection): wrap reconnect in a task

### LINKED ISSUE #1102: pylibjuju can't reconnect on python 3.12
### Description

pylibjuju on python 3.12 fails to reconnect when the websocket closes.

### Urgency

### Python-libjuju version

3.5.0.0

### Juju version

3.5.3

### Reproduce / Test

This is because of python 3.11+
-> pylibjuju is passing a coroutine to asyncio.wait, which removed the support for this in 3.11 https://github.com/juju/python-libjuju/blob/7d2107f882e5f7927001004dc2dac74a286740eb/juju/client/connection.py#L659 

python 3.11 asyncio: https://github.com/python/cpython/blob/8b275e77d235b0e813f996173fc26eaa3484737c/Lib/asyncio/tasks.py#L425

python 3.10 asyncio: https://github.com/python/cpython/blob/0c5fc27217525c4e40b4064e6979f467540c2fc8/Lib/asyncio/tasks.py#L377

gboutry:
#### Description

Wrap reconnect coroutine in a task. This is due to a breaking change in python stdlib 3.11. Before 3.11 it only emitted a warning.

Fixes: #1102

All places calling asyncio.wait are already wrapping into a task:
https://github.com/juju/python-libjuju/blob/7d2107f882e5f7927001004dc2dac74a286740eb/juju/model.py#L757
https://github.com/juju/python-libjuju/blob/7d2107f882e5f7927001004dc2dac74a286740eb/juju/utils.py#L179

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -656,7 +656,7 @@ class Connection:
                 # if it is triggered by the pinger, then this RPC call will
                 # be cancelled when the pinger is cancelled by the reconnect,
                 # and we don't want the reconnect to be aborted halfway through
-                await jasyncio.wait([self.reconnect()])
+                await jasyncio.wait([jasyncio.create_task(self.reconnect())])
                 if self.monitor.status != Monitor.CONNECTED:
                     # reconnect failed; abort and shutdown
                     log.error('RPC: Automatic reconnect failed')


james-garner-canonical (commented):
I'm not an expert here but this looks good to me. As noted, all the other calls to `wait`  already use tasks. Likewise, all the other calls to `reconnect` are wrapped (in `ensure_future`), so this seems like an appropriate solution here.

juju/client/connection.py
```diff
@@ -656,7 +656,7 @@ async def rpc(self, msg, encoder=None):
                 # if it is triggered by the pinger, then this RPC call will
                 # be cancelled when the pinger is cancelled by the reconnect,
                 # and we don't want the reconnect to be aborted halfway through
-                await jasyncio.wait([self.reconnect()])
+                await jasyncio.wait([jasyncio.create_task(self.reconnect())])
```
benhoyt, line 659: How is this different from just `await self.reconnect()`? It looks like `asyncio.wait()` is supposed to be used when you're waiting for multiple concurrent tasks. I'm not at all up on `asyncio` -- maybe @dimaqq could chime in too?

james-garner-canonical: True, I just assumed that `asyncio` was being used for a good reason, but I'm not super familiar with it either. Perhaps related to the comment above this line?
```python
# the reconnect has to be done in a separate task because,
# if it is triggered by the pinger, then this RPC call will
```

gboutry: This comment and call have been introduced in https://github.com/juju/python-libjuju/pull/148/files.

From the comment, we can infer that the caller wants the coroutine `reconnect` to be continue even if the `rpc` method is itself cancelled.

One of the side effect of the previous implementation of `asyncio.wait` was shielding the coroutine from the caller cancellation.

To make this behaviour clearer, we could instead:
```python3
await jasyncio.shield(jasyncio.create_task(self.reconnect()))
```

The shield would still yield a CancelledError, but the task inside would continue in the background.

dimaqq:
My 2c:

cancellation safety is a noble goal, but it won't be achieved by shielding this one task alone.

I would propose to keep this PR simple: only wrap coro in a task, which I believe, if a non-breaking change, as in it preserves current behaviour, good or bad.

Let's open a separate issue about cancellation.

juju/client/connection.py
dimaqq: Given that there's a single item in the wait list, wdyt about this instead?

```suggestion
                await self.reconnect()
```

Maybe with a touch-up of the comment above, clarifying that cancellation mid-reconnect is an outstanding issue.

dimaqq (approved):
Either fix will move us forward.

gboutry:
I would rather keep the current fix. It'll be the same as the 2 others places this is called with just the goal of unblocking our functional tests on Noble.

An other PR can come later reworking this more broadly in the library.

PR #1105, opened 2024-09-20
fix: move parse_storage_constraints to the bundle deployer

Aflynn50:
It seems a previous change for allow bundle storage constraints was done at the wrong level and affected regular storage constraints as well. This fix moves the previous change to only affect bundle storage constraints.

#### QA Steps

TODO

### DIFF
--- a/juju/model.py
+++ b/juju/model.py
@@ -29,7 +29,6 @@ from .charmhub import CharmHub
 from .client import client, connector
 from .client.overrides import Caveat, Macaroon
 from .constraints import parse as parse_constraints
-from .constraints import parse_storage_constraint
 from .controller import Controller, ConnectedController
 from .delta import get_entity_class, get_entity_delta
 from .errors import JujuAPIError, JujuError, JujuModelConfigError, JujuBackupError
@@ -2116,7 +2115,7 @@ class Model:
                 devices=devices,
                 dryrun=False,
                 placement=placement,
-                storage={k: parse_storage_constraint(v) for k, v in (storage or dict()).items()},
+                storage=storage,
                 trust=trust,
                 base=charm_origin.base,
                 channel=channel,
@@ -2151,7 +2150,7 @@ class Model:
                 endpoint_bindings=endpoint_bindings,
                 num_units=num_units,
                 resources=resources,
-                storage={k: parse_storage_constraint(v) for k, v in (storage or dict()).items()},
+                storage=storage,
                 placement=placement,
                 devices=devices,
                 attach_storage=attach_storage,
--- a/tests/unit/test_bundle.py
+++ b/tests/unit/test_bundle.py
@@ -1,12 +1,11 @@
 # Copyright 2023 Canonical Ltd.
 # Licensed under the Apache V2, see LICENCE file for details.
 
-import pytest
 import unittest
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


skourta:
+1. This is breaking our [test in OpenSearch](https://github.com/canonical/opensearch-operator/actions/runs/11287647997/job/31397381683#step:32:233). The issue can also be solved by explicitly converting the constraints to their `str` representation in the `parse_storage_constraint` function.

dimaqq (approved):
LGTM, but I'd appreciate a review from someone on the Juju team.

juju/model.py
```diff
@@ -2116,7 +2115,7 @@ async def _deploy(self, charm_url, application, series, config,
                 devices=devices,
                 dryrun=False,
                 placement=placement,
-                storage={k: parse_storage_constraint(v) for k, v in (storage or dict()).items()},
+                storage=storage,
```
wallyworld, line 2118: We should add validation that the storage value passed to `_deploy()` is of the expected format. Outside of the test for bundle processing, there's very little if any test coverage for that calls to `model.deploy()` validate input that I am aware of. You could image it would be easy for the unparsed storage string "ebs,1G,2" etc to get to the wrong place instead of the parsed value. We do have unit tests for `parse_storage_constraint` but I wonder if we should include a couple of different variations of the storage string besides the fully populated '"ebs,100G,1"` in the bundle tests.

tests/unit/test_bundle.py
```diff
@@ -246,19 +253,21 @@ async def test_run_local(self):
...
-                                         storage="storage",
+                                         storage={storage_label: constraints.parse_storage_constraint(storage_constraint)},
                                          devices="devices",
                                          num_units="num_units",
                                          channel="",
                                          charm_origin=ANY)
 
     async def test_run_no_series(self):
+        storage_label = "some-label"
+        storage_constraint = "ebs,100G,1"
```
wallyworld, line 346: Can we increase test coverage to include storage for more than one named storage, ie not just a single "storage-label" but a couple of storages like "data" and "logs" or something.

wallyworld (commented):
Looks good but I'd like to see a little extra validation and test coverage.

juju/model.py
james-garner-canonical: Yeah I saw that there were already unit tests for `constraint.py` so I just grabbed a valid value out of the docs to make the integration tests pass -- previously these tests just checked that the string "storage" made it through to the mock deploy unmodified. I considered using several different values here, but since the purpose of these tests isn't to check that the storage constraint parsing works for different values, as that's validated separately in unit tests, I opted not to

In terms of runtime validation, `model.deploy` expects `storage` to be a dict like `{"label": {count: 1, pool: "pool", size: 1234}}`. It tries to unpack `storage['label']` with the `**` operator, so passing a dict like `{"label": "ebs,1G,2"}` will fail at runtime with a `TypeError`, which luckily is an appropriate error here. After this, the types of the `count`, `pool`, and `size` values will be validated by `juju.client._definitions.Constraints`.

I'll add some extra unit tests to specifically cover variations on the storage args for `bundle.AddApplicationChange`. Maybe we can open a separate issue about test coverage / runtime validation for `model.deploy`.

Issue #1108, opened 2024-09-23
Integration tests seem to be non-deterministic and regularly fail on main


james-garner-canonical:
EDIT: See [my comment below](https://github.com/juju/python-libjuju/issues/1108#issuecomment-2387154292) for a table of current test failures on main.

### Description

When trying to fix an issue with test failures run against PRs (e.g. #1088 identifies a breakage with `/merge`), I had multiple integration test failures from a relatively simple PR against main (#1106 pinning a dependency to the version before a recent release).

---

To troubleshoot this, I made a simpler PR against main (#1107 editing CONTRIBUTORS), which also has integration test failures.

Here is a table showing the number of times each test failured over 4 runs of the integration tests on these two 2 PRs.

| test | #1106 | #1107 |
| --- | --- | --- |
| test_app_relation_destroy_block_until_done | 4 | 4 |
| test_deploy_bundle_with_multiple_overlays_with_include_files | 4 | 4 |
| test_deploy_bundle_with_overlay_as_argument | 4 | 4 |
| test_wait_for_idle_more_units_than_needed | 3 | 3 |
| test_upgrade_local_charm - juju... | 2 | 4 |
| test_wait_for_idle_with_not_enough_units | 2 | 2 |
| test_unit_annotations - asyncio.excep... | 2 | 0 |
| test_action - juju.errors.JujuA... | 1 | 1 |
| test_upgrade_local_charm_resource | 0 | 2 |
| test_attach_resource - asyncio.except... | 0 | 1 |

It would be ideal if the tests were deterministic.

Short of fixing the tests themselves, it would be good if the current state of the tests was prominently documented in contributing guidelines -- which test failures are likely to just be the tests being flaky and shouldn't block a merge. A separate issue can then be opened against that list to fix the individual tests.

### Urgency

Blocker for our release

### Python-libjuju version

main?

### Juju version

the version the github workflow is using (doesn't look like versioning for juju etc is printed during test setup)

### Reproduce / Test

Run integration tests on `main`.

dimaqq:
I'm running into a new frequently (or always) failing test:
`FAILED tests/integration/test_charmhub.py::test_subordinate_false_field_exists`

Maybe it deserves to be added to the list.

james-garner-canonical:
commit=50b42d013aee01536416e6334f99443f2b4f1e4c
n_jobs=50
n_failing_tests=26

| path | test | # jobs | % jobs |
| --: | :-- | --: | --: |
[...]

Issue #1099, opened 2024-09-16
Incorrect schema is used to generate the clients


SimonRichardson:
### Description

Whilst reviewing the schema outputted from Juju, it's become aware that we also output all controller and agent schema, not just the client schema. This was never the intended case. It should have just been the client schema.

Although the methods are there, it shouldn't be possible to interact with the methods using a user. Instead, you would require the right tag and permissions to do so.

With that in mind, it should be possible to drop the amount of Python code generated for all the facades. This might be something that @dimaqq might want to look into doing.

You would have to generate the schema with the additional flag `-facade-group=client`.

https://github.com/juju/juju/blob/1eadc941cf64c3a0652583c8e65d870e7cccf111/Makefile#L494-L504

### Urgency

Casually reporting

### Python-libjuju version

current

### Juju version

3.6.0

### Reproduce / Test

```python
Generate the facade with `-facade-group=client` which should reduce the number of generated Python facades, considerably.
```

james-garner-canonical:
I've been looking into this, and I have some questions.

---

**What changes might need to be made in the juju repo?**

1. Are there consumers of the full (not just client) schema? If so, then we don't want to mess with generating the full schema, which happens on every build, writing to `juju/apiserver/facades/schema.json`.

Maybe we add a `rebuild-client-schema` directive and add that to the build step, writing to `client-schema.json` and using `-facade-group=client`? Or add the directive, but make it the python-libjuju devs' responsibility to call it after a Juju release, instead of just copying the generated schema?

On the other hand, if no one should be using non-client schema, then do we want to just modify `rebuild-client` and bring the new version to the various release branches?


2. If we're going to generate new schemas that only have the client bits (dropping the typical schema size from 1.9mb to ~650kb), I guess we should do that for every current schema in `main` and `2.9` branches of python-libjuju?

Generating schemas for these older releases is made a little bit more painful by the fact that the schema generation code lives in the Juju repo. Would it be good to commit new schema generation code and new schemas to the relevant Juju branches? Alternatively, we could just generate the client only schemas without committing any changes to Juju.

---

**Is python-libjuju missing schema updates for recent releases?**

1. It doesn't seem like schemas have been updated in a while. Do they need to be? Should there be a schema file in python-libjuju for every Juju release? Or maybe only when there's a relevant change?

[...]

PR #1113, opened 2024-09-26
feat(client-only-schemas): remove schemas for EOL Juju 3.2

james-garner-canonical:
Remove `3.2.X` schemas. Delete `_client*.py` and run `make client`.

#### Description

Moving towards the goal of having schemas and generated code in python-libjuju only for the latest supported Juju versions (`3.1.9`, `3.3.6`, `3.5.4`, `3.5.3`), and using client only schemas (see #1099), this PR removes schemas for EOL Juju 3.2, and reruns code generations, removing `_client*.py` files and then running `make client`.

#### QA Steps

CI steps should all continue to pass, except for integration testing which should continue to fail with the usual suspects (see #1108 for a non-exhaustive table of tests that sometimes fail on `main`). 

#### Notes

To hopefully simplify the diffs, this is the first PR in a planned sequence of PRs that will depend on each other. Subsequent PRs will be:
1. replace current schemas with latest release client only schemas (`3.1.9`, `3.3.6`) and regenerate code
2. add client only schema for `3.4.5` and regenerate code
3. add client only schema for `3.5.3` and regenerate code


dimaqq (changes requested):
given that _client18 get removed, I'm pretty sure these lines need to be updated:

```py
juju/client/_client.py
26:from juju.client import _client7, _client1, _client3, _client4, _client2, _client17, _client6, _client11, _client10, _client5, _client9, _client18, _client19
41:    "18": _client18,
```

dimaqq:
Funnily enough, `_client12` was never imported in the first place 🚀

james-garner-canonical:
> given that _client18 get removed, I'm pretty sure these lines need to be updated:

Actually, _client18 is the only one that wasn't removed, and was instead modified. So the imports are correct, which is a relief since they're from codegen.

Easy to miss that _client18 is still there in the Files changed tab as the diff is also hidden due to being too large.

dimaqq:
Ah, so true!

dimaqq (approved):

PR #1114, opened 2024-09-27
fix: Application.local_refresh method signature

### LINKED ISSUE #881: [Bug]: Refreshing a charm with local resources doesn't work
### Description

I'm working on an upgrade test for a Kubernetes charm, where I'd like to perform the following steps:
1. Deploy published version of the charm
2. Build charm from source
3. Refresh using the locally built charm

Unfortunately, I haven't been able to use libjuju's `refresh` or `local_refresh` successfully.

Initially, I tried to run this:
```python
# deploy published version of charm [...]
charm_under_test = await ops_test.build_charm(".")
image_path = METADATA["resources"]["tensorboard-controller-image"]["upstream-source"]
await ops_test.model.applications[APP_NAME].refresh(
    path=f"{charm_under_test}", resources={"tensorboard-controller-image": image_path}
)
```
The above produces the following error:
```
FAILED tests/integration/test_upgrade.py::test_upgrade - NotImplementedError: resources option is not implemented
```
Due to the above, I switched to `local_refresh` instead (using the same syntax), which resulted in the following error:
```
FAILED tests/integration/test_upgrade.py::test_upgrade - juju.errors.JujuAPIError: missing base name or channel not valid
```
The full stack trace is this:
```
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/tests/integration/test_upgrade.py", line 81, in test_upgrade
    await ops_test.model.applications[APP_NAME].local_refresh(
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/.tox/integration/lib/python3.8/site-packages/juju/application.py", line 835, in local_refresh
    await app_facade.SetCharm(
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 483, in wrapper
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/.tox/integration/lib/python3.8/site-packages/juju/client/_client13.py", line 1638, in SetCharm
    reply = await self.rpc(msg)
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/.tox/integration/lib/python3.8/site-packages/juju/client/facade.py", line 656, in rpc
  File "/home/phoevos/canonical/kubeflow/tensorboards/charms/tensorboard-controller/.tox/integration/lib/python3.8/site-packages/juju/client/connection.py", line 628, in rpc
juju.errors.JujuAPIError: missing base name or channel not valid
```

Looking at the `local_refresh` signature, I don't get how I could specify the `base` there. BTW, refreshing using the Juju CLI works.

### Urgency

### Python-libjuju version

2.9.42.4

### Juju version

2.9.42

### Reproduce / Test

```python
N/A
```

### LINKED ISSUE #1100: Fix function signature for local_refresh
### Description

The code assumes quite a lot about the arguments.
It seems that there are no users of the function today, because the arguments are kinda broken.

Folks work around by calling juju cli instead, e.g.:

```py
# oidc-gatekeeper-operator/tests/integration/test_charm.py
149-        print("Try to refresh stable charm to locally built")
150-        # temporary measure while we don't have a solution for this:
151-        # * https://github.com/juju/python-libjuju/issues/881
152:        # Currently `application.local_refresh` doesn't work as expected.
```

### Urgency

Casually reporting

### Python-libjuju version

any

### Juju version

any

### Reproduce / Test

```python
pyright
```

dimaqq:
Fixes #1100 

Closes #881 

Charmers already pass a Path to refresh() in integration tests.
This PR standardises that.

Making arguments to local_refresh() explicit, as charmer don't get the default values.

Existing integration test is updated.

Jira: https://warthogs.atlassian.net/browse/CHARMTECH-309

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -4,7 +4,7 @@
 import hashlib
 import json
 import logging
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 from pathlib import Path
 
 from typing_extensions import deprecated
@@ -12,7 +12,7 @@ from typing_extensions import deprecated
 from . import jasyncio, model, tag, utils
 from .annotationhelper import _get_annotations, _set_annotations
 from .bundle import get_charm_series, is_local_charm
-from .client import client
+from .client import client, _definitions
 from .errors import JujuApplicationConfigError, JujuError
 from .origin import Channel
 from .placement import parse as parse_placement
@@ -706,20 +706,28 @@ class Application(model.ModelEntity):
         return await app_facade.SetConstraints(application=self.name, constraints=constraints)
 
     async def refresh(
-            self, channel=None, force=False, force_series=False, force_units=False,
-            path=None, resources=None, revision=None, switch=None):
+            self,
+            channel: Optional[str] = None,
+            force: bool = False,
+            force_series: bool = False,
+            force_units: bool = False,
+            path: Optional[Union[Path, str]] = None,
+            resources: Optional[Dict[str, str]] = None,
+            revision: Optional[int] = None,
+            switch: Optional[str] = None,
+    ):
         """Refresh the charm for this application.
 
-        :param str channel: Channel to use when getting the charm from the
+        :param str|None channel: Channel to use when getting the charm from the
             charm store, e.g. 'development'
         :param bool force_series: Refresh even if series of deployed
             application is not supported by the new charm
         :param bool force_units: Refresh all units immediately, even if in
             error state
-        :param str path: Refresh to a charm located at path
-        :param dict resources: Dictionary of resource name/filepath pairs
-        :param int revision: Explicit refresh revision
-        :param str switch: Crossgrade charm url
+        :param Path|str|None path: Refresh to a charm located at path
+        :param dict[str,str]|None resources: Dictionary of resource name/filepath pairs
+        :param int|None revision: Explicit refresh revision
+        :param str|None switch: URL of a different charm to cross-grade to
 
         """
         if switch is not None and path is not None:
@@ -743,8 +751,17 @@ class Application(model.ModelEntity):
 
         current_origin = charm_url_origin_result.charm_origin
         if path is not None or (switch is not None and is_local_charm(switch)):
-            await self.local_refresh(current_origin, force, force_series,
-                                     force_units, path or switch, resources)
+            local_path = path or switch
+            assert local_path
+
+            await self.local_refresh(
+                charm_origin=current_origin,
+                force=force,
+                force_series=force_series,
+                force_units=force_units,
+                path=local_path,
+                resources=resources,
+            )
             return
 
         origin = _refresh_origin(current_origin, channel, revision)
@@ -870,9 +887,15 @@ class Application(model.ModelEntity):
     upgrade_charm = refresh
 
     async def local_refresh(
-            self, charm_origin=None, force=False, force_series=False,
-            force_units=False,
-            path=None, resources=None):
+            self,
+            *,
+            charm_origin: _definitions.CharmOrigin,
+            force: bool,
+            force_series: bool,
+            force_units: bool,
+            path: Union[Path, str],
+            resources: Optional[Dict[str, str]],
+    ):
         """Refresh the charm for this application with a local charm.
 
         :param dict charm_origin: The charm origin of the destination charm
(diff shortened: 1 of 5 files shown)


juju/application.py
```diff
@@ -668,7 +676,7 @@ async def refresh(
             application is not supported by the new charm
         :param bool force_units: Refresh all units immediately, even if in
             error state
-        :param str path: Refresh to a charm located at path
+        :param Path|str path: Refresh to a charm located at path
```
james-garner-canonical: I think this needs to be documented as Optional, like the type annotation and the logic below.

juju/application.py
```diff
@@ -658,8 +658,16 @@ async def set_constraints(self, constraints):
...
+            self,
+            channel=None,
+            force=False,
+            force_series=False,
+            force_units=False,
+            path: Optional[Union[Path, str]] = None,
+            resources=None,
+            revision=None,
+            switch: Optional[str] = None,
+        ):
```
james-garner-canonical: Worth adding bool annotations for the `=False` cases in this PR?

Tentative idea: if these arguments aren't really `bool`s, maybe they should have a comment to that effect to make cleaning up types easier in future?

juju/application.py
```diff
@@ -822,9 +839,15 @@ async def refresh(
...
-            path=None, resources=None):
+            self,
+            *,
+            charm_origin: _definitions.CharmOrigin,
+            force: bool,
+            force_series: bool,
+            force_units: bool,
+            path: Union[Path, str],
+            resources: Optional[Dict[str, str]],
+        ):
```
james-garner-canonical: I like adding type annotations, but is it ok to remove the default values for these arguments? Are we assuming it's only called from `refresh`? I think `local_refresh` is a public method.

dimaqq: it is a public method, but charmers are unable to use it and shell out instead:

https://github.com/canonical/oidc-gatekeeper-operator/blob/57cde1fb1214c6e4ba749219f3bff67463b89e98/tests/integration/test_charm.py#L149-L162

juju/application.py
dimaqq: It seems to be that the type checkers make an assumption for args that have defaults.

james-garner-canonical: True, I'm just thinking that explicit is better than implicit.

juju/application.py
james-garner-canonical: If no one's using it, and you're intentionally removing the default arguments, then I don't see a problem here. Just checking that they weren't removed in error.

juju/application.py
dimaqq: I've decided to follow the maxim that explicit is better than implicit and typed all the args.

juju/application.py
dimaqq: done, thanks

dimaqq:
squashed and rebased to pass commit lint

james-garner-canonical (approved):
Looks good, thanks. You might need to (re?)sign your commits to merge.

dimaqq:
already signed; also CLA check is passing because it knows I'm @canonical.com

juju/application.py
```diff
@@ -822,9 +839,15 @@ async def refresh(
...
-            path=None, resources=None):
+            self,
+            *,
+            charm_origin: _definitions.CharmOrigin,
+            force: bool,
+            force_series: bool,
+            force_units: bool,
+            path: Union[Path, str],
+            resources: Optional[Dict[str, str]],
+    ):
```
Aflynn50, line 898: Looks good, hopefully the types help clear up the meaning of these args a bit for people

Aflynn50 (commented):
Looks good, could you include the integration test steps in the PR description? Either just which of the pylibjuju integration tests were run, or if local refresh is not used in there, just a code snippet that tests the functionality .

juju/application.py
```diff
@@ -658,20 +658,28 @@ async def set_constraints(self, constraints):
...
+            force_units: bool = False,
+            path: Optional[Union[Path, str]] = None,
+            resources: Optional[Dict[str, str]] = None,
+            revision: Optional[int] = None,
+            switch: Optional[str] = None,
+    ):
         """Refresh the charm for this application.
 
-        :param str channel: Channel to use when getting the charm from the
+        :param str|None channel: Channel to use when getting the charm from the
```
benhoyt, line 721: Do we need the types listed here, now that actual type annotations are present? Are these docstrings actually used to generate reference docs? If so, will it / can it use the type annotations?

dimaqq: I don't actually know. I was thinking to try and see.

juju/application.py
dimaqq: I've updated the existing integration test.

PR #1116, opened 2024-09-27
feat: deprecation warnings on select application attributes

dimaqq:
#### Description

* explicitly export attributes accessible on Application via properties
* mark some as deprecated, as these are unlikely to be available in Juju 4

### DIFF
--- a/juju/application.py
+++ b/juju/application.py
@@ -4,8 +4,10 @@
 import hashlib
 import json
 import logging
-import typing
+from typing import Dict, List, Union
 from pathlib import Path
+
+from typing_extensions import deprecated
 
 from . import jasyncio, model, tag, utils
 from .annotationhelper import _get_annotations, _set_annotations
@@ -24,6 +26,52 @@ log = logging.getLogger(__name__)
 
 
 class Application(model.ModelEntity):
+    """Represents the current state of a deployed application.
+
+    In the current library version, as well entire 2.x and 3.x series,
+    the data is supplied by Juju AllWatcher notifications, also known as deltas
+    for the specific application. The fields are declared here:
+    https://github.com/juju/juju/blob/be8a779/core/multiwatcher/types.go#L191-L209
+
+    The fields marked deprecated below will be removed in version 4.0 because
+    a different API must be used against Juju 4.
+    """
+    @property
+    def name(self) -> str:
+        return self.entity_id
+
+    @property
+    def exposed(self) -> bool:
+        return self.safe_data["exposed"]
+
+    @property
+    @deprecated("Application.owner_tag is deprecated and will be removed in v4")
+    def owner_tag(self) -> str:
+        return self.safe_data["owner-tag"]
+
+    @property
+    def life(self) -> str:
+        return self.safe_data["life"]
+
+    @property
+    @deprecated("Application.min_units is deprecated and will be removed in v4")
+    def min_units(self) -> int:
+        return self.safe_data["min-units"]
+
+    @property
+    def constraints(self) -> Dict[str, Union[str, int, bool]]:
+        return self.safe_data["constraints"]
+
+    @property
+    @deprecated("Application.subordinate is deprecated and will be removed in v4")
+    def subordinate(self) -> bool:
+        return self.safe_data["subordinate"]
+
+    @property
+    @deprecated("Application.workload_version is deprecated and will be removed in v4, use Unit.workload_version instead.")
+    def workload_version(self) -> str:
+        return self.safe_data["workload-version"]
+
     @property
     def _unit_match_pattern(self):
         return r'^{}.*$'.format(self.entity_id)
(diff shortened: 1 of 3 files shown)


juju/application.py
```diff
@@ -24,6 +25,42 @@
 
 
 class Application(model.ModelEntity):
+    @property
+    def name(self) -> str:
```
benhoyt, line 40: Where did these come from previously? Was it falling back to a `__getattr__` in `ModelEntity`?

juju/application.py
```diff
@@ -24,6 +25,42 @@
...
+        return self.safe_data["name"]
+
+    @property
+    def exposed(self) -> bool:
+        return self.safe_data["exposed"]
+
+    @property
+    def owner_tag(self) -> str:
+        warnings.warn("Deprecated", DeprecationWarning)
```
benhoyt: Should this be `stacklevel=2` to point to the caller of `owner_tag`? Also, would a more explicit warning message be helpful? "Application.owner_tag is deprecated" (and can we suggest they use something else?)

tests/unit/test_application.py
```diff
@@ -21,7 +21,7 @@ async def test_expose_with_exposed_endpoints_as_raw_dict(self, mock_conn):
         mock_facade().Expose.return_value.set_result([])
 
         app = Application(entity_id="app-id", model=Model())
-        app.name = "panther"
+        app.model.state.state = {"application": {"app-id": [{"name": "panther"}]}}
```
benhoyt: Is this change required because now these attribute aren't settable? Were they before?

juju/application.py
dimaqq: yes that's correct.
and that in turn was .data that is filled in by all watcher deltas

juju/application.py
dimaqq: Yes I need to make this better.

tests/unit/test_application.py
dimaqq: setting an attribute in user could would have been a mistake, as those were not propagated anywhere. there's no __setattr__.

the test used to set the value directly on the object as a way to monkey-patch the attribute that's needed by the code under test.

arguably it's not a very good test, but doubt that improving tests to represent AllWatcher mechanics better is worthwhile given that we're going to get rid of AllWatcher.

james-garner-canonical (commented):
I had a couple of questions that I accidentally left "pending" all weekend, but thankfully Ben asked them today. Do you think it would be worth describing anything about `Application`'s properties in the class docstring?

dimaqq:
This package already imports typing_extensions via some dependency.

I've made it an explicit dep, allowing for a simpler `@deprecated` decorator.

Here's a test run, showing correct source for the warning:

```py
/code/python-libjuju/example.py:28: DeprecationWarning: Application.owner_tag is deprecated and will be removed in v4
  owner_tag.......... {app.owner_tag!r}
/code/python-libjuju/example.py:30: DeprecationWarning: Application.min_units is deprecated and will be removed in v4
  min_units.......... {app.min_units!r}
/code/python-libjuju/example.py:32: DeprecationWarning: Application.subordinate is deprecated and will be removed in v4
  subordinate........ {app.subordinate!r}
/code/python-libjuju/example.py:34: DeprecationWarning: Application.workload_version is deprecated and will be removed in v4, use Unit.workload_version instead.
  workload_version... {app.workload_version!r}
```

juju/application.py
dimaqq: done

dimaqq:
I see doc changes as orthogonal, so maybe in a separate PR?

dimaqq:
CI failures are unrelated, #1108

james-garner-canonical (commented):
The `@deprecated` decorator is cool, didn't know about that. One or two result in rather long lines, but I think we can look at that later if we introduce linting with `ruff`/etc.

The CI failures are all known failing tests. 3/5 fail 100% of the time on main.
test_wait_for_idle_more_units_than_needed fails 80% of the time on main.
test_subordinate_false_field_exists fails around 25% of the time on main.

The changes made in this PR seem reasonable to me, but I would like the source of truth for the new property names and their type annotations to be mentioned somewhere, if not in a docstring then perhaps somewhere it will show up in the git log.
```
"""Class representing a juju application.

Explicit properties are annotated with data types inferred from ...
See ... juju documentation page?
"""
```
I think I want this because I don't know where they come from or why they're correct, and maybe a future reader will wonder the same thing. But if you don't think this is worthwhile here then let's skip it for now.

Also, maybe we want a `<5` restriction on the `typing_extensions` dependency just to be conservative? I know the existing dependencies are loosely specified, but that already caused at least one CI breakage.

dimaqq:
> The changes made in this PR seem reasonable to me, but I would like the source of truth for the new property names...

Done, ptal

juju/application.py
```diff
@@ -24,6 +26,52 @@
...
+    """Represents the current state of a deployed application.
+
+    In the current library version, as well entire 2.x and 3.x series,
+    the data is supplied by Juju AllWatcher notifications, also known as deltas
+    for the specific application. The fields are declared here:
+    https://github.com/juju/juju/blob/be8a779/core/multiwatcher/types.go#L191-L209
+
+    The fields marked deprecated below will be removed in version 4.0 because
+    a different API must be used against Juju 4.
+    """
```
james-garner-canonical, line 38: Look good to me, thanks for documenting that

james-garner-canonical (approved):

PR #1143, opened 2024-10-07
ci: run integration tests serially

james-garner-canonical:
#### Description

Integration tests are incredibly flakey currently. Perhaps this is related to the parallel execution of the tests. Certainly the fact that the tests are randomly distributed across threads doesn't help with figuring out what works and what doesn't.

#### QA Steps

Run integration tests and see.

PR #1145, opened 2024-10-08
ci: serial integration tests with those that always fail disabled 

dimaqq:
- James's #1143 
- disable `test_deploy_bundle_with_overlay_as_argument`
- disable `test_deploy_bundle_with_multiple_overlays_with_include_files`

### DIFF
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -307,6 +307,7 @@ async def test_deploy_local_bundle_with_overlay_multi():
         assert 'ghost' not in model.applications
 
 
+@pytest.mark.skip(reason="always fails")
 @base.bootstrapped
 @pytest.mark.bundle
 async def test_deploy_bundle_with_overlay_as_argument():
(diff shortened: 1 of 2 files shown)


dimaqq:
1st run:
```
FAILED tests/integration/test_model.py::test_deploy_bundle_local_charms - asy...
FAILED tests/integration/test_model.py::test_deploy_bundle - requests.excepti...
```

2nd run:
```
FAILED tests/integration/test_model.py::test_deploy_bundle_local_charm_series_manifest
```

3rd run:
```
FAILED tests/integration/test_model.py::test_deploy_bundle_local_charms - asy...
```

james-garner-canonical:
While serialising the tests and disabling the ones that always fail seems like it should be an improvement and eliminate a major source of unpredictability in the tests, there still seems to be a long tail of flakiness, with some regular offenders in addition to the ones that always fail. [New table here.](https://github.com/juju/python-libjuju/pull/1143#issuecomment-2399161307)

All the failing tests so far have been from `test_model.py` except for `test_relate_with_offer` from `test_crossmodel.py`. That seems like an important clue, but also maybe we could just put `test_model.py` and `test_crossmodel.py` in a separate job and not require it to pass for now.

benhoyt:
> All the failing tests so far have been from test_model.py

I think this is telling us we should spend some time digging into those test_model failures and trying to get a picture of what's failing and why. There may be similar kinds of failures that we can address in one fell swoop.

james-garner-canonical:
Always failing tests disabled in #1149 
Thanks @dimaqq !

PR #1149, opened 2024-10-09
ci: serialise integration tests and quarantine problematic tests

james-garner-canonical:
Tests in `integration/test_model.py` seem to be flaky even when run serially. All tests in `integration/test_crossmodel.py` are currently skipped, except one which used to be skipped, and is currently flaky even when run serially.

This PR:
* Serialises all integration tests following #1143
* Skips two tests from `test_model.py` that seem to always fail currently, whether run in serial or in parallel, following #1145
* Moves the flaky tests noted above into a separate job, so that the job running the remaining integration tests will hopefully have a shot at succeeding

As a bonus feature, this split of the tests into two runners with `-n 1` seems to be faster than the original method of running all the integration tests in a single runner with `-n auto` (which worked out to be 4 processes on github).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -121,3 +121,29 @@ jobs:
       - name: Run integration
         # Force one single concurrent test
         run: tox -e integration
+
+  integration-quarantine:
+    name: Quarantined Integration Tests
+    needs: [lint, unit-tests]
+    timeout-minutes: 150
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python:
+          - "3.10"
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Setup operator environment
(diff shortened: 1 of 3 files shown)


dimaqq (approved):
That works.

I would rather prefer a custom pytest marker, as that would allow putting entire files as well as individual tests in quarantine.

I suppose that could be changed later just as well.

james-garner-canonical:
> I would rather prefer a custom pytest marker, as that would allow putting entire files as well as individual tests in quarantine.

I can see the value in that.

My thinking is  that this separate job is not really intended as a permanent quarantine zone, but rather we'll eventually fix the model and cross model tests, and then rename the job to 'Integration - model' or something. We'd keep the separate job for easier debugging and faster test runtime. I think 'Quarantined Integration Tests' works as the name for now since contributors will understand when it fails.

A you say, if we go in the direction of wanting to quarantine individual tests in the future we could add a pytest marker in the future.

PR #1150, opened 2024-10-09
fix: make client_facades reflect facades in generated client code

james-garner-canonical:
#### Description

`client_facades` in `juju/client/connection.py` is manually updated. It doesn't reflect the actual facade code present, which was recently generated from the current 3.1.X and 3.3.0 schemas currently checked in. Nor did it reflect the state of the generated code with the previous set of schemas (same + 3.2.X schemas), nor the state of the actually checked in code before the last regeneration of the code (which contained some orphan facades not present in the schemas).

This PR updates `client_schemas` and adds a test to ensure that its facade versions always match the schemas present in the generated client code. The test can be run with `tox -e validate`, and is run in a separate github job as it didn't seem to fit in well as either a unit or integration test. In future we could add more tests to this group to validate other aspects of the codebase related to the code generation.

#### QA Steps

All non-quarantined tests should continue to work.

```
tox -e validate
```
Should pass currently.

#### Notes & Discussion

Does the current state of client_facades make sense? One thing that stood out to me was the addition of `Admin`. I'm wondering if any non-client facades are being added here, since facade code is generated from the full schemas still. If that's the case, then this should be deferred till after we switch to client-only schemas.


tests/validate/test_facades.py
```diff
@@ -0,0 +1,116 @@
...
+class Versions(TypedDict, total=True):
+    versions: List[int]
+
+
+ClientFacades = Dict[str, Versions]
+
+
+@pytest.fixture
+def project_root(request: pytest.FixtureRequest):
+    return request.config.rootpath
```
dimaqq:
```suggestion
def project_root(pytestconfig: pytest.Config) -> Path:
    return pytestconfig.rootpath
```

tests/validate/test_facades.py
```diff
@@ -0,0 +1,116 @@
...
+                f'not {actual}'
+                if actual is not None
+                else 'but is not present'
+            )
+            print(f'    {name!r} {expected_msg} {actual_msg}')
+
+        assert not errors
+
+    @classmethod
+    def generate_client_facades(cls, project_root: Path) -> ClientFacades:
```
dimaqq: function name seems a bit misleading

tests/validate/test_facades.py
```diff
@@ -0,0 +1,116 @@
...
+
+ClientFacades = Dict[str, Versions]
+
+
+@pytest.fixture
+def project_root(request: pytest.FixtureRequest):
+    return request.config.rootpath
+
+
+class TestFacades:
```
dimaqq: Given a dedicated file for this test, I'd drop the class and refactor to module-level tests and fixtures

tests/validate/test_facades.py
```diff
@@ -0,0 +1,116 @@
...
+        errors: List[Tuple[str, Optional[List[int]], Optional[List[int]]]] = []
+        all_names = sorted(set(connection.client_facades).union(good_facades))
+        for name in all_names:
+            expected = self._versions_from_facades(good_facades, name)
+            actual = self._versions_from_facades(client_facades, name)
+            if expected != actual:
+                errors.append((name, expected, actual))
+
+        if errors:
+            print('The following errors were found in connection.client_facades:')
```
dimaqq: Let's refactor `connection.client_facades` and remove the intermediate "versions" key.

Then the test can be reduced to:

```suggestion
        generated: Dict[str, List[int]] = scan_the_client_NN_files()
        
        assert client_facades == generated
```
        
pytest will neatly show the difference is there is any.

dimaqq (changes requested):
Nice!

I think now that we know how to extract and compare, the test can be greatly simplified.

tests/validate/test_facades.py
james-garner-canonical: Happy to remove all this if the pytest output is actually useful. I probably didn't scroll up enough, but I added it because the test failure message for `good_facades == client_facades` didn't look useful to me, as it heavily truncated both dicts.

EDIT: I guess it wasn't useful because of the nested dict with `{'versions': [...]}` ?

james-garner-canonical: > Let's refactor connection.client_facades and remove the intermediate "versions" key.

We *could* do this but I'd rather leave it in till we finish the schema work as I'd also have to refactor my various `client_facades` generation scripts (from `api/facadeversions.go`, from `_clientN.py`, from schema files). I mean it's a pretty simple refactor, but still worth a separate PR I think.

EDIT: If you think it's worth it, we could do that PR before this one.

james-garner-canonical:
Thanks for the review, @dimaqq. The integration test failures aren't the usual ones -- a lot of `KeyError`s like last time I tried changing `client_facades`. I'll investigate further.

---

`await model.deploy(...` seems to be pretty reliably broken -- that's `juju.Model.deploy`.
It breaks on `await self.deploy_types[schema].resolve(...`.
`Model.deploy_types` is defined in `Model.__init__` like so:
```python
self.deploy_types = {
    Schema.LOCAL: LocalDeployType(),
    Schema.CHARM_HUB: CharmhubDeployType(self._resolve_charm),
}
```
We're at least breaking with charmhub charms, given this:
```python
# in juju.bundle
def is_local_charm(charm_url: str):
    return charm_url.startswith('.') or charm_url.startswith('local:') or os.path.isabs(charm_url)
```
And given these example failing lines in tests:
```python
# test_action
app = await model.deploy('juju-qa-test')

# test_get_set_config
app = await model.deploy(
    'ubuntu',
    application_name='ubuntu',
    series='jammy',
    channel='stable',
    config={
        'hostname': 'myubuntu',
    },
    constraints={
        'arch': 'amd64',
        'mem': 256 * MB,
    },
)

# test_deploy_charmhub_charm
app = await model.deploy('ubuntu')

# test_upgrade_local_charm
app = await model.deploy('ubuntu', series='focal', revision=15, channel='latest/stable')
```
So let's assume we mostly just want to investigate `CharmhubDeployType.resolve`

That method breaks on `charm_url, origin = await self.charm_resolver(url, origin, force, series, model_conf)`, where `self.charm_resolver` is just the argument to `CharmhubDeployType.__init__`, so that takes us to `Model._resolve_charm`, which breaks on this line (edited for clarity by me):
```python
supported_series = result.get(
    'supported_series',
    result.unknown_fields['supported-series']  # KeyError!
)
```
So `'supported-series'` isn't in the `result`, nor is it in `result.unknown_fields`.

Summary for the below: we use the `Charms` facade of version 7 instead of `Charms` version 6 -- tests passed with 6. The change is due to listing version 7 in `client_facades` -- previously it would have been in the code, just ignored because of not being listed.

There aren't any errors along the way, but the `result` is missing the expected field, so we really just need to see the result.

---

This result comes from `charms_facade.ResolveCharms` a few lines up, where `charms_facade` is essentially `client.CharmsFacade.from_connection(self.connection())`. `self.connection()` should be a connection to the relevant Juju model, so let's dig into `client.CharmsFacade.from_connection`.

Here it seems we end up in `_client.TypeFactory.from_connection`, which must be a common implementation for getting the correct generated facade. The facade we want is the name of the class minus the `'Facade'` suffix. So we're trying to get the `Charms` facade for a given version.

What version though? We get the target version from `connection.facades.get`, where `connection` is ultimately a `juju.client.connection.Connection`. Its `facades` attribute is set in the `Connection.connect` constructor to an empty `dict`. If `get` returns `None` though, `from_connection` will error out, and that's not happening here, so it must get populated somehow at some point. This is presumably in `Connection._build_facades`, so let's assume that's been called before we get to where we are. `_build_facades` basically iterates over a list of facades returned from the controller, and tries to find a version in common between `client_facades` (or the caller's `specified_facades`) and the facades returned by the controller.

Now we're getting somewhere! Changing `client_facades` is what broke all these tests. If a match isn't found, the version isn't set, and `log.warning` is called -- but that's not what's happening here, because that would result in `None` from `connection.facades.get`, so it seems like we can assume that a match is found between `client_facades` and the facades reported by the controller.

`_client.TypeFactory.from_connection` then uses the version with `lookup_facade`, which strangely also does some fuzzy version matching. It starts at the requested version and steps downwards, seeing if any of the versions down to 1 have the facade, returning the first match. This must succeed here, since it will error if it doesn't find a match, and `from_connection` goes on to instantiate the returned facade class, connect it to the `connection`, and return it.

So we do get the `Charms` facade we're looking for, and that's what `CharmhubDeployType.resolve` uses to call `ResolveCharms`, which we currently have defined in `_client6.py` and `_client7.py`. Which one is being used here?

Well, `client_facades` used to have `'Charms': {'versions': [5, 6]}` but now it has `'Charms': {'versions': [6, 7]}`. Tests pass with `[5, 6]` even though there's no definition in `_client5.py` (or lower), so we must have been using `6`. Since it's broken now, we can probably assume that we're now using `7`.

The only difference in their implementation is the contents of the dictionary they use as the argument to `self.rpc`, and the only difference in those contents is the `version` -- `6` vs `7`.

`Charms.rpc` resolves to `facade.Type.rpc` (manually defined base class for all the generated classes), which just calls `self.connection.rpc` -- that's the same connection that `_client.TypeFactory.from_connection` passed to the `Charms` object after instantiating it.

Heading to `juju.client.connection.Connection.rpc`, we can see that there's a bunch of logic and ways for errors to be raised. But we know that it doesn't raise an error! It just doesn't have a field that `juju.model.Model._resolve_charm` expects, specifically `'supported-series'`. So really we just want to see what's in that result ... (which should probably have been where we first looked anyway ...)

tests/validate/test_facades.py
dimaqq: There you go #1155

dimaqq:
It seems that "supported-series" field was removed after Juju 3.2.3
What does the library need this field for, really?

dimaqq:
Ahh I see your latest commit, it makes sense that "bases" replaced "series".
I'm not sure if the values are the same though.

See:
https://discourse.charmhub.io/t/transition-from-series-to-base-in-juju-4-0/14127
https://juju.is/docs/juju/base

james-garner-canonical:
> Ahh I see your latest commit, it makes sense that "bases" replaced "series". I'm not sure if the values are the same though.
> 
> See:
> https://discourse.charmhub.io/t/transition-from-series-to-base-in-juju-4-0/14127
> https://juju.is/docs/juju/base

Yeah I didn't think it would work, but figured it was worth seeing what would happen. Thanks for finding that documentation.

dimaqq:
How about this:

special-case that one specific facade, so that the rest of the PR can be refactored, rebased on top of #1155 and merged.

open an issue to resolve series vs bases. I think a short discussion with someone from Juju is on order, depending on e.g. what range of Juju versions the library should support.

dimaqq:
https://github.com/juju/python-libjuju/issues/1156

juju/client/connection.py
```diff
@@ -22,84 +22,82 @@
...
-    'Charms': {'versions': [5, 6]},
-    'Cleaner': {'versions': [2]},
-    'Client': {'versions': [5, 6]},
-    'Cloud': {'versions': [1, 2, 3, 4, 5, 7]},
-    'Controller': {'versions': [9, 11]},
-    'CrossModelRelations': {'versions': [1, 2]},
-    'CrossModelSecrets': {'versions': [1]},
-    'CrossController': {'versions': [1]},
-    'CredentialManager': {'versions': [1]},
-    'CredentialValidator': {'versions': [1, 2]},
```
james-garner-canonical, line 51: This is in a big chunk because of sorting -- all the `CAAS` facades were sorted at the end of the Cs, but should be at the start.

james-garner-canonical:
I believe integration tests are passing now, they're just rerunning from the latest minor push. Please take a look, @dimaqq. I think I need to document `excluded_facades` somewhere, but not sure where would be best.

tests/validate/test_facades.py
```diff
@@ -0,0 +1,82 @@
...
+    facades_by_version: Dict[int, Set[str]] = {}
+    # {facade_version: {facade_name, ...}, ...}
+    for version, file in files_by_version:
+        module = importlib.import_module(f'juju.client.{file.stem}')
+        facades = {
+            name.removesuffix("Facade")
+            for name in dir(module)
+            if not (name.startswith('_') or name in ignore)
+        }
+        facades_by_version[version] = facades
```
dimaqq: Would it be better to both distinguish classes that are facades and get it's versions here?

```py
  class MachinerFacade(Type):
      name = 'Machiner'
      version = 5
```

the .version attr?

james-garner-canonical: True, here I'm just assuming that the version in the filename is accurate. It would make sense to check that. We could check that the name matches as well. I'm guessing they have to match because of how codgen works, but we could add tests for this to make it easier to debug if something changed with that and breaks what we're trying to test here.

I wonder if you're also asking about building the `client_facades` mapping from facade name to version numbers in this pass instead of this intermediate mapping from version numbers to facades. I think you're right here too, this would be much more direct.

dimaqq: Sorry I didn't mean to suggest extra checks.

I was proposing to rely on codegen being correct and to simplify stuff, in effect to remove checks.

dimaqq: High level comment:

It's pretty common to write complex code and extra tests while developing something, because we're dipping into the unknown and learning how it actually works. It's fine to have fancy stuff in the branch history.

When the PR is getting ready to be merged, I would normally expect that things get simplified / streamlined and/or more understandable option chosen. I'd rather merge polished, concise code.

tests/validate/test_facades.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_class_name_matches_facade_name(project_root: Path) -> None:
+    ignore_names = dir(_definitions)
+    for file in (project_root / 'juju' / 'client').glob('_client[0-9]*.py'):
+        module = importlib.import_module(f'juju.client.{file.stem}')
+        for cls_name in dir(module):
+            if cls_name.startswith('_') or cls_name in ignore_names:
+                continue
+            assert getattr(module, cls_name).name == cls_name.removesuffix('Facade')
```
dimaqq: Is there a mismatch precedent?

If there is not, let's remove these tests.

tests/validate/test_facades.py
```diff
@@ -0,0 +1,74 @@
...
+    client_facades = cast(ClientFacades, connection.client_facades)
+    expected_facades = make_client_facades_from_generated_code(project_root)
+    assert {
+        k: v['versions'] for k, v in client_facades.items()
+    } == {
+        k: v['versions'] for k, v in expected_facades.items()
+    }
+
+
+def make_client_facades_from_generated_code(project_root: Path) -> ClientFacades:
```
dimaqq:
```suggestion
@pytest.fixture
def generated_code_facades(project_root: Path) -> ClientFacades:
```

wdyt?

tests/validate/test_facades.py
```diff
@@ -0,0 +1,74 @@
...
+    """Return a client_facades dictionary from generated code under project_root.
+    """
+    ignore_names = dir(_definitions)
```
dimaqq: I'm not a fan of this... I think that facade classes are quite different from definitions in the first place.

tests/validate/test_facades.py
james-garner-canonical: Yeah I'm glad I checked that it matches, but it should by definition, so I'll remove these.

tests/validate/test_facades.py
james-garner-canonical: That makes sense, thank you

tests/validate/test_facades.py
james-garner-canonical: Sounds good

tests/validate/test_facades.py
```diff
@@ -0,0 +1,74 @@
...
+    facades: Dict[str, List[int]] = {}
+    for file in (project_root / 'juju' / 'client').glob('_client[0-9]*.py'):
+        module = importlib.import_module(f'juju.client.{file.stem}')
+        for cls_name in dir(module):
+            if cls_name.startswith('_') or cls_name in ignore_names:
+                continue
+            cls = getattr(module, cls_name)
+            if cls.version in excluded_facades.get(cls.name, []):
+                continue
+            facades.setdefault(cls.name, []).append(cls.version)
```
dimaqq: Maybe I'm splitting hairs, but I feel this is still a tad too complex.

See this:

```py
In [8]: bad = juju.client._definitions.__dict__.values()

In [9]: good = juju.client._client2.__dict__.values()

In [11]: for thing in good:
    ...:     try:
    ...:         thing.version
    ...:         thing.schema
    ...:     except AttributeError:
    ...:         pass  # not a facade
    ...:     else:
    ...:         print(thing.__name__)

In [12]: for thing in bad:
    ...:     try:
    ...:         thing.version
    ...:         thing.schema
    ...:     except AttributeError:
    ...:         pass
    ...:     else:
    ...:         print(thing.__name__)
    ...:
[no output]
```

I would fudge/filter (ignore/excluded) after getting a full list.

james-garner-canonical: I'm ok with using duck typing to check for Facade classes in the `_clientN.py` modules instead of filtering out their imports, if you prefer that, replacing this:
```py
            if cls_name.startswith('_') or cls_name in ignore_names:
                continue
```
with this:
```py
            try:
                cls.name
                cls.version
            except AttributeError:
                continue
```

We could also switch from filtering out items in the loop like this:
```py
            if cls.version in excluded_facades.get(cls.name, []):
                continue
```
with filtering in a second loop, something like this:
```py
    for facade, versions in facades.items():
        excluded_versions = excluded_facades.get(facade, [])
        versions[:] = [
            version for version in versions
            if version not in excluded_versions
        ]
```
Let me know what you'd prefer or if you had something else in mind.

dimaqq (approved):

tests/validate/test_facades.py
```diff
@@ -0,0 +1,51 @@
...
+    return pytestconfig.rootpath
+
+
+@pytest.fixture
+def generated_code_facades(project_root: Path) -> ClientFacades:
+    """Return a client_facades dictionary from generated code under project_root.
+    """
+    facades: Dict[str, List[int]] = defaultdict(list)
+    for file in (project_root / 'juju' / 'client').glob('_client[0-9]*.py'):
+        module = importlib.import_module(f'juju.client.{file.stem}')
```
dimaqq:
```suggestion
    for file in project_root.glob("juju/client/_client*.py"):
        module = importlib.import_module(f'juju.client.{file.stem}')
```

As `.stem` already returns file name without dirs or suffix.

james-garner-canonical: Sorry to ask a bunch of questions about this minor suggestion, but I was hoping you could elaborate on a couple of things:
1. Do you prefer `project_root.glob('juju/client/...` over `(project_root / 'juju' / 'client').glob('...` for readability reasons? I prefer the format I used because it seems better to me to do operations explicitly outside of strings where possible, but I'm happy to change this.
2. Do you prefer globbing for `'_client*.py'` rather than `'_client[0-9]*.py'` for terseness, or did you explicitly want to change the glob behaviour? I used `[0-9]` to exclude `_client.py` which doesn't contain any facades. It should still work since we use duck typing to check for facade types. It initially seemed more correct to me to just explicitly check the numbered files, but do you prefer to include `_client.py`?

dimaqq: I suppose both, though weakly.

Maybe it's just me, but I tend to think of glob like shell expansion, thus preferring what I see in the shell (and maybe an IDE) over everything-is-an-object.

Same goes for `[0-9]`, like what does this do? how does this cover `_client18`, oh, there's a `*` but not a `+` because it's glob and not regex...

In other words, I think that `project_root.glob("juju/client/_client*.py")` is easier to understand, especially for someone less familiar with the codebase.

james-garner-canonical: Makes sense, thanks for explaining. The `[0-9]*` to catch `18` bugged me too, and separately I've been bitten by a glob like that matching too much anyway, so I'll follow your suggestions here.

PR #1151, opened 2024-10-09
ci: use apt instead of pip to install tox

james-garner-canonical:
#### Description

Github's rollout of 24.04 for ubuntu-latest has now reached python-libjuju, so `pip install tox` no longer works -- 24.04 prevents installing packages using pip.

This PR switches all our `run: pip install tox` lines in github actions to `run: sudo apt install tox`.

#### QA Steps

Tests no longer fail when trying to install tox, as they started doing in the last 12 hours.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install dependencies
-        run: apt install tox
+        run: sudo apt install tox
       - name: Run linter
         run: |
           tox -e lint
@@ -60,7 +60,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install dependencies
-        run: apt install tox
+        run: sudo apt install tox
       - name: Run unit tests
         run: tox -e py3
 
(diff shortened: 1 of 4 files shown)


james-garner-canonical:
Oh looks like this is the wrong approach as the attempt to install tox happens in an earlier step anyway.
[test.yaml](https://github.com/juju/python-libjuju/blob/1f5d36114aebb06728b628b5e3da44e29ede4502/.github/workflows/test.yaml#L134-L141)
```
      - name: Setup operator environment
        uses: charmed-kubernetes/actions-operator@main
        with:
          provider: lxd
          juju-channel: 3.4/stable
```

[Test output](https://github.com/juju/python-libjuju/actions/runs/11263529539/job/31321644719)
```
Run charmed-kubernetes/actions-operator@main
Install core snap
Install LXD
Initialize LXD
Install tox
  /usr/bin/which tox
  /usr/bin/which pip
  /usr/bin/pip
  pip is available, install tox
  /usr/bin/pip install tox
  error: externally-managed-environment
```
So we can probably remove our manual installation of tox altogether. And the correct fix seems to be running our setup python step before running setup operator environment -- see [here](https://github.com/charmed-kubernetes/actions-operator/issues/82#issuecomment-2368992683).

PR #1152, opened 2024-10-09
ci: run setup-python before actions-operator to fix tox install

james-garner-canonical:
#### Description

Github's rollout of `24.04` for `ubuntu-latest` has now reached `python-libjuju`, so `pip install tox` no longer works -- `24.04` prevents installing packages using `pip`. This causes an error in 'Setup operator environment' using `charmed-kubernetes/actions-operator@main`. The fix is to run our 'Setup Python' step using `actions/setup-python` before `actions-operator` instead of after.

Since this takes care of installing `tox`, we can remove `pip install tox` for the jobs using these steps.

Also bump `setup-python` to `v5` for integration-quarantine, matching the recent PR bumping the version for other jobs.

#### QA Steps

Tests no longer fail when trying to install `tox`, as they started doing in the last 12 hours.

#### Notes

Fix documented in charmed-kubernetes/actions-operator [here](https://github.com/charmed-kubernetes/actions-operator/issues/82#issuecomment-2368992683).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -79,6 +79,10 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v4
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
       - name: Setup operator environment
         uses: charmed-kubernetes/actions-operator@main
         with:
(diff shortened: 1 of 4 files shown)


dimaqq (approved):

james-garner-canonical:
I don't think we noticed the significance of this at the time, but we got our first green CI tick here -- the quarantined integration tests suddenly pass!

It's unexpected but makes sense that they could be affected here, since this is a version bump for their actions afterall.

james-garner-canonical:
But the first commit also passes with a green tick, and that's the one where I forgot to bump the versions ...

Issue #1156, opened 2024-10-10
Support Juju 3.3+ bases instead of or in addition to series


dimaqq:
### Description

Older Juju used "series" like "focal"
New Juju uses "bases" like "ubuntu@24.04"

See:
https://discourse.charmhub.io/t/transition-from-series-to-base-in-juju-4-0/14127
https://juju.is/docs/juju/base

### Urgency

It would save from a weird workaround

### Code I'd Like to Run

```python
...
```

dimaqq:
Need to decide how we do this:
 - drop Juju 3.2 support and move all code to bases?
 - or dual-wield series and bases depending on juju controller version?

For 3.x series of this library, we have to make a table of series --> bases, like "focal" -> ubuntu@xx.xx

For 4.x series, we could force all users to upgrade, maybe?

james-garner-canonical:
> For 4.x series, we could force all users to upgrade, maybe?

I think this is reasonable -- drop series entirely and only use bases in python-libjuju targeting Juju 4.0.

If no users are requesting support for bases, do we need to support them in 3.x of this library? I know that we probably *should*, but for example Juju 3.6 seems to still supports Charms facade version 6, which still uses series instead of bases.

PR #1158, opened 2024-10-14
test: bump series and pin revision for local charm refresh tests

james-garner-canonical:
#### Description

A recent update to the 'ubuntu' charm requires the use of the v2 metadata format when refreshing from a local charm. The v2 format replaces series with bases, among other changes. Properly supporting this will need to wait on updating python-libjuju to handle bases as well as (or instead of) series -- see issue #1156.

In the meantime, fix the test breakage by pinning the install of 'ubuntu' in broken tests to the previous revision (24). This revision has a mismatch with the specification of the "focal" series (20.04), so also bump the series in the local charm to "jammy".

#### QA Steps

CI tests should now pass again.

### DIFF
--- a/tests/integration/charm/metadata.yaml
+++ b/tests/integration/charm/metadata.yaml
@@ -1,5 +1,5 @@
 name: charm
-series: ["focal"]
+series: ["jammy"]
 summary: "test"
 description: "test"
 maintainers: ["test"]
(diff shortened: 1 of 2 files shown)


james-garner-canonical:
Note: unrelated to this PR, the current state of quarantined integration tests is kinda flaky. On this PR, one test failed on the first run; second run, another test failed; third run, all passed. Non-quarantined tests seem to be more reliable so far.

dimaqq (approved):

james-garner-canonical:
Why didn't this merge?

PR #1159, opened 2024-10-15
style: type annotate constraints.py

james-garner-canonical:
#### Description

While looking into #1105, I found it useful to type annotate `juju/constraints.py` while figuring out how things work. 

#### QA Steps

There should be no runtime changes in behaviour. Unit and integration tests should continue to pass, especially `tests/unit/test_constraints.py`.

#### Notes & Discussion

I also broke the longer regexes down across multiple lines and added comments to help me understand them.

I don't mind cutting this PR back to use more minimal type annotations, etc if that's deemed more appropriate in review (e.g. Dict[str, Union[int, float, bool]] instead of the custom TypedDicts).


dimaqq (approved):

dimaqq:
Let's rebase and merge this.

PR #1160, opened 2024-10-15
feat(client-only-schemas): replace existing schemas with client-only versions

james-garner-canonical:
#### Description

Moving towards the goal of having schemas and generated code in `python-libjuju` only for the latest supported Juju versions (`3.1.9`, `3.3.6`, `3.5.4`, `3.5.3`), and using client only schemas (see #1099), this PR replaces the existing `3.1.X` and `3.3.0` schemas with their client-only  counterparts.

#### QA Steps

Tests pass, but is there anything else we should check here?


juju/client/facade.py
```diff
@@ -350,6 +350,8 @@ def buildTypes(schema, capture):
     for kind in sorted((k for k in schema.types if not isinstance(k, str)),
                        key=lambda x: str(x)):
         name = schema.types[kind]
+        if not name:
```
dimaqq, line 353: A comment would be useful here.

dimaqq (approved):

juju/client/facade.py
james-garner-canonical: Yeah you got me thinking and I might try to further debug this weird case where an empty name comes up in the  3.1.0 client-only codegen (not with any other client-only schemas, and not with 3.1.0 full schema).

PR #1163, opened 2024-10-16
ci: integration tests against multiple juju versions

dimaqq:
Integration tests against Juju 3.1, 3.3, 3.4, 3.5 (was: 3.4)
Unit tests against Python 3.8~3.13 (was: 3.9~3.10)

Notes: Juju 3.6 deprecated some bases (?) and many integration tests fail. Given that it's in beta, deferring that to a separate PR.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -188,3 +188,4 @@ jobs:
           juju-channel: ${{ matrix.juju }}
       - name: Run integration
         run: tox -e integration-quarantine
+        continue-on-error: true  # so that other juju versions are tested
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -110,6 +110,7 @@ jobs:
           # * test_ssh
           # * ...
           # - "3.6/beta"
+    continue-on-error: false  # don't ignore if tests against some juju version fail
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -153,6 +154,7 @@ jobs:
       - name: Run integration
         # Force one single concurrent test
         run: tox -e integration
+        continue-on-error: true  # so that other juju versions are tested
 
   integration-quarantine:
     name: Quarantined Integration Tests
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,6 +2,10 @@ name: Testing
 
 on: [push, pull_request, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     name: Linter
@@ -98,6 +102,9 @@ jobs:
         juju:
           - "3.1/stable"
           - "3.3/stable"
+          - "3.4/stable"
+          - "3.5/stable"
+          - "3.6/beta"
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -109,7 +116,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          juju-channel: 3.4/stable
+          juju-channel: ${{ matrix.juju }}
       # 2023-01-11 Commented until we discover a
       # clear approach for this.
       # - name: Set proxy in controller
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -138,6 +158,7 @@ jobs:
       - name: Run integration
         # Force one single concurrent test
         run: tox -e integration
+        continue-on-error: true  # so that other juju versions are tested
```
james-garner-canonical: Am I understanding correctly that `continue-on-error` being `true` here means that a failure won't prevent the other Juju versions from being run, but `continue-on-error=false` at the `integration` level means that a failure will still cause the `integration` job as a whole to be marked as failed?

.github/workflows/test.yaml
```diff
@@ -159,6 +185,6 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          juju-channel: 3.4/stable
+          juju-channel: ${{ matrix.juju }}
       - name: Run integration
         run: tox -e integration-quarantine
```
james-garner-canonical, line 191: Without `continue-on-error` here, will `integration-quarantine` just cancel other Juju versions after the first Juju version failure? Do we want the same `continue-on-error` setup here as in `integration`?

.github/workflows/test.yaml
```diff
@@ -2,6 +2,10 @@ name: Testing
 
 on: [push, pull_request, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
james-garner-canonical, line 7: I read the documentation on this but I'm not 100% sure on the effect. At first I thought it would force all these tests to run sequentially, but reading further I think since `github.workflow` refers to this workflow, 'Testing', while `github.ref` is the branch, that this will cancel the old 'Testing' job for a branch when new commits get pushed, which seems sensible. Is that correct?

james-garner-canonical (commented):
A couple of learning questions from me, and one that might require action -- should `integration-quarantine` have the same setup with `continue-on-error` as `integration` does?

.github/workflows/test.yaml
james-garner-canonical: Yeah looks like the first failure in quarantined blocked the rest of the jobs from running. I think we should have them all run. Sometimes they all pass and it would be good to see if there's any pattern regarding which juju versions they tend to fail or pass on.

.github/workflows/test.yaml
dimaqq: Yes, that's the intention anyway.

.github/workflows/test.yaml
dimaqq: Good point

.github/workflows/test.yaml
dimaqq: yes that's how it works.

.github/workflows/test.yaml
dimaqq: done.

.github/workflows/test.yaml
```diff
@@ -138,6 +158,7 @@ jobs:
       - name: Run integration
         # Force one single concurrent test
         run: tox -e integration
+        continue-on-error: true  # so that other juju versions are tested
 
   integration-quarantine:
```
james-garner-canonical, line 163: I guess we also need `continue-on-error=false` at this level, assuming that having it at this level is correct for `integration`.

james-garner-canonical: The commented on region of code that github chooses made this unclear -- it's referring to the `integration-quaraintine` scope.

james-garner-canonical: To further attempt to be clear, we have:
`integration.continue-on-error = false`
`integration[Run integration].continue-on-error = true`

`integration-quarantine[Run integration].continue-on-error = true`

And assuming that the setup for `integration` is correct, I think we should also have:
`integration-quarantine.continue-on-error = false`

james-garner-canonical (approved):

.github/workflows/test.yaml
dimaqq: done, thanks

PR #1167, opened 2024-10-17
ci: use fail-fast, not continue-on-error

dimaqq:
Looks like I was wrong in #1163 
Using https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast instead.
Skip tests that are guaranteed to fail against older Juju controllers.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -176,7 +176,7 @@ jobs:
           - "3.5/stable"
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -97,6 +97,7 @@ jobs:
     timeout-minutes: 150
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python:
           # We will reduce the workload to 3.10 to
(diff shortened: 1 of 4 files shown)


SimonRichardson (commented):
The postgres charm should have an assumes block in the metadata yaml, which should state that it doesn't work for less than 3.3.0.

dimaqq:
It probably does.
I don't have the error snippet ready, but the error suggested that several charm revisions were considered, but there was no charm revision that could be run.
I've opened https://github.com/juju/python-libjuju/pull/1173 with a single skip disabled to catch the actual detailed error.

The point of this PR is that it doesn't make sense to run a test that uses secrets facade v2 or deploys postgresql against older Juju versions, these tests will always fail.

dimaqq:
Here's the error that is encountered against Juju 3.1:

```
Charm feature requirements cannot be met:
  - charm requires at least one of the following:
    - charm requires all of the following:
      - charm requires feature "juju" (version < 3.0.0) but model currently supports version 3.1.10
    - charm requires all of the following:
      - charm requires feature "juju" (version >= 3.4.3) but model currently supports version 3.1.10
    - charm requires all of the following:
      - charm requires feature "juju" (version >= 3.5.1) but model currently supports version 3.1.10

Feature descriptions:
  - "juju": the version of Juju used by the model

For additional information please see: https://juju.is/docs/olm/supported-features
```

dimaqq:
Limiting this PR to only fixing the CI, without any code changes.

james-garner-canonical (approved):

PR #1168, opened 2024-10-18
fix: don't use lexical sorting for version numbers in codegen

james-garner-canonical:
#### Description

Previously, python-libjuju iterated over schemas keyed by their version string (e.g. `'3.1.9'`) using lexical sorting. For a given facade version, a definition in a higher versioned schema was intended to overwrite any prior definition saved (see `generate_facades` function in `facade.py`). With lexical sorting, `'3.1.10'` would be sorted in between `'3.1.1'` and `'3.1.2'`, which would not lead to the desired behaviour. This commit fixes this problem by using a tuple of integers as the sorting key. A special case is required for the version string `'latest'`, and we use `(9000, 9000, 9000)`.

#### QA Steps

Codegen doesn't change with current schemas.

#### Notes & Discussion

When `Juju 9001` comes out (well, any Juju version > `9000.9000.9000`), we'll need to update the special case for `'latest'`. Actually we can probably just remove the `'latest'` special case if we ever tidy up codegen, as it's not used currently, and is probably just intended as a convenience for testing and development, where it still doesn't even really seem necessary.

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -13,7 +13,7 @@ import typing
 from collections import defaultdict
 from glob import glob
 from pathlib import Path
-from typing import Any, Mapping, Sequence, TypeVar
+from typing import Any, Dict, List, Mapping, Sequence, Tuple, TypeVar
 
 import typing_inspect
 
@@ -926,11 +926,22 @@ def generate_definitions(schemas):
     return definitions
 
 
-def generate_facades(schemas):
+def sortable_schema_version(version_string: str) -> Tuple[int, int, int]:
+    """Return a sorting key in the form (major, minor, micro) from a version string."""
+    # 'latest' is special cased in load_schemas and should come last
+    if version_string == 'latest':
+        return (9000, 9000, 9000)
+    # raise ValueError if string isn't in the format A.B.C
+    major, minor, micro = version_string.split('.')
+    # raise ValueError if major, minor, and micro aren't int strings
+    return (int(major), int(minor), int(micro))
+
+
+def generate_facades(schemas: Dict[str, List[Schema]]) -> Dict[str, Dict[int, codegen.Capture]]:
     captures = defaultdict(codegen.Capture)
 
     # Build the Facade classes
-    for juju_version in sorted(schemas.keys()):
+    for juju_version in sorted(schemas.keys(), key=sortable_schema_version):
         for schema in schemas[juju_version]:
             cls, source = buildFacade(schema)
             cls_name = "{}Facade".format(schema.name)


juju/client/facade.py
```diff
@@ -926,11 +926,22 @@ def generate_definitions(schemas):
...
 
 
-def generate_facades(schemas):
+def sortable_schema_version(version_string: str) -> Tuple[int, int, int]:
+    """Return a sorting key in the form (major, minor, micro) from a version string."""
+    # 'latest' is special cased in load_schemas and should come last
+    if version_string == 'latest':
+        return (9000, 9000, 9000)
+    # raise ValueError if string isn't in the format A.B.C
+    major, minor, micro = version_string.split('.')
```
dimaqq, line 935: wdyt about using what already do for the client and server version:

```
from packaging import version
version.parse(server_version)
```

juju/client/connector.py

james-garner-canonical: Sounds like a good idea, if you want to take care of it to push this over the finish line that would be cool, otherwise I'll hopefully get to it some time during the sprint

juju/client/facade.py
```diff
@@ -926,11 +926,22 @@ def generate_definitions(schemas):
...
+    # raise ValueError if major, minor, and micro aren't int strings
+    return (int(major), int(minor), int(micro))
+
+
+def generate_facades(schemas: Dict[str, List[Schema]]) -> Dict[str, Dict[int, codegen.Capture]]:
     captures = defaultdict(codegen.Capture)
 
     # Build the Facade classes
-    for juju_version in sorted(schemas.keys()):
+    for juju_version in sorted(schemas.keys(), key=sortable_schema_version):
```
dimaqq, line 944: Is the idea here to go from oldest to newest because the latter iteration overwrites the ealier?

james-garner-canonical:
Yes, see the `.clear` call a few lines down

juju/client/facade.py
james-garner-canonical: Email reply isn't as cool as I was hoping. But yeah, I think it does overwrite.
The relevant line is:
```
 captures[schema.version].clear(cls_name)
```

With this PR I'm just slavishly trying to keep the original intent working. But maybe this being technically broken is an opportunity for us to make some bigger changes?

Certainly in terms of implementation it would be nice to get rid of all the inheriting from `dict`, and have those classes just do operations on an internal `dict` instead -- at most implement the specific `dict` methods that are expected to be used and forward them to the internal `dict` rather than inheriting every `dict` method and overriding some of them  ...

james-garner-canonical: Oh and note that `schema.version` is a facade version, like `7`.

`juju_version` is the Juju schema version, like `'3.1.10'`.

juju/client/facade.py
```diff
@@ -926,11 +926,22 @@ def generate_definitions(schemas):
     return definitions
 
 
-def generate_facades(schemas):
+def sortable_schema_version(version_string: str) -> Tuple[int, int, int]:
+    """Return a sorting key in the form (major, minor, micro) from a version string."""
+    # 'latest' is special cased in load_schemas and should come last
+    if version_string == 'latest':
+        return (9000, 9000, 9000)
```
dimaqq, line 933: Why special-case latest?
We don't use files with the "latest" marker, do we?

dimaqq:
Closing in favour of #1174

juju/client/facade.py
james-garner-canonical: Late reply, but I special cased it in sorting because the code that loads in schemas special cases it as a permitted version. There was no need to special casing in sorting when sorting was simply lexicographic, since "latest" would sort after all the numerical strings.

PR #1174, opened 2024-10-22
fix: use version sorting in codegen, not lexical

dimaqq:
Cherry-picked from #1168 and simplified.

Rationale Juju micro versions can get larger than 9, e.g. 2.9.51.

When 3.5.10 comes around, we want it to take precedence over 3.5.9 and not get wedged between 3.5.1 and 3.5.2

Keeping the current codegen mode of operation where it starts with the oldest version and whacks some internal state on encountering a latter version.

(We'll deal with that in a separate PR)

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -13,7 +13,7 @@ import typing
 from collections import defaultdict
 from glob import glob
 from pathlib import Path
-from typing import Any, Mapping, Sequence, TypeVar
+from typing import Any, Dict, List, Mapping, Sequence, Tuple, TypeVar
 
 import typing_inspect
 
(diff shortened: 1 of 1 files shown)


juju/client/facade.py
```diff
@@ -926,11 +927,11 @@ def generate_definitions(schemas):
     return definitions
 
 
-def generate_facades(schemas):
+def generate_facades(schemas: Dict[str, List[Schema]]) -> Dict[str, Dict[int, codegen.Capture]]:
     captures = defaultdict(codegen.Capture)
 
     # Build the Facade classes
-    for juju_version in sorted(schemas.keys()):
+    for juju_version in sorted(schemas.keys(), key=packaging.version.parse):
```
james-garner-canonical, line 934: This is nice. It'll break on the "latest" string, right? If so, let's just remove the special casing in `load_schemas` (line 955). Existing code pasted below for convenient reference.

```python
def load_schemas(options):
    schemas = {}
    for p in sorted(glob(options.schema)):
        if 'latest' in p:
            juju_version = 'latest'
        else:
            try:
                juju_version = re.search(JUJU_VERSION, p).group()
            except AttributeError:
                print("Cannot extract a juju version from {}".format(p))
                print("Schemas must include a juju version in the filename")
                raise SystemExit(1)
```

james-garner-canonical: Done

james-garner-canonical (approved):

PR #1170, opened 2024-10-18
fix: supported series computation needed for juju 3.6

dimaqq:
Covered by existing integration tests

PR #1175, opened 2024-10-22
fix: juju 3.6  draft log traffic

dimaqq:
Companion to #1170 to figure out why https://charmhub.io/juju-qa-test?channel=2.0/stable fails to deploy.

Suspecting that client-side picks 24.04 base when this charm version only supports 22.04


dimaqq:
Juju impl https://github.com/juju/juju/blob/d4e212992bd972af65d7a0f8f8e0331effa59b2f/core/charm/computedbase.go#L47-L71

PR #1179, opened 2024-11-07
chore: bring schemas up to date with juju releases

### LINKED ISSUE #1099: Incorrect schema is used to generate the clients

james-garner-canonical:
#### Description

Finally resolves #1099, adding all the (client-only) schemas from missed releases.

#### QA Steps

Integration tests are still a bit flakey (we seem to get some connection closed type errors on a couple of python versions at the moment), but otherwise tests should all pass.


dimaqq (approved):
No code changes?
sweet!

PR #1182, opened 2024-11-08
feat: use base argument instead of series

### LINKED ISSUE #1156: Support Juju 3.3+ bases instead of or in addition to series

dimaqq:
#### Description

- support both 6 and 7 versions of Charms facade.
- deprecate the `series=` argument
- [to do] introduce `base=` argument

#### QA Steps

- [to do] validate using existing integration tests

Fixes #1156

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -84,12 +84,7 @@ jobs:
           - "3.3/stable"
           - "3.4/stable"
           - "3.5/stable"
-          # A bunch of tests fail with juju.errors.JujuError: base: ubuntu@15.04/stable
-          # * test_subordinate_units
-          # * test_destroy_unit
-          # * test_ssh
-          # * ...
-          # - "3.6/beta"
+          - "3.6/candidate"
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -149,6 +144,7 @@ jobs:
           - "3.3/stable"
           - "3.4/stable"
           - "3.5/stable"
+          - "3.6/candidate"
     steps:
       - name: Check out code
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


dimaqq:
@SimonRichardson early preview: a way to support Juju 3.6.

In a nutshell:
- use Facade v6 for old Juju (3.0 ~ 3.2.x)
  - exact same code path as before
- use Facade v7 for new Juju (3.3.0 ~ latest)
  - drop client-side series computation, let the controller decide
  - to do: handle explicit base
- deprecate `series=` kwarg

Major to do: support charms' existing integration tests against Juju 3.3.0 ~ 3.5.4
- convert series= to base=
- propagate this or explicit base=

Gmerold:
Hi @dimaqq, 
Do you plan to pick up this PR? It looks like its intent is exactly what my team needs for the tests to start passing ;)
Thanks!

gatici:
Hello @dimaqq,
We are having a problem similar to https://github.com/juju/python-libjuju/issues/1240. Currently, it is a blocker for us.
I assume this PR will fix this issues. We appreciate if you could complete it as soon as possible. 
Many thanks!

dimaqq:
I think so, either I or someone from juju. Realistically in January though.

dimaqq:
P.S. this PR is currently very raw. It doesn’t yet solve anything. The high level direction is clear though.

dimaqq:
this PR has grown stale, a fresh approach would be better.

PR #1186, opened 2024-11-12
chore: ruff up the codebase

dimaqq:
What this PR does

In short, we'll be relying on ruff via pre-commit as opposed to flake8 via tox.

- switches from flake8 to ruff
- reformats the source code
- reformats the docstrings where safe, conservatively
- resolves ruff complaints where that's straightforward
- silences with a `FIXME` where public API could break or code change is not obvious or where too much manual work is required, those can be fixed in separate PRs
- updates the CI and docs for the above

Check list

- [x] pre-commit config
- [x] ruff over the codebase
- [x] resolve or silence errors
- [x] kill old linter and github actions
- [x] GitHub actions to check pre-commit
- [x] run pre-commit after codegen
- [x] add to docs

PR #1187, opened 2024-11-13
publish wheels

dimbleby:
- publish wheels as well as sdists
- direct invocation of `setup.py` is deprecated these days, prefer `build` to create those distributions
- fix the makefile so that it uses the virtual environment that it creates

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,18 +43,23 @@ docs:
 .PHONY: build-test
 build-test:
 	rm -rf venv
-	python3 -m venv venv
-	. venv/bin/activate
-	python3 setup.py sdist
-	pip install ./dist/juju-${VERSION}.tar.gz
-	python3 -c "from juju.controller import Controller"
-	rm ./dist/juju-${VERSION}.tar.gz
+	$(PY) -m venv venv
+	. venv/bin/activate && \
+		pip install build && \
+		python -m build && \
+		pip install ./dist/juju-${VERSION}.tar.gz && \
+		python -c "from juju.controller import Controller"
+	rm -rf dist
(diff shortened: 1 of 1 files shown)


dimbleby:
(you might want to update your bot so that it points to something more useful than 404 not found)

dimaqq:
Thank you @dimbleby 

May I ask the context for your request? Are you using this library or is this a generic concern?

P.S. I was going to improve the build and publication process, though I was thinking jump in the deep end, and go for uv and trusted publishing... That being said, I like simplicity of this PR.

I'll get back to this after #1186 where `pyproject.toml` is introduced and perhaps `setup.py` can be dropped.

dimaqq:
WDYT about this: https://github.com/juju/python-libjuju/pull/1186/commits/cb4a962f99c9f265da2300380a7d04330ee2c904 inspired by your PR?

dimbleby:
If you want to work in uv-world then I would simply replace `python -m build` with `uv build`.

Your commit leaves a virtualenv created but unused - though that is a side-issue.

I'm here not as a user but just someone trying to make the python ecosystem better, one project at a time...

dimaqq:
Thank you @dimbleby , I didn't think `uv build` would work out of the box, but it does indeed!

Incorporated into https://github.com/juju/python-libjuju/pull/1186/commits/f5b93e79873f66907522800ce7d24db29fb4cd19

dimaqq:
Closing this PR, as I've added `uv build` in #1186

dimaqq:
Thank you for pointing me to the obvious!

PR #1097, opened 2024-08-30
[VERY DRAFT] Dima's hacks

dimaqq:

Issue #1178, opened 2024-11-07
Chore: simplify jasyncio


dimaqq:
### Description

Use native asyncio primitives instead of re-exporting them via jasyncio.

See e.g. https://github.com/juju/python-libjuju/pull/1097/files#diff-41d205b5da01238d45d1bf3e3ba62df2bf78aad615c58f391787cecbdf7e4e08

### Urgency

Casually wishlisting

### Code I'd Like to Run

```python
...
```

EdmilsonRodrigues:
Hello, I am new here and I want to contribute, I just read the CONTRIBUTING.md, and I'm with this doubt, can I take this one? Is there another one that might need help from new or good first issues?

dimaqq:
Anyone can take this one, PRs are always welcome.

Given that `juju.jasyncio.abc` could have been used by library users, a proper fix may look something like this:
- keep `juju/jasyncio.py`, don't use it, let it issue a deprecation warning is imported
- move the few unique functions to another file, e.g. `juju/_jasyncio.py` and update the code that uses those functions
- update the code that uses e.g. `jasyncio.sleep()` to `asyncio.sleep()`

EdmilsonRodrigues:
Can we already close this one?

dimaqq:
Yep, it's done

PR #1189, opened 2024-11-14
chore: type hints for the connnection class

dimaqq:
Pull the Connection type hints from #1097

Only minor cleanup beyond that, no logical changes.

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -27,6 +27,7 @@ from .facade import _JSON, _RICH_JSON, TypeEncoder
 from .facade_versions import client_facade_versions, known_unsupported_facades
 
 SPECIFIED_FACADES: TypeAlias = dict[str, dict[Literal["versions"], Sequence[int]]]
+_WebSocket: TypeAlias = "websockets.legacy.client.WebSocketClientProtocol"
 
 LEVELS = ["TRACE", "DEBUG", "INFO", "WARNING", "ERROR"]
 log = logging.getLogger("juju.client.connection")
@@ -125,6 +126,7 @@ class Connection:
     _retry_backoff: float
     uuid: str | None
     messages: IdQueue
+    _ws: _WebSocket | None
 
     @classmethod
     async def connect(
@@ -303,7 +305,7 @@ class Connection:
             context.check_hostname = False
         return context
 
-    async def _open(self, endpoint, cacert):
+    async def _open(self, endpoint, cacert) -> tuple[_WebSocket, str, str, str]:
         if self.is_debug_log_connection:
             assert self.uuid
             url = f"wss://user-{self.username}:{self.password}@{endpoint}/model/{self.uuid}/log"
@@ -726,7 +728,9 @@ class Connection:
         if len(endpoints) == 0:
             raise errors.JujuConnectionError("no endpoints to connect to")
 
-        async def _try_endpoint(endpoint, cacert, delay):
+        async def _try_endpoint(
+            endpoint, cacert, delay
+        ) -> tuple[_WebSocket, str, str, str]:
             if delay:
                 await jasyncio.sleep(delay)
             return await self._open(endpoint, cacert)
@@ -738,6 +742,8 @@ class Connection:
             jasyncio.ensure_future(_try_endpoint(endpoint, cacert, 0.1 * i))
             for i, (endpoint, cacert) in enumerate(endpoints)
         ]
+        result: tuple[_WebSocket, str, str, str] | None = None
+
         for attempt in range(self._retries + 1):
             for task in jasyncio.as_completed(tasks):
                 try:
@@ -760,12 +766,11 @@ class Connection:
             # only executed if inner loop's else did not continue
             # (i.e., inner loop did break due to successful connection)
             break
-        else:
-            # impossible, work around https://github.com/microsoft/pyright/issues/8791
-            assert False  # noqa: B011
 
         for task in tasks:
             task.cancel()
+
+        assert result  # loop raises or sets the result
 
         self._ws = result[0]
         self.addr = result[1]
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -16,7 +16,7 @@ import macaroonbakery.bakery as bakery
 import macaroonbakery.httpbakery as httpbakery
 import websockets
 from dateutil.parser import parse
-from typing_extensions import Self, overload
+from typing_extensions import Self, TypeAlias, overload
 
 from juju import errors, jasyncio, tag, utils
 from juju.client import client
@@ -25,6 +25,8 @@ from juju.version import CLIENT_VERSION
 
 from .facade import _JSON, _RICH_JSON, TypeEncoder
 from .facade_versions import client_facade_versions, known_unsupported_facades
+
+SPECIFIED_FACADES: TypeAlias = dict[str, dict[Literal["versions"], Sequence[int]]]
 
 LEVELS = ["TRACE", "DEBUG", "INFO", "WARNING", "ERROR"]
 log = logging.getLogger("juju.client.connection")
@@ -136,8 +138,7 @@ class Connection:
         max_frame_size: int | None = None,
         retries=3,
         retry_backoff=10,
-        specified_facades: dict[str, dict[Literal["versions"], Sequence[int]]]
-        | None = None,
+        specified_facades: SPECIFIED_FACADES | None = None,
         proxy=None,
         debug_log_conn=None,
         debug_log_params={},
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -26,7 +26,7 @@ from juju.version import CLIENT_VERSION
 from .facade import _JSON, _RICH_JSON, TypeEncoder
 from .facade_versions import client_facade_versions, known_unsupported_facades
 
-SPECIFIED_FACADES: TypeAlias = dict[str, dict[Literal["versions"], Sequence[int]]]
+SPECIFIED_FACADES: TypeAlias = "dict[str, dict[Literal['versions'], Sequence[int]]]"
 _WebSocket: TypeAlias = "websockets.legacy.client.WebSocketClientProtocol"
 
 LEVELS = ["TRACE", "DEBUG", "INFO", "WARNING", "ERROR"]
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -121,6 +121,7 @@ class Connection:
     _retries: int
     _retry_backoff: float
     uuid: str | None
+    messages: IdQueue
 
     @classmethod
     async def connect(
@@ -373,7 +374,7 @@ class Connection:
         if self.proxy is not None:
             self.proxy.close()
 
-    async def _recv(self, request_id):
+    async def _recv(self, request_id: int) -> dict[str, Any]:
         if not self.is_open:
             raise websockets.exceptions.ConnectionClosed(
                 websockets.frames.Close(
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -1,5 +1,6 @@
 # Copyright 2023 Canonical Ltd.
 # Licensed under the Apache V2, see LICENCE file for details.
+from __future__ import annotations
 
 import base64
 import json
@@ -9,12 +10,13 @@ import urllib.request
 import warnings
 import weakref
 from http.client import HTTPSConnection
-from typing import Dict, Literal, Optional, Sequence
+from typing import Any, Literal, Sequence
 
 import macaroonbakery.bakery as bakery
 import macaroonbakery.httpbakery as httpbakery
 import websockets
 from dateutil.parser import parse
+from typing_extensions import Self
 
 from juju import errors, jasyncio, tag, utils
 from juju.client import client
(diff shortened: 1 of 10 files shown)


juju/client/connection.py
```diff
@@ -117,28 +106,41 @@ class Connection:
...
         bakery_client=None,
-        max_frame_size=None,
+        max_frame_size: int | None = None,
         retries=3,
         retry_backoff=10,
-        specified_facades: Optional[
-            Dict[str, Dict[Literal["versions"], Sequence[int]]]
-        ] = None,
+        specified_facades: dict[str, dict[Literal["versions"], Sequence[int]]]
+        | None = None,
```
james-garner-canonical: Is this ruff's autoformatting? There must be a nicer way to format this haha

juju/client/connection.py
```diff
@@ -534,7 +536,7 @@ async def _do_ping():
             log.debug("ping failed because of closed connection")
             pass
 
-    async def rpc(self, msg, encoder=None):
+    async def rpc(self, msg: dict, encoder=None) -> dict:
```
james-garner-canonical: Do you have enough info to parametrize `dict` here? Even as `str, Any` or `str, JSONData`.

(Does pyright complain about the unparametrized generic?)

juju/client/connection.py
```diff
@@ -744,8 +746,13 @@ async def _try_endpoint(endpoint, cacert, delay):
             # only executed if inner loop's else did not continue
             # (i.e., inner loop did break due to successful connection)
             break
+        else:
+            # impossible, work around https://github.com/microsoft/pyright/issues/8791
+            assert False  # noqa: B011
+
```
james-garner-canonical, line 769: Ok so this is a `for ... else`, which I mentally substitute as `for ... nobreak`. Reading through the nested loops above, I see that we will either break or raise. So now I understand that this is basically to tell pyright that `result = await task` will always have been evaluated after we exit the loop

I think this is a bit hard to follow (more the fault of the original code than anything), maybe we could do something like this instead
```py
result: TaskResult | None = None
# nasty loops here
assert result is not None  # loop will raise if it fails to set result
```

or instead add `# pyright: ignore [reportPossiblyUnboundVariable]` where we refer to result, e.g.
```python
        self._ws = result[0]  # pyright: ignore [reportPossiblyUnboundVariable]
        self.addr = result[1]  # pyright: ignore [reportPossiblyUnboundVariable]
        self.endpoint = result[2]  # pyright: ignore [reportPossiblyUnboundVariable]
        self.cacert = result[3]  # pyright: ignore [reportPossiblyUnboundVariable]
```

juju/model.py
```diff
@@ -257,7 +259,9 @@ def get_entity(self, entity_type, entity_id, history_index=-1, connected=True):
 class ModelEntity:
     """An object in the Model tree"""
 
-    def __init__(self, entity_id, model, history_index=-1, connected=True):
+    entity_id: str
+
+    def __init__(self, entity_id: str, model: Model, history_index=-1, connected=True):
```
james-garner-canonical: Better to explicitly annotate `history_index` and `connected` as well if they're actually `int` and `bool` respectively?

juju/model.py
```diff
@@ -3227,16 +3231,16 @@ def make_archive(self, path):
         zf.close()
         return path
 
-    def _check_type(self, path):
+    def _check_type(self, path: str) -> str:
         """Check the path"""
-        s = os.stat(str(path))
+        s = os.stat(path)
```
james-garner-canonical, line 3242: I think calling `str` here is intending to support `path` to be `str | Path`? Though that's unnecessary since python 3.6 when lots of stdlib stuff started accepting `Path` as an argument.

I see that we already `from pathlib import Path` in this module. But I also see that a lot of these methods are clearly written with the assumption that the argument type is a `str`. So I think your decision to use `str` here is correct

juju/utils.py
```diff
@@ -20,11 +22,11 @@
...
     """Wrapper around asyncio.create_subprocess_exec."""
-    p = await jasyncio.create_subprocess_exec(
+    p = await asyncio.create_subprocess_exec(
         *cmd,
-        stdin=jasyncio.subprocess.PIPE,
-        stdout=jasyncio.subprocess.PIPE,
-        stderr=jasyncio.subprocess.PIPE,
+        stdin=asyncio.subprocess.PIPE,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
```
james-garner-canonical, line 29: why the change to not using `jasyncio` here? I notice that it's still imported (line 18)

juju/utils.py
```diff
@@ -93,20 +95,31 @@ class IdQueue:
...
     """
 
-    def __init__(self, maxsize=0):
-        self._queues = defaultdict(partial(jasyncio.Queue, maxsize))
-
-    async def get(self, id_):
+    _queues: dict[int, asyncio.Queue[dict[str, Any] | Exception]]
+
+    def __init__(self):
+        self._queues = defaultdict(asyncio.Queue)
```
james-garner-canonical, line 101: Behaviour change? Or is it always called with the default maxsize internally, and we're confident it isn't used externally?

juju/utils.py
```diff
@@ -93,20 +95,31 @@ class IdQueue:
...
         return value
 
-    async def put(self, id_, value):
+    async def put(self, id_: int, value: dict[str, Any]):
         await self._queues[id_].put(value)
 
-    async def put_all(self, value):
+    async def put_all(self, value: Exception):
         for queue in self._queues.values():
             await queue.put(value)
```
james-garner-canonical, line 124: Interesting signatures, I won't pry too deeply here, I assume they just reflect how the methods are actually used.

But do you think it's worth annotating `-> None:` (in general) to make it clear that the return type has been considered?

juju/model.py
```diff
@@ -2923,7 +2927,7 @@ async def wait_for_idle(
         status=None,
         wait_for_at_least_units=None,
         wait_for_exact_units=None,
-    ):
+    ) -> None:
```
james-garner-canonical, line 2936: I guess it might be a lot of work to annotate all the other arguments if they're unusually typed?

I'd be in favour of annotating any that are easy to annotate tho, e.g. if `raise_on_error/blocked` and `wait_for_active` are `bool`, if `timeout`, `idle_period`, `check_freq` are `float` (or `int`?), and I'm guessing that `wait_for_..._units` are `int | None`. Is `status` a `str | None`?

james-garner-canonical (changes requested):
Sorry, I made a lot of comments. You can resolve most without any changes, but I think it would be nice to use something a bit more readable for the `for ... else` one.

juju/client/connection.py
dimaqq: I've changed from `Optional` to `x|None` and I guess the `|` was the convenient place for ruff to split the line...

I think I'll make a type alias for the outer dict, that should solve it.

juju/client/connection.py
dimaqq: Let me look into this, I think the full type is probably `dict[str, Any]` which is still better than plain `dict`.

dimaqq: P.S. it should complain, it does complain in other contexts... 🤔

juju/model.py
dimaqq: pyright seems to imply the type based on the default.
is that not enough?

juju/utils.py
dimaqq: yes maxsize is always the default.

juju/model.py
dimaqq: There will be a separate PR for .wait_for_idle()

For now I'm sprinkled a few type hints, mostly because I needed them somewhere else.
Does this pass as "it's better than before, let's leave other improvements till later" check?

juju/model.py
james-garner-canonical: My concern is that a later reader doesn't necessarily know whether they're not type annotated because the type inferred from the defaults is obviously correct, or because it's like `wait_for_idle` where we're just gradually adding types. Whereas if you add even the obvious annotations, it's clear that the types for this `__init__` method have been fully worked out.

juju/model.py
james-garner-canonical: Yeah this makes sense, especially since you're working on a new `wait_for_idle`

juju/utils.py
dimaqq: indeed these type hints are for semantically makes sense to call these methods with.

the implementation doesn't care, ofc.

I'm not sure about None return types. I think the annotation is useful only in a limited way -- prevents someone accidentally using the return value, or shows them not to in an IDE.

juju/client/connection.py
dimaqq: I've made types for this thing... pushing now.

juju/client/connection.py
dimaqq: done

juju/client/connection.py
dimaqq: done

juju/client/facade.py
```diff
@@ -13,13 +14,22 @@
...
+from typing import Any, Mapping, Sequence
 
 import packaging.version
 import typing_inspect
+from typing_extensions import TypeAlias
 
 from . import codegen
 
+# Plain JSON, what is received from Juju
+_JSON_LEAF: TypeAlias = None | bool | int | float | str
```
james-garner-canonical: Does this work in python 3.8, or do you need to use a Union? We have annotations from the future, but these `|`s aren't in an annotation

pyproject.toml
```diff
@@ -212,6 +212,7 @@ ignore = [
 [tool.pyright]
 # These are tentative
 # include = ["**/*.py"]
-pythonVersion = "3.8" # check no python > 3.8 features are used
-pythonPlatform = "All"
+pythonVersion = "3.10"
```
james-garner-canonical: Is this ok, don't we still have to run on 20.04?

pyproject.toml
```diff
@@ -212,6 +212,7 @@ ignore = [
...
 typeCheckingMode = "strict"
+useLibraryCodeForTypes = true
+reportGeneralTypeIssues = true
```
james-garner-canonical, line 218: What do these do?

juju/client/connection.py
```diff
@@ -26,6 +26,8 @@
 from .facade import _JSON, _RICH_JSON, TypeEncoder
 from .facade_versions import client_facade_versions, known_unsupported_facades
 
+SPECIFIED_FACADES: TypeAlias = dict[str, dict[Literal["versions"], Sequence[int]]]
+
```
james-garner-canonical, line 31: Looks good, I'm surprised ruff doesn't complain about / fix the case - maybe because it's upper snake? https://docs.astral.sh/ruff/rules/snake-case-type-alias/

juju/client/connection.py
```diff
@@ -536,7 +537,19 @@ async def _do_ping():
...
+    ) -> dict[str, _JSON]: ...
+
+    @overload
+    async def rpc(
+        self, msg: dict[str, _RICH_JSON], encoder: TypeEncoder
+    ) -> dict[str, _JSON]: ...
+
+    async def rpc(
+        self, msg: dict[str, Any], encoder: json.JSONEncoder | None = None
+    ) -> dict[str, Any]:
```
james-garner-canonical: If it's a pain to narrow this definition, let's just leave it with the Any's, but just pointing it out in case you meant to narrow it further

pyproject.toml
dimaqq: pyright.useLibraryCodeForTypes [boolean]: This setting is deprecated in favor of python.analysis.useLibraryCodeForTypes. It will be removed at a future time.

python.analysis.useLibraryCodeForTypes [boolean]: Determines whether pyright reads, parses and analyzes library code to extract type information in the absence of type stub files. Type information will typically be incomplete. We recommend using type stubs where possible. The default value for this option is true.

may help with libraries that don't give us types, like pymacaroons


reportGeneralTypeIssues is actuallyu "error" by default, so maybe I don;'t need this setting.


to be honest, I'm just poking a black cat in the black room in the dark when it comes to pyright settings 🐈‍⬛

juju/client/connection.py
dimaqq: Huh.. interesting...

juju/client/connection.py
dimaqq: I don't think it matters. The future me gets to fix it if it's ever a problem.

pyproject.toml
dimaqq: Maybe I poke that bear tomorrow. Maybe, just maybe I can do better 🤔

juju/client/facade.py
dimaqq: stringified.

juju/model.py
dimaqq: noted.

juju/client/connection.py
dimaqq: done, thank you!

pyproject.toml
dimaqq: done, thank you for not letting be lazy

james-garner-canonical (approved):

juju/utils.py
dimaqq: this was general cleanup, though I'm not doing the bulk 
 of it, leaving that for #1178

PR #1193, opened 2024-11-18
chore: dummy change to test CI

dimaqq:
Somehow a single unit test fails in my setup now, `TestJujuData.test_verify_controller_uninitialized`

this PR is only to check that main is in fact stable.


dimaqq:
cancellation flake confirmed, typically evident in test_list_secrets

however, it's not that the problem is with this test, there are cases where other tests fail.

maybe the solution is to increase the default timeout overall.

PR #1194, opened 2024-11-19
chore: nuke per-facade schema

dimaqq:
Individual _clientNN.py files don't need the schemata inlined, remove that field from codegen.


james-garner-canonical (approved):
Nice, this makes the generated code a lot easier for humans to look at. Hopefully no users have used `XFacade.schema` though

dimaqq:
No matches in super-tox collection of charms or COU 🎉

PR #1198, opened 2024-11-19
chore: remove pytest-xdist

dimaqq:
pytest-xdist somehow eats live logs.

to test, run something like: `uvx tox -e unit -- --log-cli-level=INFO`


james-garner-canonical (approved):
Looks good to me since we run integration tests serially now, so parallel execution is only being used for unit tests, which seems unnecessary since they're fast enough anyway

PR #1104, opened 2024-09-20
draft: live jrpc calls using separate event in a helper thread

dimaqq:

PR #1203, opened 2024-11-20
chore: pull types from wip PR for upcoming Model.wait_for_idle

dimaqq:
Improve type hints.
Sourced from #1104 
These type hints will be used to upstream the new wait_for_idle()


juju/model.py
```diff
@@ -232,7 +267,9 @@ def apply_delta(self, delta):
         entity = self.get_entity(delta.entity, delta.get_id())
         return entity.previous(), entity
 
-    def get_entity(self, entity_type, entity_id, history_index=-1, connected=True):
+    def get_entity(
+        self, entity_type, entity_id, history_index=-1, connected=True
```
james-garner-canonical, line 272: [not a blocker] Am I tripping or did you land type annotations for this line in a previous PR?

juju/model.py
```diff
@@ -879,7 +924,10 @@ def add_local_charm(self, charm_file, series="", size=None):
            instead.
 
         """
-        conn, headers, path_prefix = self.connection().https_connection()
+        connection = self.connection()
+        assert connection
+
```
james-garner-canonical: [not a blocker] Can this `assert` be removed? Looks like `connection` will raise if it can't return an actual `Connection` object. If an `assert` is actually needed, it should probably go just before line 63 of juju/client/connector.py

Line +1324 of this PR has a second `assert` on a `connection` call that can probably go too

juju/model.py
```diff
@@ -1271,11 +1320,13 @@ async def _all_watcher():
                         del allwatcher.Id
                         continue
                     except websockets.ConnectionClosed:
-                        monitor = self.connection().monitor
+                        connection = self.connection()
+                        assert connection
```
james-garner-canonical: [not a blocker] Second assert that could be removed?

juju/status.py
```diff
@@ -8,26 +9,26 @@
 
 log = logging.getLogger(__name__)
 
-""" derive_status is used to determine the application status from a set of unit
-status values.
 
-:param statues: list of known unit workload statues
-
-"""
+def derive_status(statuses: list[str]):
```
james-garner-canonical:
```suggestion
def derive_status(statuses: list[str]) -> str:
```

juju/url.py
```diff
@@ -20,18 +23,23 @@ def __str__(self):
 
 
 class URL:
+    """Private URL class for this library internals only."""
```
james-garner-canonical:
```suggestion
    """Private URL class for this library internals only.
    
    Should be instantiated by calling the `URL.parse` constructor.

    If calling `URL` directly without providing the `name` argument,
    the `name` property will be `None`, and must be corrected manually
    before use.
    """
```
`name` argument is a little funky, but your type annotation solution seems good to me. I'd like to add a little bit to the docstring to explain what's going on.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "macaroonbakery>=1.1,<2.0",
     "pyRFC3339>=1.0,<2.0",
     "pyyaml>=5.1.2",
-    "websockets>=8.1,<14.0",
+    "websockets>=13.0.1,<14.0",
```
james-garner-canonical, line 28: [not a blocker] Nice, I just checked and 13 is the last major version of websockets to support python 3.8, so this works for us =)

juju/model.py
```diff
@@ -1177,11 +1225,12 @@ def name(self):
...
 
     @property
-    def info(self):
+    def info(self) -> ModelInfo:
         """Return the cached client.ModelInfo object for this Model.
 
         If Model.get_info() has not been called, this will return None.
         """
+        assert self._info is not None
         return self._info
```
james-garner-canonical: [blocker?] Behaviour change. Intended? Update docstring if so:
```
        If Model.get_info() has not been called, this will raise an ...
```
 Otherwise preserve behaviour and have return type be `ModelInfo | None`

james-garner-canonical (changes requested):
Looks great, just one potential blocker -- behaviour change for `Model.info` which I guess just means either correct the docstring or revert the behaviour change (and change the type annotation accordingly).

pyproject.toml
dimaqq: IIRC type hints got improved in 12 and/or 13.

juju/model.py
dimaqq: Ehh one integration test tripped on this:

```
  File "/home/runner/work/python-libjuju/python-libjuju/juju/model.py", line 1233, in info
    assert self._info is not None
AssertionError
```

I need to examine this closer.
Also, I'm beginning to appreciate all the hard work you did to get integration tests up and running again 🙇🏻

juju/model.py
dimaqq: I had added type hints to `__init__` which looks kinda similar: `entity_id: str, model: Model`. 

Perhaps that's what you're thinking?

james-garner-canonical: Yeah, that was it https://github.com/juju/python-libjuju/pull/1189/commits/56bb71c9eaaf1965378101c2ecdd77df06d57a68

I guess `entity_type` is a bit of a mystery box, so I won't suggest adding more type hints here for now

juju/model.py
dimaqq: indeed, that's what happens when I try to pull type hints from two separate branches.
done!

juju/model.py
dimaqq: done

juju/status.py
dimaqq: It turns out that fixing this properly was kinda hard.

Please review 🙇🏻

juju/status.py
```diff
@@ -3,42 +3,70 @@
...
+else:
+    from backports.strenum import StrEnum
+
 from .client import client
 
 log = logging.getLogger(__name__)
 
 
-def derive_status(statuses: list[str]):
+class StatusStr(StrEnum):
```
james-garner-canonical, line 19: We can probably use `StatusStr(str, Enum)` instead of `backports`, but I'll leave the choice to you

.github/workflows/test.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
james-garner-canonical, line 49: Ah so each set of unit tests won't block the others, makes sense

juju/client/connection.py
```diff
@@ -323,10 +325,6 @@ async def _open(self, endpoint, cacert) -> tuple[_WebSocket, str, str, str]:
             sock = self.proxy.socket()
             server_hostname = "juju-app"
 
-        def _exit_tasks():
-            for task in jasyncio.all_tasks():
-                task.cancel()
-
```
james-garner-canonical, line 329: Is this just an unused inner function ..?

juju/model.py
james-garner-canonical: Any decision here?

juju/status.py
```diff
@@ -1,43 +1,72 @@
...
-        if status in severities and severities[status] > severities[current]:
-            current = status
-    return current
+    error = "error"
+    blocked = "blocked"
+    waiting = "waiting"
+    maintenance = "maintenance"
+    active = "active"
+    terminated = "terminated"
+    unknown = "unknown"
```
james-garner-canonical, line 31: uppercase?

juju/status.py
```diff
@@ -1,43 +1,72 @@
...
+
+    :param statuses: list of known unit workload statuses
+    """
+    current: StatusStr = StatusStr.unknown
+    for status in statuses:
+        try:
+            status = StatusStr(status)
+        except ValueError:
+            continue
+        if (new_level := severity_map.get(status)) and new_level > severity_map[
```
james-garner-canonical: If we're already making sure every `status` is a valid `StatusStr`, can we just assume that it's in the `severity_map`? It's surely an error to add a status to `StatusStr` without adding a corresponding severity, so we probably shouldn't hide it

pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "Python library for Juju"
 readme = "docs/readme.rst"
 license = { file = "LICENSE" }
 maintainers = [{name = "Juju Ecosystem Engineering", email = "juju@lists.ubuntu.com"}]
-requires-python = ">=3.8"
+requires-python = ">=3.8.6"
```
james-garner-canonical, line 8: Is this for backports?

tox.ini
```diff
@@ -18,6 +18,7 @@ passenv =
     HOME
     TEST_AGENTS
     LXD_DIR
+# FIXME would it be easier to `pip install -e .`?
```
james-garner-canonical: to avoid redundant dependency specification? It does seem like there should be a better way than this

pyproject.toml
dimaqq: Yep. That's what they state.

We've got to support py 3.8, because we want a single virtualenv to be usable for both unit and integration tests.

Arguably the deployment base may not have been updated... 20.4 started with 3.8.2 or something, or maybe it has been updated...

I figure requiring a recent 3.8 in tests should be fine.

tox.ini
dimaqq: Please see a separate PR, #1205

juju/status.py
dimaqq: that needs a separate PR.

there are many more status values in juju codebase, and conceivably a new value could be added in the future.

I'll make a separate test that ensures StatusStr and severity_map consistency.

dimaqq: done.

juju/status.py
dimaqq: That's a good question.
Today, I feel that lower case reads nicely for StrEnum specifically, as that's the exact value the user knows from Juju cli.

If someone changes that later, I don't mind.

dimaqq:
re self._info, pls reload

juju/model.py
dimaqq: added explicit ValueError and refactored the code that uses this path.

juju/client/connection.py
dimaqq: yep, added a year or two back, then usage removed and this was left dangling.

pyproject.toml
james-garner-canonical: Sweet, let's just see if this breaks anyone, low likelihood I'm guessing

juju/model.py
```diff
@@ -1177,11 +1224,15 @@ def name(self):
         return self._info.name
 
     @property
-    def info(self):
+    def info(self) -> ModelInfo:
         """Return the cached client.ModelInfo object for this Model.
 
         If Model.get_info() has not been called, this will return None.
```
james-garner-canonical: I'm happy to assume that raising a `JujuModelError` instead of returning `None` when `self._info` hasn't been populated is a safe change to make, but let's update the docstring to be accurate.
```suggestion
        Raises a JujuModelError if the model is not connected.
```

james-garner-canonical (approved):
Looking good, just one docstring inaccuracy that could be fixed before merging

dimaqq:
I'll need to either wait for https://github.com/canonical/pylxd/issues/615

Or rip out the tox.ini and deps changes and chuck them aside... to a separate PR.

juju/model.py
dimaqq: done

PR #1208, opened 2024-11-22
Drop use of Monitor._ws.open and Connection._ws.closed

### LINKED ISSUE #1207: AttributeError: 'ClientConnection' object has no attribute 'open'
### Description

When trying to use [zaza](https://github.com/openstack-charmers/zaza) with juju 2.9 and a latest version of websockets the following exception is raised due to changes in the library.

```
  File "/home/ubuntu/charms/charm-mysql-innodb-cluster/src/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 834, in _connect_with_login
  File "/home/ubuntu/charms/charm-mysql-innodb-cluster/src/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 923, in login
    return await self.rpc({                                                                                                                                                    
  File "/home/ubuntu/charms/charm-mysql-innodb-cluster/src/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 606, in rpc
    if self.monitor.status == Monitor.DISCONNECTED:                                                                                                                            
  File "/home/ubuntu/charms/charm-mysql-innodb-cluster/src/.tox/func-target/lib/python3.10/site-packages/juju/client/connection.py", line 212, in status
    if stopped or not connection._ws.open:                                                                                                                                     
AttributeError: 'ClientConnection' object has no attribute 'open'
```

### Urgency

### Python-libjuju version

2.9.49

### Juju version

2.9.49

### Reproduce / Test

```python
git clone https://opendev.org/openstack/charm-mysql-innodb-cluster.git
cd charm-mysql-innodb-cluster
tox -e func-target
```

freyes:
The websockets library has dropped the properties "open" and "closed" since websockets-13.0[0], this patch makes the changes needed to check for the connection's state as suggested by the documentation.

[0] https://github.com/python-websockets/websockets/commit/7c8e0b9d6246cd7bdd304f630f719fc55620f89a

Fixes #1207

### DIFF
--- a/juju/client/connection.py
+++ b/juju/client/connection.py
@@ -17,6 +17,7 @@ import macaroonbakery.httpbakery as httpbakery
 import websockets
 from dateutil.parser import parse
 from typing_extensions import Self, TypeAlias, overload
+from websockets.protocol import State
 
 from juju import errors, jasyncio, tag, utils
 from juju.client import client
@@ -92,7 +93,7 @@ class Monitor:
                 and connection._receiver_task.cancelled()
             )
 
-        if stopped or not connection._ws.open:
+        if stopped or connection._ws.state is not State.OPEN:
             return self.ERROR
 
         # everything is fine!
@@ -357,7 +358,7 @@ class Connection:
             tasks_need_to_be_gathered.append(self._debug_log_task)
             self._debug_log_task.cancel()
 
-        if self._ws and not self._ws.closed:
+        if self._ws and self._ws.state is not State.CLOSED:
             await self._ws.close()
 
         if not to_reconnect:


dimaqq:
I imagine that would allow us to upgrade to websockets 14 or remove the dependency upper bound, wouldn’t it?

dimaqq:
Pls reword commit messages to follow conventional commits

dimaqq:
https://github.com/python-websockets/websockets/blob/59d4dcf779fe7d2b0302083b072d8b03adce2f61/src/websockets/protocol.py#L57-L60

The websockets library defines 4 values for `State`.

This PR references only two.

I wonder what about the other two.

juju/client/connection.py
```diff
@@ -357,7 +358,7 @@ async def close(self, to_reconnect: bool = False):
             tasks_need_to_be_gathered.append(self._debug_log_task)
             self._debug_log_task.cancel()
 
-        if self._ws and not self._ws.closed:
+        if self._ws and self._ws.state is not State.CLOSED:
```
dimaqq, line 361: Looking through websockets code, the library already handles the corner cases and we should close the connection unconditionally.

dimaqq:
Continued in #1216

Issue #1000, opened 2024-01-03
Unable to upload .tar file as local resource to deployed charmed application


NucciTheBoss:
## The problem

When trying to run the integration tests for the slurmd charmed operator, our CI pipeline fails with the following stacktrace:

```python
Traceback (most recent call last):
  File "/home/runner/work/slurmd-operator/slurmd-operator/tests/integration/test_charm.py", line 48, in test_build_and_deploy
    await asyncio.gather(
  File "/home/runner/work/slurmd-operator/slurmd-operator/.tox/integration/lib/python3.10/site-packages/juju/model.py", line 1820, in deploy
    resources = await self.add_local_resources(application_name,
  File "/home/runner/work/slurmd-operator/slurmd-operator/.tox/integration/lib/python3.10/site-packages/juju/model.py", line 2052, in add_local_resources
    data = p.read_text() if p.exists() else ''
  File "/usr/lib/python3.10/pathlib.py", line 1135, in read_text
    return f.read()
  File "/usr/lib/python3.10/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

This error is raised when we try to attach our resource  `lbnl-nhc-1.4.3.tar.gz` to the slurmd application. NHC (Node Health Check) is a utility for measuring the health of compute nodes in your cluster; it is commonly used to collect observability metrics. When we try to attach NHC, `pathlib.Path(...).read_text(...)` tries to decode the .tar.gz file as an utf-8 text file rather than a binary file. This causes python-libjuju to raise the above `UnicodeDecodeError` which then torpedoes our CI pipeline. The block of code below is what is responsible for this issue: 

https://github.com/juju/python-libjuju/blob/2581b0ced1df6201c6b7fd8cc0b20dcfa9d97c51/juju/model.py#L2050-L2052

This could potentially be fixed by switch the implementation to `pathlib.Path(...).read_bytes(...)`

## Steps to reproduce

1. _Assume that you already have a cloud and Juju controller set up..._
2. `git clone -b jaime/drop-focal-support git@github.com:jaimesouza/slurmd-operator.git`
3. `cd slurmd-operator`
4. `tox run -e integration`
5. Wait a few minutes...
6. See stacktrace with `UnicodeDecodeError`

hmlanigan:
Closed prematurely by the bot.

PR #1147, opened 2024-10-08
Enable attachment of binary files as charm resources

nrobinaubertin:
#### Description

This enables the attachment of binary files as charm resources.

Fixes: https://github.com/juju/python-libjuju/issues/1000

PR #1218, opened 2024-11-26
fix: enable attachment of binary files as charm resources

dimaqq:
I've reset author and re-signed commits from #1147 to get the CLA check to pass.


dimaqq:
Fixes #1000

james-garner-canonical (approved):

PR #1219, opened 2024-11-26
feat: new wait for idle

dimaqq:
Pull the new Model.wait_for_idle implementation from #1104 

- [x] pull in the unit tests
- [x] pull in the implementation
- [x] new unit tests pass
- [x] all unit tests pass
- [x] feature flag
- [x] run integration tests with and without the feature flag
- [x] implement idle/ready distinction properly
- [x] merge #1242 and rebase

### DIFF
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -14,7 +14,7 @@ import typing
 from collections import defaultdict
 from glob import glob
 from pathlib import Path
-from typing import Any, Mapping, Sequence
+from typing import Any, Mapping, Sequence, TypeVar, overload
 
 import packaging.version
 import typing_inspect
@@ -183,7 +183,7 @@ class TypeRegistry(dict):
         return self.get_ref_type(obj["$ref"])
 
 
-CLASSES = {}
+CLASSES: dict[str, type[Type]] = {}
 factories = codegen.Capture()
 
 
@@ -479,35 +479,46 @@ def ReturnMapping(cls):  # noqa: N802
     def decorator(f):
         @functools.wraps(f)
         async def wrapper(*args, **kwargs):
-            nonlocal cls
             reply = await f(*args, **kwargs)
-            if cls is None:
-                return reply
-            if "error" in reply:
-                cls = CLASSES["Error"]
-            if typing_inspect.is_generic_type(cls) and issubclass(
-                typing_inspect.get_origin(cls), Sequence
-            ):
-                parameters = typing_inspect.get_parameters(cls)
-                result = []
-                item_cls = parameters[0]
-                for item in reply:
-                    result.append(item_cls.from_json(item))
-                    """
-                    if 'error' in item:
-                        cls = CLASSES['Error']
-                    else:
-                        cls = item_cls
-                    result.append(cls.from_json(item))
-                    """
-            else:
-                result = cls.from_json(reply["response"])
-
-            return result
+            return _convert_response(reply, cls=cls)
 
         return wrapper
 
     return decorator
+
+
+@overload
+def _convert_response(response: dict[str, Any], *, cls: type[SomeType]) -> SomeType: ...
+
+
+@overload
+def _convert_response(response: dict[str, Any], *, cls: None) -> dict[str, Any]: ...
+
+
+def _convert_response(response: dict[str, Any], *, cls: type[Type] | None) -> Any:
+    if cls is None:
+        return response
+    if "error" in response:
+        cls = CLASSES["Error"]
+    if typing_inspect.is_generic_type(cls) and issubclass(
+        typing_inspect.get_origin(cls), Sequence
+    ):
+        parameters = typing_inspect.get_parameters(cls)
+        result = []
+        item_cls = parameters[0]
+        for item in response:
+            result.append(item_cls.from_json(item))
+            """
+            if 'error' in item:
+                cls = CLASSES['Error']
+            else:
+                cls = item_cls
+            result.append(cls.from_json(item))
+            """
+    else:
+        result = cls.from_json(response["response"])
+
+    return result
 
 
 def make_func(cls, name, description, params, result, _async=True):
@@ -663,7 +674,7 @@ class Type:
         return result
 
     @classmethod
-    def from_json(cls, data):
+    def from_json(cls, data: Type | str | dict[str, Any] | list[Any]) -> Type | None:
         def _parse_nested_list_entry(expr, result_dict):
             if isinstance(expr, str):
                 if ">" in expr or ">=" in expr:
@@ -740,6 +751,9 @@ class Type:
         except KeyError:
             return default
         return getattr(self, attr, default)
+
+
+SomeType = TypeVar("SomeType", bound=Type)
 
 
 class Schema(dict):
--- a/juju/client/facade.py
+++ b/juju/client/facade.py
@@ -499,10 +499,14 @@ def _convert_response(response: dict[str, Any], *, cls: type[Type] | None) -> An
     if cls is None:
         return response
     if "error" in response:
+        # TODO: I don't think this ever happens,
+        # errors are handled by Connection.rpc(),
+        # though, admittedly the shape is different.
         cls = CLASSES["Error"]
     if typing_inspect.is_generic_type(cls) and issubclass(
         typing_inspect.get_origin(cls), Sequence
     ):
+        # TODO: I'm not sure this ever happens either.
         parameters = typing_inspect.get_parameters(cls)
         result = []
         item_cls = parameters[0]
--- a/juju/model/__init__.py
+++ b/juju/model/__init__.py
@@ -3302,13 +3302,14 @@ class Model:
         deadline = None if timeout is None else started + timeout
 
         async def status_on_demand():
-            yield _idle.check(
-                await self.get_status(),
-                apps=apps,
-                raise_on_error=raise_on_error,
-                raise_on_blocked=raise_on_blocked,
-                status=status,
-            )
+            while True:
+                yield _idle.check(
+                    await self.get_status(),
+                    apps=apps,
+                    raise_on_error=raise_on_error,
+                    raise_on_blocked=raise_on_blocked,
+                    status=status,
+                )
 
         async for done in _idle.loop(
             status_on_demand(),
--- a/juju/model/__init__.py
+++ b/juju/model/__init__.py
@@ -1,10 +1,12 @@
 # Copyright 2023 Canonical Ltd.
 # Licensed under the Apache V2, see LICENCE file for details.
+"""Represent Juju Model, as in the workspace into which applications are deployed."""
+
 from __future__ import annotations
 
 import asyncio
 import base64
-import collections
+import collections.abc
 import hashlib
 import json
 import logging
@@ -13,6 +15,7 @@ import re
 import stat
 import sys
 import tempfile
+import time
 import warnings
 import weakref
 import zipfile
@@ -20,23 +23,26 @@ from concurrent.futures import CancelledError
 from datetime import datetime, timedelta
 from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, Mapping, overload
+from typing import TYPE_CHECKING, Any, Iterable, Literal, Mapping, overload
 
 import websockets
 import yaml
 from typing_extensions import deprecated
 
-from . import provisioner, tag, utils
-from .annotationhelper import _get_annotations, _set_annotations
-from .bundle import BundleHandler, get_charm_series, is_local_charm
-from .charmhub import CharmHub
-from .client import client, connection, connector
-from .client.overrides import Caveat, Macaroon
-from .constraints import parse as parse_constraints
-from .constraints import parse_storage_constraints
-from .controller import ConnectedController, Controller
-from .delta import get_entity_class, get_entity_delta
-from .errors import (
+from .. import provisioner, tag, utils
+from ..annotationhelper import _get_annotations, _set_annotations
+from ..bundle import BundleHandler, get_charm_series, is_local_charm
+from ..charmhub import CharmHub
+from ..client import client, connection, connector
+from ..client._definitions import ApplicationStatus as ApplicationStatus
+from ..client._definitions import MachineStatus as MachineStatus
+from ..client._definitions import UnitStatus as UnitStatus
+from ..client.overrides import Caveat, Macaroon
+from ..constraints import parse as parse_constraints
+from ..constraints import parse_storage_constraints
+from ..controller import ConnectedController, Controller
+from ..delta import get_entity_class, get_entity_delta
+from ..errors import (
     JujuAgentError,
     JujuAPIError,
     JujuAppError,
@@ -49,27 +55,37 @@ from .errors import (
     JujuUnitError,
     PylibjujuError,
 )
-from .exceptions import DeadEntityException
-from .names import is_valid_application
-from .offerendpoints import ParseError as OfferParseError
-from .offerendpoints import parse_local_endpoint, parse_offer_url
-from .origin import Channel, Source
-from .placement import parse as parse_placement
-from .secrets import create_secret_data, read_secret_data
-from .tag import application as application_tag
-from .url import URL, Schema
-from .version import DEFAULT_ARCHITECTURE
+from ..exceptions import DeadEntityException
+from ..names import is_valid_application
+from ..offerendpoints import ParseError as OfferParseError
+from ..offerendpoints import parse_local_endpoint, parse_offer_url
+from ..origin import Channel, Source
+from ..placement import parse as parse_placement
+from ..secrets import create_secret_data, read_secret_data
+from ..tag import application as application_tag
+from ..url import URL, Schema
+from ..version import DEFAULT_ARCHITECTURE
+from . import _idle
 
 if TYPE_CHECKING:
-    from .application import Application
-    from .client._definitions import FullStatus
-    from .constraints import StorageConstraintDict
-    from .machine import Machine
-    from .relation import Relation
-    from .remoteapplication import ApplicationOffer, RemoteApplication
-    from .unit import Unit
+    from ..application import Application
+    from ..client._definitions import FullStatus
+    from ..constraints import StorageConstraintDict
+    from ..machine import Machine
+    from ..relation import Relation
+    from ..remoteapplication import ApplicationOffer, RemoteApplication
+    from ..unit import Unit
 
-log = logging.getLogger(__name__)
+log = logger = logging.getLogger(__name__)
+
+
+def use_new_wait_for_idle() -> bool:
+    val = os.getenv("JUJU_NEW_WAIT_FOR_IDLE")
+    if not val:
+        return False
+    if val.isdigit():
+        return bool(int(val))
+    return val.title() != "False"
 
 
 class _Observer:
@@ -632,9 +648,9 @@ class Model:
 
     def __init__(
         self,
-        max_frame_size=None,
-        bakery_client=None,
-        jujudata=None,
+        max_frame_size: int | None = None,
+        bakery_client: Any = None,
+        jujudata: Any = None,
     ):
         """Instantiate a new Model.
 
@@ -2664,14 +2680,21 @@ class Model:
             results[tag.untag("action-", a.action.tag)] = a.status
         return results
 
-    async def get_status(self, filters=None, utc=False) -> FullStatus:
+    async def get_status(self, filters=None, utc: bool = False) -> FullStatus:
         """Return the status of the model.
 
         :param str filters: Optional list of applications, units, or machines
             to include, which can use wildcards ('*').
-        :param bool utc: Display time as UTC in RFC3339 format
+        :param bool utc: Deprecated, display time as UTC in RFC3339 format
 
         """
+        if utc:
+            warnings.warn(
+                "Model.get_status() utc= parameter is deprecated",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
         client_facade = client.ClientFacade.from_connection(self.connection())
         return await client_facade.FullStatus(patterns=filters)
 
@@ -2998,7 +3021,7 @@ class Model:
 
     async def wait_for_idle(
         self,
-        apps: list[str] | None = None,
+        apps: Iterable[str] | None = None,
         raise_on_error: bool = True,
         raise_on_blocked: bool = False,
         wait_for_active: bool = False,
@@ -3011,7 +3034,7 @@ class Model:
     ) -> None:
         """Wait for applications in the model to settle into an idle state.
 
-        :param List[str] apps: Optional list of specific app names to wait on.
+        :param Iterable[str]|None apps: Optional list of specific app names to wait on.
             If given, all apps must be present in the model and idle, while other
             apps in the model can still be busy. If not given, all apps currently
             in the model must be idle.
@@ -3037,6 +3060,7 @@ class Model:
             units of all apps need to be `idle`. This delay is used to ensure that
             any pending hooks have a chance to start to avoid false positives.
             The default is 15 seconds.
+            Exact behaviour is undefined for very small values and 0.
 
         :param float check_freq: How frequently, in seconds, to check the model.
             The default is every half-second.
@@ -3053,6 +3077,21 @@ class Model:
             going into the idle state. (e.g. useful for scaling down).
             When set, takes precedence over the `wait_for_units` parameter.
         """
+        if use_new_wait_for_idle():
+            await self.new_wait_for_idle(
+                apps=apps,
+                raise_on_error=raise_on_error,
+                raise_on_blocked=raise_on_blocked,
+                wait_for_active=wait_for_active,
+                timeout=timeout,
+                idle_period=idle_period,
+                check_freq=check_freq,
+                status=status,
+                wait_for_at_least_units=wait_for_at_least_units,
+                wait_for_exact_units=wait_for_exact_units,
+            )
+            return
+
         if wait_for_active:
             warnings.warn(
                 "wait_for_active is deprecated; use status",
@@ -3065,16 +3104,18 @@ class Model:
             wait_for_at_least_units if wait_for_at_least_units is not None else 1
         )
 
-        timeout = timedelta(seconds=timeout) if timeout is not None else None
-        idle_period = timedelta(seconds=idle_period)
+        timeout_ = timedelta(seconds=timeout) if timeout is not None else None
+        idle_period_ = timedelta(seconds=idle_period)
         start_time = datetime.now()
-        # Type check against the common error of passing a str for apps
-        if apps is not None and (
-            not isinstance(apps, list) or any(not isinstance(o, str) for o in apps)
-        ):
-            raise JujuError(f"Expected a List[str] for apps, given {apps}")
 
-        apps = apps or self.applications
+        if isinstance(apps, (str, bytes, bytearray, memoryview)):
+            raise TypeError(f"apps must be a Iterable[str], got {apps=}")
+
+        apps_ = list(apps or self.applications)
+
+        if any(not isinstance(o, str) for o in apps_):
+            raise TypeError(f"apps must be a Iterable[str], got {apps_=}")
+
         idle_times: dict[str, datetime] = {}
         units_ready: set[str] = set()  # The units that are in the desired state
         last_log_time: datetime | None = None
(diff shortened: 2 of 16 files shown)


dimaqq:
@james-garner-canonical I'd appreciate an early review.

I'm expecting integration tests to pass 🎉 

There's still some refactoring to be done... and I'm happy to hear your thoughts about that and areas to improve in general.

dimaqq:
Transient error:
`juju.errors.JujuUnitError: 'docker-registry/0' workload has errored: 'hook failed: "install"'`

I’ve restarted that one job

juju/model.py
```diff
@@ -3212,6 +3245,222 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+                "wait_for_active is deprecated; use status",
+                DeprecationWarning,
+                stacklevel=3,
+            )
+            status = "active"
+
+        _wait_for_units = (
+            wait_for_at_least_units if wait_for_at_least_units is not None else 1
+        )
+        apps = apps or list(self.applications)
```
james-garner-canonical: I see there was some type checking here in the old implementation that raised a `JujuError` if `apps` wasn't a `list` of `str` (or `None`), with a note that passing a `str` for the `apps` argument was a common error.

Looks like now if you passed a `str` or any other iterable, `_check_idle` would `return False` on line 3330 instead, leading to either timeout or run forever. Well, unless your  app names were all one letter and you passed a string concatenating their names, that would work like you'd passed them in a list.

Worth adding the `JujuError` back? And maybe a unit test that the new implementation handles bad arguments the same way that the old implementation did?

juju/model.py
```diff
@@ -3212,6 +3245,222 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+
+        # Order of errors:
+        #
+        # Machine error (any unit of any app from apps)
+        # Agent error (-"-)
+        # Workload error (-"-)
+        # App error (any app from apps)
+        #
+        # Workload blocked (any unit of any app from apps)
+        # App blocked (any app from apps)
```
james-garner-canonical: I see that this order was previously defined in the `_raise_for_status` inner function.
(I also note the use of Workload instead of Unit, though a `JujuUnitError` is still what's raised).

I think returning and raising early is a much nicer implementation. Are there any concerns with the order of operations not matching exactly? I think where the order of operations is timing related (when exactly in time we query juju), it doesn't matter so much -- the variance in status due to timing washes out the difference there I think. What are your thoughts here?

I like this face btw (-"-)

juju/model.py
```diff
@@ -3212,6 +3245,222 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+                    "Waiting for app %r units %s/%s",
+                    app_name,
+                    len(ready_units),
+                    _wait_for_units,
+                )
+                return False
+
+        if busy := [n for n, t in idle_times.items() if expected_idle_since < t]:
+            logger.info("Waiting for %s to be idle enough", busy)
+            return False
```
james-garner-canonical: Suggestions:
- rename `idle_times` to `last_active` or `last_non_idle` (more obvious what we're tracking here)
- could we use a property or something to exfiltrate this state between calls instead of a mutable argument? e.g. `self._last_active_unit_times` (any concurrency concerns with this?)
- I like the one liner, but to understand it I mentally expand to more verbose and explicit names, (and swap conditional direction)
```py
busy_units = [
    unit_name
    for unit_name, last_active_time
    in last_active.items()
    if last_active_time > required_idle_since
]
if busy_units: ...
```

juju/model.py
```diff
@@ -3212,6 +3245,222 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+
+        return True
```
james-garner-canonical: Below I've pasted my understanding of the new and original wait for idle logic. Differences in order of operations are:
- original could raise errors before all apps are available, new can't
- original could return early once enough units were ready (assuming no errors had been accumulated in that pass through the apps) -- new will check for all errors before checking if units are ready (an improvement I think)

I suspect that both of these divergences are fine.

---

new

0. do all apps exist? if not, stop and request to check again later
1. check all apps and units for errors in the appropriate order, raising immediately if the user cares
2. for each app, figure out which units are ready -- if there are less than we're waiting for, stop and request to check again later
3. check if units have been idle for long enough -- if not, stop and request to heck again later
4. we're done

original

0. iterate over each app
1. does the app exist yet? if not, stop processing this app and request to iterate again
2. is the app in error/blocked and is that a problem? record it if so
3. are we waiting for enough units, and if so are there enough?
    - if there aren't, stop processing this app
    - if there are, stop processing apps entirely (but we'll start again if any earlier apps requested it -- not any later ones though -- smells like a bug)
4. now check the app's units for errors/blocked and record it if the user cares
5. now figure out if the app status means we need to start again
6. after stopping iteration
    - if any errors were recorded, raise the highest priority one
    - if any apps needed to start again, restart iteration, otherwise we're done

tests/unit/test_wait_for_idle.py
```diff
@@ -0,0 +1,540 @@
+# Copyright 2024 Canonical Ltd.
```
james-garner-canonical: I like the new unit tests, just a note about testing the case where `apps` is a `str` since old implementation raises an error here that I don't believe the new implementation does (instead silently timing out?)

tests/unit/test_wait_for_idle.py
```diff
@@ -0,0 +1,540 @@
...
+    kwargs["apps"] = ["hexanator"]
+    kwargs["wait_for_at_least_units"] = 2
+    kwargs["_wait_for_units"] = 2
+    kwargs["status"] = "active"
+
+    idle = await model_fake(full_status_response)._check_idle(**kwargs)
+    assert not idle
+
+
+# FIXME eventual timeout
```
james-garner-canonical: Not clear to me what this is referring to

juju/unit.py
```diff
@@ -15,7 +15,9 @@
 
 
 class Unit(model.ModelEntity):
-    name: str
+    @property
+    def name(self) -> str:
+        return self.entity_id
```
james-garner-canonical, line 20: Was this a type annotation error? name never actually set outside our tests?

juju/client/facade.py
```diff
@@ -479,37 +479,48 @@ def ReturnMapping(cls):  # noqa: N802
...
 
+@overload
+def _convert_response(response: dict[str, Any], *, cls: type[SomeType]) -> SomeType: ...
+
+
+@overload
+def _convert_response(response: dict[str, Any], *, cls: None) -> dict[str, Any]: ...
+
+
+def _convert_response(response: dict[str, Any], *, cls: type[Type] | None) -> Any:
```
james-garner-canonical, line 498: I see why you split this out, very helpful for testing. I think the overload isn't quite correct according to the implementation, since we can get an error type depending on the response contents, and the code will return a list if the `cls` was defined by a schema with `"type": "array"`.
```py
if TYPE_CHECKING:
    from juju.client._definitions import Error
@overload
def _convert_response(
    response: dict[str, Any], *, cls: type[SomeType]
) -> SomeType | list[SomeType] | Error: ...
@overload
def _convert_response(response: dict[str, Any], *, cls: None) -> dict[str, Any]: ...
def _convert_response(response: dict[str, Any], *, cls: type[Type] | None) -> Any:
```
I think `Any` is fine as the return type of the concrete implementation to avoid touching the implementation itself too much. Could change `dict[str, Any]` to `dict[str, JSONData]` but not a big deal.

james-garner-canonical (commented):
Looks great, so much nicer than the previous implementation. Really cool to have the tests passing on both. Was wait_for_idle tested much previously?

juju/model.py
dimaqq: The client code is run against a snapshot of Juju state, so order can be deterministic, as far as Juju allows that.

The reason I wanted to preserve the semantics of the exceptions this method raises is that there are existing integration tests `with pytest.raises(JujuUnitError)` that I would rather not accidentally break.

Arguably those tests should catch the exception by superclass, but they are what they are.

juju/model.py
dimaqq: Let me think about that.
last active time is not actually tracked.
What is tracked is "unit X was continuously idle since sampled at 11:22:33", for some degree of "continuously" obv.

juju/model.py
dimaqq: I'm refactoring this mess... wait for it!

juju/unit.py
dimaqq: There are, weirdly two sources of truth for the name:
- any ModelEntity, when created is pinned by entity id.
- then state updates are dispatched to safe data by entity id.
- safe_data["name"] exists and equals entity id because of the dispatch.
- __getattr__(self, "name") used to fetch it from safe_data

I'm basically making this a little more sane.
It's true that I didn't have to make this change in this very PR, but it helped me with types.... or tests for something.

james-garner-canonical: I have no objections to making the change, it just jumped out at me. Thanks for explaining

juju/model.py
james-garner-canonical: Ah yep, so `last_active_time` is not really an accurate name. `idle_since` maybe

juju/model.py
james-garner-canonical: 100% agree that raising the exact same exceptions is best for backwards compatibility

juju/model.py
dimaqq: done

juju/model.py
dimaqq: done

juju/model.py
dimaqq: Please have a look now :)

tests/unit/test_wait_for_idle.py
dimaqq: done

tests/unit/test_wait_for_idle.py
dimaqq: moved and expanded the list of missing tests

juju/client/facade.py
dimaqq: Let me think about that...
I thought that if there was an error, it would get raised, rather than returned.
But maybe I misread the code.

dimaqq:
@james-garner-canonical the big refactor is done, I hope you'll like the result :)

still a draft, as some unit tests are missing and I have one internal concern, but should be ready for review :)

juju/client/facade.py
james-garner-canonical: There's something like
```py
if 'error' in reply:
    cls = CLASSES['Error']
... # later
result = cls.from_json(reply['response'])
return result
```
And `Error` is a `Type` subclass in `_definitions.py` -- I don't think it raises on being instantiated or anything unless the `response` is messed up

juju/model/idle.py
```diff
@@ -0,0 +1,228 @@
...
+            rv.units.add(unit_name)
+            assert unit.agent_status
+            assert unit.workload_status
+
+            if unit.agent_status.status != "idle":
+                continue
+            if status and unit.workload_status.status != status:
+                continue
+
+            ready_units.append(unit)
```
james-garner-canonical: I guess `ready_units` is a leftover that can be removed?

juju/model/__init__.py
```diff
@@ -3204,14 +3248,87 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+            raise ValueError("apps must be a Sequence[str]")
+
+        idle_loop = idle._loop(
+            apps=apps,
+            timeout=timeout,
+            wait_for_exact_units=wait_for_exact_units,
+            wait_for_units=wait_for_units,
+            idle_period=idle_period,
+        )
+        next(idle_loop)  # Prime the generator
```
james-garner-canonical: Can we eliminate the need for this by just setting `status` to `None` unconditionally on the first line of `_loop` instead of using `= yield`?

(But see comments on `def _loop` -- I think a regular object would be a lot easier to follow internally than a generator)

juju/model/idle.py
```diff
@@ -0,0 +1,228 @@
...
+    wait_for_exact_units: int | None = None,
+    wait_for_units: int,
+    idle_period: float,
+) -> Generator[bool, _CheckStatus | None, None]:
+    """The outer, time-dependents logic of a wait_for_idle loop."""
+    status: _CheckStatus | None = yield False  # None at first
+    deadline = None if timeout is None else time.monotonic() + timeout
+    idle_since: dict[str, float] = {}
+
+    while True:
```
james-garner-canonical: I think the purpose of having a generator here is to encapsulate some initially provided parameters and some internal state (`idle_since`), is that right? If so, I think it would be worth making this a regular class instead. Example below of what I think should be pretty much a drop in equivalent which seems easier to understand (to me)
```py
class _loop:  # class IdleLoop
    def __init__(
        self,
        *,
        apps: Sequence[str],
        timeout: float  | None,
        wait_for_exact_units: int | None = None,
        wait_for_units: int,
        idle_period: float,
    ):
        self.apps = apps
        self.deadline = None if timeout is None else time.monotonic() + timeout
        self.wait_for_exact_units = wait_for_exact_units
        self.wait_for_units = wait_for_units
        self.idle_period = idle_period
        self.idle_since: dict[str, float] = {}

    def send(self, status: _CheckStatus | None) -> bool:  # def run/check
        # body of while True loop with self. where appropriate
```

juju/model/idle.py
```diff
@@ -0,0 +1,228 @@
...
+        #
+        # However, the "idle_since" check may need to be updated
+        # Right now, a "busy" unit is stamped with ready := since right now
+        #
+        # This is fine for idle_period > 0
+        # But may be problematic for idle_period===0 and wait_for_units==0
+        # (there's a single use case for that in the wild)
+        #
+        # Specifically, a unit in an error state is not ready
+        # But it gets stamped with "now" and seems "idle enough"
```
james-garner-canonical: Interesting. What are the semantics of the original `wait_for_idle` here, and how important is it to match them? The behaviour you've described seems to be clearly buggy. A simple solution might be to stamp a "busy" unit with ready := since some time in the future.

juju/model/idle.py
```diff
@@ -0,0 +1,228 @@
...
+                n for n in status.ready_units if n.startswith(f"{app_name}/")
+            ]
+            if len(ready_units) < wait_for_units:
+                logger.info(
+                    "Waiting for app %r units %s >= %s",
+                    app_name,
+                    len(status.ready_units),
+                    wait_for_units,
+                )
+                rv = False
```
james-garner-canonical: Since the loop never sets `rv =True` I guess we could break any time we set `rv = False`? But it's cheap since we're just iterating over data we've already queried from Juju, and the logging is nice, so I'm -1 on my own suggestion here ...

juju/model/idle.py
```diff
@@ -0,0 +1,228 @@
...
+                    wait_for_exact_units,
+                )
+                rv = False
+
+        # FIXME possible interaction between "wait_for_units" and "idle_period"
+        # Assume that we've got some units ready and some busy
+        # What are the semantics for returning True?
+        if busy := [n for n, t in idle_since.items() if t > expected_idle_since]:
+            logger.info("Waiting for %s to be idle enough", busy)
+            rv = False
```
james-garner-canonical: My first thought is something like this:
```py
ready = [n for n, t in idle_since.items() if t < expected_idle_since]
if len(ready) < wait_for_units:
    logger.info(...)
    rv = False
```
But maybe it would make more sense to instead use `idle_since` to filter units out of `ready_units` before the `for app in apps` above (line 76) -- or have that loop use `idle_since` internally if you want the logging to be able to distinguish between whether a unit is ready at all vs ready for long enough?

juju/model/__init__.py
```diff
@@ -3064,16 +3103,21 @@ async def wait_for_idle(
             wait_for_at_least_units if wait_for_at_least_units is not None else 1
         )
 
-        timeout = timedelta(seconds=timeout) if timeout is not None else None
-        idle_period = timedelta(seconds=idle_period)
+        timeout_ = timedelta(seconds=timeout) if timeout is not None else None
+        idle_period_ = timedelta(seconds=idle_period)
```
james-garner-canonical, line 3108: Names like this always strike me as potentially very error prone for future changes. Is the rename to get the types right?

juju/model/__init__.py
```diff
@@ -3064,16 +3103,21 @@ async def wait_for_idle(
...
+            raise ValueError("apps must be a Sequence[str] and not a plain type")
+
+        apps_ = apps or list(self.applications)
+
+        if not isinstance(apps_, collections.abc.Sequence):
+            raise ValueError("apps must be a Sequence[str]")
+
+        if any(not isinstance(o, str) for o in apps_):
+            raise ValueError("apps must be a Sequence[str]")
+
```
james-garner-canonical: So this is the first actual change this PR makes to old wait for idle. `ValueError` seems appropriate here (or `TypeError`), but is it ok to switch from raising a `JujuError`? Presumably no one is catching these currently in their tests anyway so it's probably ok

juju/model/__init__.py
```diff
@@ -3064,16 +3103,21 @@ async def wait_for_idle(
...
-        if apps is not None and (
-            not isinstance(apps, list) or any(not isinstance(o, str) for o in apps)
-        ):
-            raise JujuError(f"Expected a List[str] for apps, given {apps}")
 
-        apps = apps or self.applications
+        if isinstance(apps, (str, bytes, bytearray, memoryview)):
+            raise ValueError("apps must be a Sequence[str] and not a plain type")
+
+        apps_ = apps or list(self.applications)
```
james-garner-canonical: I could look and try to figure this out but I'll just be lazy and ask -- why the change from `self.applications` to `list(self.applications)`?

james-garner-canonical (commented):
I haven't looked at the tests (though I noticed the addition of `freezegun`), but I've left some comments on the new implementation. My main concern is that the generator/coroutine business would be a lot easier (for me) to follow if we used a regular class/object instead (see comment on `def _loop`) -- it took me a while to get my head around. I do like the separation of the time-based checks. Lmk if you want me to look at anything else in particular

dimaqq:
@james-garner-canonical I've refactored the loop to be an [async] filter.

It turns out that I've misunderstood the definition of "idle".

In short:

- ready units: depends on workload (and possibly agent, machine, etc.)
- idle units: depends on agent only

This means that I have to update the `idle.check` code.

Edit: done now.

juju/model/__init__.py
dimaqq: because of the type hint really, as self.applications is a dict, and keys are something special... anyway both are iterable, but type hints force some kind of cast.

juju/model/__init__.py
dimaqq: I've changed this to TypeError now.

juju/model/__init__.py
dimaqq: yes, and I explicitly don't want to refactor the old idle loop, lest I break it and not notice.

juju/model/idle.py
dimaqq: I'll have to rewrite this entirely.
https://matrix.to/#/!aEDlwVgqmZMLeZIcUP:ubuntu.com/$q0-_izqQi03-Eeqbpp259srutArMADqW2pVO_EDgIbY?via=ubuntu.com

juju/model/idle.py
dimaqq: Maybe, I'll play with this a bit later.
There's a slight difference in logging, but that's not important.

juju/model/idle.py
dimaqq: done

juju/model/__init__.py
dimaqq: done

juju/model/idle.py
dimaqq: done

juju/model/idle.py
dimaqq: https://matrix.to/#/!UPTldjWgQImSaCAjYf:ubuntu.com/$A20c0covFbezl2YplO2kqkGFvsp0n5-2Vnsrx1HiHao?via=ubuntu.com&via=matrix.org

juju/model/idle.py
dimaqq: I've changed this to an async filter now.

Class code may be more straightforward, but I'll keep that as a backup plan.

juju/model/__init__.py
dimaqq: done, it's async filter now.

juju/model/idle.py
dimaqq: good catch, done

juju/model.py
dimaqq: your reading of the old loop is pretty close, but I think I've found one caveat: https://matrix.to/#/!UPTldjWgQImSaCAjYf:ubuntu.com/$WRycclIa2h__O7A4vpJHuN4jBRygLeEQMYVS3gU3JFk?via=ubuntu.com&via=matrix.org

dimaqq:
@james-garner-canonical idle implemented, most comments addressed, tests pass again. please review :)

juju/model/_idle.py
```diff
@@ -0,0 +1,215 @@
...
+            ]
+            if len(ready_units) < wait_for_units:
+                logger.info(
+                    "Waiting for app %r units %s >= %s",
+                    app_name,
+                    len(status.ready_units),
+                    wait_for_units,
+                )
+                yield False
+                continue
```
dimaqq: Ah, this is wrong, the intention is to continue the outer loop, not the inner!

dimaqq: fixed and covered with tests.

juju/client/facade.py
dimaqq: I've looked through the code here and in Connection.rpc() and I'm pretty sure that an "error" can't happen here.
I've added comments to that end.

juju/model/__init__.py
```diff
@@ -3205,14 +3246,87 @@ def _raise_for_status(entities: dict[str, list[str]], status: Any):
...
+                    status=status,
+                )
+
+        async for done in _idle.loop(
+            status_on_demand(),
+            apps=apps,
+            wait_for_exact_units=wait_for_exact_units,
+            wait_for_units=wait_for_units,
+            idle_period=idle_period,
+        ):
```
james-garner-canonical, line 3320: Suggestion: simplify this for noobs like me ... btw at first I thought this could be simplified without changing anything else, but after making it to `_idle.loop` I realised that my original simplification was totally incorrect, which suggests that the current version is a bit tricky to follow ... the below suggestion assumes a later suggestion to refactor `_idle.loop` to use a class, which I think makes things a lot easier to follow -- it certainly does for me
```suggestion
        timing_checker = _idle.TimingChecker()
        while True:
            full_status = await self.get_status()
            check_status = _idle.check(
                raw_status,
                apps=apps,
                raise_on_error=raise_on_error,
                raise_on_blocked=raise_on_blocked,
                status=status,
            )
            done = timing_checker.check(
                check_status,
                apps=apps,
                wait_for_exact_units=wait_for_exact_units,
                wait_for_units=wait_for_units,
                idle_period=idle_period,
            )
```

juju/model/_idle.py
```diff
@@ -0,0 +1,215 @@
...
+
+    for unit_name, unit in units.items():
+        if unit.machine:
+            machine = full_status.machines[unit.machine]
+            assert isinstance(machine, MachineStatus)
+            assert machine.instance_status
+            if machine.instance_status.status == "error" and raise_on_error:
+                raise JujuMachineError(
+                    f"{unit_name!r} machine {unit.machine!r} has errored: {machine.instance_status.info!r}"
+                )
```
james-garner-canonical, line 137: Suggestion: make it clear up front that we're only interested in unit.machine -- there's no elif/else coming
```suggestion
        if not unit.machine:
            continue
        machine = full_status.machines[unit.machine]
        assert isinstance(machine, MachineStatus)
        assert machine.instance_status
        if machine.instance_status.status == "error" and raise_on_error:
            raise JujuMachineError(
                f"{unit_name!r} machine {unit.machine!r} has errored: {machine.instance_status.info!r}"
            )
```

juju/model/_idle.py
```diff
@@ -0,0 +1,215 @@
...
+    # Workload error (-"-)
+    # App error (any app from apps)
+    #
+    # Workload blocked (any unit of any app from apps)
+    # App blocked (any app from apps)
+    units: dict[str, UnitStatus] = {}
+
+    for app_name in apps:
+        units.update(_app_units(full_status, app_name))
+
```
james-garner-canonical, line 128: Suggestion: I really like how the loops below all do one thing, much nicer to read and follow than the original wait_for_idle. I notice that the first few only do anything if `raise_on_error`, while the next couple only do something if `raise_on_blocked`. How about something like:

```python
    if raise_on_error:
        _check_for_errors(full_status, apps, units)
    if raise_on_blocked:
        _check_for_blocked(full_status, apps, units)
```

juju/model/_idle.py
```diff
@@ -0,0 +1,215 @@
...
+            )
+
+    for app_name in apps:
+        app = full_status.applications[app_name]
+        assert isinstance(app, ApplicationStatus)
+        assert app.status
+        if app.status.status == "blocked" and raise_on_blocked:
+            raise JujuAppError(f"{app_name!r} is blocked: {app.status.info!r}")
+
+    rv = CheckStatus(set(), set(), set())
```
james-garner-canonical, line 174: Minor suggestion: in the `CheckStatus` definition, use `= field(default_factory=set)`. Reasoning: I think `rv = CheckStatus()` will look more natural to readers (like me =))

juju/model/_idle.py
```diff
@@ -0,0 +1,215 @@
...
+                logger.info(
+                    "Waiting for app %r units %s == %s",
+                    app_name,
+                    len(ready_units),
+                    wait_for_exact_units,
+                )
+                yield False
+                break
+        else:
+            yield True
```
james-garner-canonical, line 98: Suggestion: I really think that a class to encapsulate the state, and a single synchronous method call call with returns, is so much easier to follow than the async generator, and it seems like it's pretty straightforwardly logically equivalent -- just a lot easier to follow for noobs like me

```suggestion
class TimingChecker:
    """The outer, time-dependent logic of a wait_for_idle loop."""
    def __init__(self):
        self.idle_since: dict[str, float] = {}
    
    def check(
        status: CheckStatus,
        *
        apps: AbstractSet[str],
        wait_for_exact_units: int | None = None,
        wait_for_units: int,
        idle_period: float,
    ) -> bool:
        logger.info("wait_for_idle iteration %s", status)
        now = time.monotonic()

        if not status:
            return False

        expected_idle_since = now - idle_period

        # FIXME there's some confusion about what a "busy" unit is
        # are we ready when over the last idle_period, every time sampled:
        # a. >=N units were ready (possibly different each time), or
        # b. >=N units were ready each time
        for name in status.units:
            if name in status.idle_units:
                self.idle_since[name] = min(now, idle_since.get(name, float("inf")))
            else:
                self.idle_since[name] = float("inf")

        if busy := {n for n, t in self.idle_since.items() if t > expected_idle_since}:
            logger.info("Waiting for units to be idle enough: %s", busy)
            return False

        for app_name in apps:
            ready_units = [
                n for n in status.ready_units if n.startswith(f"{app_name}/")
            ]
            if len(ready_units) < wait_for_units:
                logger.info(
                    "Waiting for app %r units %s >= %s",
                    app_name,
                    len(status.ready_units),
                    wait_for_units,
                )
                return False
            if (
                wait_for_exact_units is not None
                and len(ready_units) != wait_for_exact_units
            ):
                logger.info(
                    "Waiting for app %r units %s == %s",
                    app_name,
                    len(ready_units),
                    wait_for_exact_units,
                )
                return False
        return True
```

james-garner-canonical (changes requested):
Exciting updates, glad tests are passing now.

I'm finding it hard to follow the logical changes between versions on this PR, I think primarily because the generator adds a fair bit of mental overhead for me -- it's kind of nicer now (though the new outer async loop construct took me some time to digest), but I still have to put in more mental work to follow the various `yield ... break`s and `yield ... continue`s.

For now I've made some requests for changes to make it easier for me to follow what's going on. Sorry if that seems like wasted effort -- it did take me a while, but I kind of had to do it anyway to follow the code.

Let me know if you want me to look closer at specific logical changes (looks like there are changes to the `idle_since` logic?). I still haven't looked in detail at your tests either, would you like further review there?

Also, what's the latest with running charmers' integration tests on the latest version? Are there still failures with the new wait_for_idle that pass with the old one?

dimaqq:
I've pushed a separate branch with this refactor only: https://github.com/juju/python-libjuju/pull/1245 PTAL

dimaqq:
closing in favour of #1245

PR #1231, opened 2024-12-02
chore: do not merge, only testing that the feature flag works

dimaqq:
A bunch of integration tests are expected to randomly fail.


dimaqq:
figured it out: `tox` ate my env var.

PR #1234, opened 2024-12-05
chore: [release 2.9.49.1] and fix tests

dimaqq:
#1235 + an attempt to fix tests

- [x] restrict Python versions (debatable)
- [x] restrict websockets in tox.ini and docs
- [x] homogenise dependency versions in setup.py, tox.ini, requirements.txt
- [ ] limit the charmhub-sourced charm revision to fix integration tests
- [ ] something else?


dimaqq:
@james-garner-canonical here's a branch that tries to do "as much as possible", which is not what we want after Ben's comment about discussion with Alex and Jon

dimaqq:
This is very much on the back-burner / may never get done.
We've released 2.9.49.1 as a hotfix even though 2.9 is under security maintenance only.

PR #1235, opened 2024-12-05
chore: restrict websockets version and release 2.9.49.1

dimaqq:
Ben brought this up at our stand-up this morning and we've decided to make a "blind" release, meaning:

- minimum possible change set
- not running tests
- not waiting for Jenkins pre-merge check (which is out of sync wrt. 2.9)

I've previously pushed https://test.pypi.org/project/juju/2.9.49.1/ along these lines, and James was volunteered to make the same release officially.

---

the key problem is that 2.9 branch uses asynctest for unit tests, and this library is old and does not support pyhton 3.12
meanwhile there's a single jira pre-merge job that's used for both 2.9 and main (3.6), and that job runs new Python.

basically 2.9 and 3.x have diverged so much that a single jira job cannot cover both

The rest can be fixed, and I have https://github.com/juju/python-libjuju/pull/1234 (draft) but that's a long road, because other stuff has changed (e.g. latest charm X revision got bumped on charmhub, which blocks specific integration test)


dimaqq:
@james-garner-canonical this is the content of https://test.pypi.org/project/juju/2.9.49.1/

docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 ---------
 
+2.9.49.1
+^^^^^^^^
+
+Friday, November 29th, 2024
```
james-garner-canonical, line 7: Thanks, the only thing I was going to change was this date. Want to apply this or should I take care of it?
```suggestion
Friday Dec 5 2024
```

james-garner-canonical: Or better to just have it match the test release?

james-garner-canonical (approved):

docs/changelog.rst
dimaqq: Let's match the test release.

dimaqq:
I've pushed the same files to https://pypi.org/project/juju/2.9.49.1/#files

anvial:
Discussed with Dima and decided that we could merge this PR, ignoring the action results.

"the key problem is that 2.9 branch uses asynctest for unit tests, and this library is old and does not support pyhton 3.12
meanwhile there's a single jira pre-merge job that's used for both 2.9 and main (3.6), and that job runs new Python."

PR #1242, opened 2024-12-17
chore: remove juju.loop, deprecated before 3.0

dimaqq:
juju.loop was deprecated before the 3.0 release.
finally remove it.


benhoyt:
Is this used/referenced in any of our charm's integration tests?

dimaqq:
Nope, not used anywhere at all. It’s a left-over from 2.x days.

james-garner-canonical (approved):
Looks good, loop isn't being used anywhere internally, and tests all pass.

I see that the only change besides deleting loop and its tests is updating the debian example code.

I did not realise this library had a debian package. I'm not sure how that's being tracked, but https://pypi.debian.net/juju has python-libjuju's latest releases. The package doesn't seem to be in the ubuntu repositories though, at least I can't find it on noble. I wonder what the story is here?

What to do about the debian package might be a bigger question than this, but I do wonder (a) if we *should* bump the required juju version in `debian/control` to 3.0 in main now, and (b) whether doing so has any affect on anything ...

PR #1245, opened 2024-12-19
feat: new wait for idle

dimaqq:
Same as #1219 but using dumb classes instead of async generator.

### DIFF
--- a/juju/model/_idle.py
+++ b/juju/model/_idle.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import logging
 import time
 from dataclasses import dataclass
-from typing import AsyncIterable
+from typing import AbstractSet, AsyncIterable
 
 from ..client._definitions import (
     ApplicationStatus,
@@ -35,7 +35,7 @@ class CheckStatus:
 async def loop(
     foo: AsyncIterable[CheckStatus | None],
     *,
-    apps: frozenset[str],
+    apps: AbstractSet[str],
     wait_for_exact_units: int | None = None,
     wait_for_units: int,
     idle_period: float,
@@ -101,7 +101,7 @@ async def loop(
 def check(
     full_status: FullStatus,
     *,
-    apps: frozenset[str],
+    apps: AbstractSet[str],
     raise_on_error: bool,
     raise_on_blocked: bool,
     status: str | None,
--- a/juju/model/_idle.py
+++ b/juju/model/_idle.py
@@ -44,7 +44,7 @@ async def loop(
     idle_since: dict[str, float] = {}
 
     async for status in foo:
-        logger.warning("FIXME unit test debug %r", status)
+        logger.info("wait_for_idle iteration %s", status)
         now = time.monotonic()
 
         if not status:
@@ -52,51 +52,50 @@ async def loop(
             continue
 
         expected_idle_since = now - idle_period
-        rv = True
 
         # FIXME there's some confusion about what a "busy" unit is
         # are we ready when over the last idle_period, every time sampled:
         # a. >=N units were ready (possibly different each time), or
         # b. >=N units were ready each time
         for name in status.units:
-            if name in status.ready_units:
+            if name in status.idle_units:
                 idle_since[name] = min(now, idle_since.get(name, float("inf")))
             else:
                 idle_since[name] = float("inf")
+
+        if busy := {n for n, t in idle_since.items() if t > expected_idle_since}:
+            logger.info("Waiting for units to be idle enough: %s", busy)
+            yield False
+            continue
 
         for app_name in apps:
             ready_units = [
                 n for n in status.ready_units if n.startswith(f"{app_name}/")
             ]
             if len(ready_units) < wait_for_units:
-                logger.warn(
+                logger.info(
                     "Waiting for app %r units %s >= %s",
                     app_name,
                     len(status.ready_units),
                     wait_for_units,
                 )
-                rv = False
+                yield False
+                continue
 
             if (
                 wait_for_exact_units is not None
                 and len(ready_units) != wait_for_exact_units
             ):
-                logger.warn(
+                logger.info(
                     "Waiting for app %r units %s == %s",
                     app_name,
                     len(ready_units),
                     wait_for_exact_units,
                 )
-                rv = False
+                yield False
+                continue
 
-        # FIXME possible interaction between "wait_for_units" and "idle_period"
-        # Assume that we've got some units ready and some busy
-        # What are the semantics for returning True?
-        if busy := [n for n, t in idle_since.items() if t > expected_idle_since]:
-            logger.warn("Waiting for %s to be idle enough", busy)
-            rv = False
-
-        yield rv
+        yield True
 
 
 def check(
@@ -175,7 +174,6 @@ def check(
     rv = CheckStatus(set(), set(), set())
 
     for app_name in apps:
-        ready_units = []
         app = full_status.applications[app_name]
         assert isinstance(app, ApplicationStatus)
         for unit_name, unit in _app_units(full_status, app_name).items():
@@ -183,16 +181,12 @@ def check(
             assert unit.agent_status
             assert unit.workload_status
 
-            if unit.agent_status.status != "idle":
-                continue
-            if status and unit.workload_status.status != status:
-                continue
+            if unit.agent_status.status == "idle":
+                rv.idle_units.add(unit_name)
 
-            ready_units.append(unit)
-            rv.ready_units.add(unit_name)
+            if not status or unit.workload_status.status == status:
+                rv.ready_units.add(unit_name)
 
-    # FIXME
-    # rv.idle_units -- depends on agent status only, not workload status
     return rv
 
 
--- a/juju/model/_idle.py
+++ b/juju/model/_idle.py
@@ -1,0 +1,221 @@
+# Copyright 2024 Canonical Ltd.
+# Licensed under the Apache V2, see LICENCE file for details.
+"""Implementation of Model.wait_for_idle(), analog to `juju wait_for`."""
+
+from __future__ import annotations
+
+import logging
+import time
+from dataclasses import dataclass
+from typing import AsyncIterable
+
+from ..client._definitions import (
+    ApplicationStatus,
+    FullStatus,
+    MachineStatus,
+    UnitStatus,
+)
+from ..errors import JujuAgentError, JujuAppError, JujuMachineError, JujuUnitError
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class CheckStatus:
+    """Return type check(), represents single loop iteration."""
+
+    units: set[str]
+    """All units visible at this point."""
+    ready_units: set[str]
+    """Units in good status (workload, agent, machine?)."""
+    idle_units: set[str]
+    """Units with stable agent status (FIXME details)."""
+
+
+async def loop(
+    foo: AsyncIterable[CheckStatus | None],
+    *,
+    apps: frozenset[str],
+    wait_for_exact_units: int | None = None,
+    wait_for_units: int,
+    idle_period: float,
+) -> AsyncIterable[bool]:
+    """The outer, time-dependents logic of a wait_for_idle loop."""
+    idle_since: dict[str, float] = {}
+
+    async for status in foo:
+        logger.warning("FIXME unit test debug %r", status)
+        now = time.monotonic()
+
+        if not status:
+            yield False
+            continue
+
+        expected_idle_since = now - idle_period
+        rv = True
+
+        # FIXME there's some confusion about what a "busy" unit is
+        # are we ready when over the last idle_period, every time sampled:
+        # a. >=N units were ready (possibly different each time), or
+        # b. >=N units were ready each time
+        for name in status.units:
+            if name in status.ready_units:
+                idle_since[name] = min(now, idle_since.get(name, float("inf")))
+            else:
+                idle_since[name] = float("inf")
+
+        for app_name in apps:
+            ready_units = [
+                n for n in status.ready_units if n.startswith(f"{app_name}/")
+            ]
+            if len(ready_units) < wait_for_units:
+                logger.warn(
+                    "Waiting for app %r units %s >= %s",
+                    app_name,
+                    len(status.ready_units),
+                    wait_for_units,
+                )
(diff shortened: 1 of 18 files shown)


juju/model/_idle.py
```diff
@@ -0,0 +1,217 @@
...
+
+        if not status:
+            return False
+
+        expected_idle_since = now - self.idle_period
+
+        # FIXME there's some confusion about what a "busy" unit is
+        # are we ready when over the last idle_period, every time sampled:
+        # a. >=N units were ready (possibly different each time), or
+        # b. >=N units were ready each time
```
james-garner-canonical: If I'm understanding the current logic correctly, we currently have (b), right? Because at time `now`, we're checking if there are >=N units that have been ready for long enough. I think this is what people would expect `idle_period` to mean -- each unit is only counted as ready if its agent has been idle for long enough.

The logic in original `wait_for_idle` seems to be the same (as far as I'm following it), so it seems like you're doing the right thing here.

juju/model/_idle.py
```diff
@@ -0,0 +1,217 @@
...
+            raise JujuUnitError(
+                f"{unit_name!r} workload is blocked: {unit.workload_status.info!r}"
+            )
+
+    for app_name in apps:
+        app = full_status.applications[app_name]
+        assert isinstance(app, ApplicationStatus)
+        assert app.status
+        if app.status.status == "blocked" and raise_on_blocked:
+            raise JujuAppError(f"{app_name!r} is blocked: {app.status.info!r}")
```
james-garner-canonical: I still quite like the idea of doing something like this:
```python
    if raise_on_error:
        _check_for_errors(full_status, apps, units)
    if raise_on_blocked:
        _check_for_blocked(full_status, apps, units)
```

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+            "mysql-test-app/0",
+            "mysql-test-app/1",
+        },
+        {
+            "grafana-agent-k8s/0",
+            "hexanator/0",
+            "mysql-test-app/0",
+            "mysql-test-app/1",
+        },
+    )
```
james-garner-canonical: What do you think about using keyword arguments with `CheckStatus`? Reading on github, I forget the order of the arguments haha ... of course in this case all three are identical, but actually I've forgotten what the arguments even are -- and it would definitely be very helpful where the values are different. Not a blocker for tests I reckon, but it could be nice for our future selves

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+        {"mysql-test-app/0", "mysql-test-app/1"},
+        {"mysql-test-app/0", "mysql-test-app/1"},
+    )
+
+
+def test_active_units(full_status: FullStatus, kwargs):
+    kwargs["apps"] = ["mysql-test-app"]
+    kwargs["status"] = "active"
+    status = check(full_status, **kwargs)
+    assert status == CheckStatus({"mysql-test-app/0", "mysql-test-app/1"}, set(), ANY)
```
james-garner-canonical: here's a case where keyword arguments would definitely be handy

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+
+
+def convert(data: dict[str, Any]) -> FullStatus:
+    return _convert_response(data, cls=FullStatus)
+
+
+@pytest.fixture
+def full_status(response) -> FullStatus:
+    return convert(response)
```
james-garner-canonical, line 274: nit: fixtures at the top?

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+            "grafana-agent-k8s/0",
+            "hexanator/0",
+            "mysql-test-app/0",
+            "mysql-test-app/1",
+        },
+    )
+
+
+def test_check_status_missing_app(full_status: FullStatus, kwargs):
+    kwargs["apps"] = ["missing", "hexanator"]
```
james-garner-canonical, line 29: Where `kwargs` is being modified before calling, what do you think about repeating yourself and just defining it in the test?
```suggestion
    kwargs = {
        apps: ["missing", "hexanator"],
        raise_on_error: False,
        raise_on_blocked: False,
        status: None,
    }
```

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def test_missing_app(full_status: FullStatus, kwargs):
+    kwargs["apps"] = ["missing"]
+    status = check(full_status, **kwargs)
+    assert status is None
+
+
+def test_no_units(response: dict[str, Any], kwargs):
+    response["response"]["applications"]["hexanator"]["units"].clear()
```
james-garner-canonical, line 53: I wouldn't suggest defining the response inline though since it's massive, so I guess we're going to be using this pattern anyway

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+    kwargs["apps"] = ["hexanator"]
+    status = check(convert(response), **kwargs)
+    assert status == CheckStatus(set(), set(), set())
+
+
+def test_app_error(response: dict[str, Any], kwargs):
+    app = response["response"]["applications"]["hexanator"]
+    app["status"]["status"] = "error"
+    app["status"]["info"] = "big problem"
```
james-garner-canonical, line 62: xD

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+        check(convert(response), **kwargs)
+
+    assert "hexanator/0" in str(e)
+    assert "workload problem" in str(e)
+
+
+def test_machine_ok(response: dict[str, Any], kwargs):
+    app = response["response"]["applications"]["hexanator"]
+    app["units"]["hexanator/0"]["machine"] = "42"
+    # https://github.com/dimaqq/juju-schema-analysis/blob/main/schemas-juju-3.5.4.txt#L3611-L3674
```
james-garner-canonical: 404

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def test_machine_error(response: dict[str, Any], kwargs):
+    app = response["response"]["applications"]["hexanator"]
+    app["units"]["hexanator/0"]["machine"] = "42"
+    response["response"]["machines"] = {
+        "42": {
+            "instance-status": {
+                "status": "error",
+                "info": "Battery low. Try a potato?",
```
james-garner-canonical, line 174: xD

tests/unit/test_idle_check.py
```diff
@@ -0,0 +1,269 @@
...
+    app["units"]["hexanator/0"]["workload-status"]["status"] = "blocked"
+    app["units"]["hexanator/0"]["workload-status"]["info"] = "small problem"
+
+    kwargs["apps"] = ["hexanator"]
+    kwargs["raise_on_blocked"] = True
+
+    with pytest.raises(JujuUnitError) as e:
+        check(convert(response), **kwargs)
+
+    assert "small problem" in str(e)
```
james-garner-canonical, line 213: Maybe overkill, and maybe you already have it and I've missed it, but what do you think about duplicating these tests with `"raise_on_blocked" = False`?

Similarly, are there any tests that cover the error case when `"raise_on_error" = False`?

tests/unit/test_idle_loop.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2024 Canonical Ltd.
+# Licensed under the Apache V2, see LICENCE file for details.
+from __future__ import annotations
+
+from typing import AbstractSet, Iterable
+
+from freezegun import freeze_time
```
james-garner-canonical, line 7: What do you think about making `now` an argument to `Loop.next` to eliminate this dependency? The nice `unroll` function you have here could just take an iterable of times as well as statuses for minimal refactoring of tests, which would (imo) be nicer than using the various freeze_gun commands. Not a blocker tho by any means

tests/unit/test_idle_loop.py
```diff
@@ -0,0 +1,121 @@
...
+    return [loop.next(s) for s in statuses]
+
+
+def test_wait_for_apps():
+    def checks():
+        yield None
+        yield None
+
+    assert unroll(
+        checks(),
```
james-garner-canonical, line 35:
```suggestion
def test_wait_for_apps():
    assert unroll(
        [None, None],
```

In general a plain list seems nicer than defining a generator, though I see the generator is doing extra work with `freeze_gun` in some cases -- though that would be unnecessary if we made time an argument too.

tests/unit/test_model.py
```diff
@@ -251,27 +251,28 @@ async def test_no_args(self):
...
 
-        with self.assertRaises(JujuError):
+        with self.assertRaises(TypeError):
             # apps arg has to be a List[str]
-            await m.wait_for_idle(apps=3)
+            await m.wait_for_idle(apps=3)  # type: ignore
 
-        with self.assertRaises(JujuError):
+        with self.assertRaises(TypeError):
             # apps arg has to be a List[str]
```
james-garner-canonical, line 265: I remember noticing this change before, though I'm not sure we talked about it -- is there any benefit to changing the error type here?

james-garner-canonical (commented):
Thanks, it probably seems silly but this is significantly easier for me to follow. `check` and `Loop.next` look good to me.

I've commented on your `FIXME` note -- also a minor refactor suggestion to put the the error checking loops behind a `raise_on_error` flag (and likewise for blocked).

Also a few nitpicks on the tests just to do my due diligence, but I'm guessing you're fine with things as is. The most significant note is that it would probably be worth duplicating the `check` tests for blocked/errors to cover the `"raise_on_{blocked,error}" = False` cases (unless you've already covered that and I've just missed it).

james-garner-canonical:
I notice there are a couple of failing integration tests (they all passed last run in #1219) -- related to this refactor, or just more test flakiness?

tests/unit/test_idle_check.py
dimaqq: good catch, fixed.

tests/unit/test_model.py
dimaqq: Only that it's clearer where the error originates: client-side or Juju.

tests/unit/test_idle_loop.py
dimaqq: I'm not a fan of Java-style dependency injection in Python, so -1 on making current time an explicit argument.

tests/unit/test_idle_loop.py
dimaqq: I'd rather not.

james-garner-canonical: Fair enough, I don't think its necessary at all, just an idea

dimaqq:
failed (quaranteed) integration test:
> juju.errors.JujuError: refreshing charm "ch:amd64/focal/hello-juju-8": unexpected content-type from server "text/html"

and stable integration test:
> Post "https://api.charmhub.io/v2/charms/refresh": net/http: TLS handshake timeout

looks like charmhub had a glitch, restarted.

tests/unit/test_idle_loop.py
james-garner-canonical: I don't know if I really see passing the timestamp of the current status as an argument as dependency injection, but I'm fine with not making it an argument

juju/model/_idle.py
dimaqq: done

tests/unit/test_idle_check.py
dimaqq: adding but only once at the top to serve as a reference.

juju/model/_idle.py
dimaqq: comment was out of date.
actually it's: N ready units and separately from that all units have been continuously idle.

tests/unit/test_idle_check.py
dimaqq: on the second thought, went over tests and added kwargs in places.

tests/unit/test_idle_check.py
dimaqq: done

tests/unit/test_idle_check.py
dimaqq: actually I've dropped them to the bottom on purpose.
my thinking was "most important / high level" things at the top.

tests/unit/test_idle_check.py
dimaqq: Doable, but that gets rather tedious / blows up the code size.

tests/unit/test_idle_check.py
dimaqq: added, thanks

tests/unit/test_model.py
james-garner-canonical: That's a good enough reason for me =)

tests/unit/test_idle_check.py
james-garner-canonical: Totally reasonable. I think I scrolled down to find out what was in `kwargs`

juju/model/_idle.py
james-garner-canonical: Ah, cool, I think that's also totally reasonable -- the apps have been idle for requested time, and the specified number of units are in whatever ready status the user requested

dimaqq:
tests passed. https://github.com/juju/python-libjuju/actions/runs/12404794817

dimaqq:
@james-garner-canonical I've implemented most suggestions, pelase review :)

dimaqq:
real charm CI runs:
- https://github.com/canonical/mysql-router-operator/actions/runs/12409228356?pr=196
- https://github.com/dimaqq/seldon-core-operator/actions/runs/12409256210

james-garner-canonical (approved):
Looks great! Real charm CI runs are reassuring too.

Interesting how [this charm tests run](https://github.com/canonical/mysql-router-operator/actions/runs/12409228356?pr=196) has two timeouts with *old* wait for idle in `test_upgrade_from_edge` (3.4/stable on focal and jammy) -- which both passed with new wait for idle. Since it's a timeout I guess this could be a flaky test generally, but I do wonder if this indicates something -- was old wait for idle more eager to bail out with a timeout error rather than check everything? I don't think this necessarily needs further investigation right now, just noting it

PR #1256, opened 2025-03-18
fix: dummy change to test CI

dimaqq:


dimaqq:
Indeed it fails.
I've opened https://bugs.launchpad.net/anbox-cloud/+bug/2103485 for the anbox bundle

PR #1044, opened 2024-04-17
Fix issue with microk8s proxy

Aflynn50:
The temporary file used to store the ca_cert was set to delete itself on close. This meant that when it was accessed multiple times it would no longer be present.

Set to not delete on close and to remove file when the proxy is deleted.

#### QA Steps
```
juju bootstrap microk8s
juju add-model default
juju deploy cos-lite
```
In python
```
from juju import model
m = model.Model()
await m.connect()
await m.create_offer("grafana:grafana-dashboard", "grafana-dashboards")
await m.disconnect()
```
Check all resource in /tmp have been cleaned up

Fixes #1040

PR #1265, opened 2025-05-21
fix: creating offer when connecting to existing model, fixes #1264

### LINKED ISSUE #1264: Creating offer fails when connecting to already existing model
### Description

I am seeing the same error as described in #1040, however only when trying to create an offer with a `Model` object instantiated manually and connected to an already existing Juju model. For example, the following fails with a `FileNotFoundError: [Errno 2] No such file or directory` error thrown while loading the CA file:

```python
model = Model()
await model.connect(model_that_already_exists)
await model.create_offer("postgresql-k8s:database", offer_name="pgsql")
```

However, the following works:

```python
controller = Controller()
await controller.connect(controller_name)
model = await controller.add_model(new_model_name)
model.create_offer("postgresql-k8s:database", offer_name="pgsql")
```

It appears that the [PR](https://github.com/juju/python-libjuju/pull/1044) that fixed this issue was later reverted by [a8640247ae8dbc1aef4546da20c35c0e1d1f51e7](https://github.com/juju/python-libjuju/commit/a8640247ae8dbc1aef4546da20c35c0e1d1f51e7)

Full traceback can be found here (sorry for the poor wrapping): [https://pastebin.ubuntu.com/p/4TqMNhgh89/](https://pastebin.ubuntu.com/p/4TqMNhgh89/)

### Urgency

Casually reporting

### Python-libjuju version

edge

### Juju version

3.6.5-genericlinux-amd64

### Reproduce / Test

```python
model = Model()
await model.connect(model_that_already_exists)
await model.create_offer("postgresql-k8s:database", offer_name="pgsql")
```

dimaqq:
#### Description

#1044 fixed proxy object reuse some time ago
I accidentally undid that change in #1186 as I didn't realise that multiple connect/close calls were expected on a given instance.
This PR re-applies the change, and cleans up init.

(I've rearranged the initialisation order, because `__del__()` is run on a partially-initialised instance if `__init__()` fails half-way through. That's something that happens in unit tests, although unlikely to affect legitimate use.)

Fixes: #1264 creating offer fails when connecting to existing model

#### QA Steps

- No regression in unit or integration tests
- Eyeball the code change

All CI tests need to pass.

Fixes #1264

### DIFF
--- a/juju/client/proxy/kubernetes/proxy.py
+++ b/juju/client/proxy/kubernetes/proxy.py
@@ -39,7 +39,7 @@ class KubernetesProxy(Proxy):
             raise ValueError(f"Invalid port number: {remote_port}")
 
         if ca_cert:
-            with tempfile.NamedTemporaryFile(delete=True) as f:
+            with tempfile.NamedTemporaryFile(delete=False) as f:
                 f.write(bytes(ca_cert, "utf-8"))
             self.temp_ca_file = config.ssl_ca_cert = f.name
 
--- a/juju/client/proxy/kubernetes/proxy.py
+++ b/juju/client/proxy/kubernetes/proxy.py
@@ -1,6 +1,7 @@
 # Copyright 2023 Canonical Ltd.
 # Licensed under the Apache V2, see LICENCE file for details.
 import logging
+import os
 import tempfile
 
 from kubernetes import client
@@ -29,21 +30,18 @@ class KubernetesProxy(Proxy):
         self.namespace = namespace
         self.remote_port = remote_port
         self.service = service
+        self.temp_ca_file = None
+        self.port_forwarder = None
 
         try:
             self.remote_port = int(remote_port)
         except ValueError:
             raise ValueError(f"Invalid port number: {remote_port}")
 
-        self.port_forwarder = None
-
         if ca_cert:
-            self.temp_ca_file = tempfile.NamedTemporaryFile()  # noqa: SIM115
-            self.temp_ca_file.write(bytes(ca_cert, "utf-8"))
-            self.temp_ca_file.flush()
-            config.ssl_ca_cert = self.temp_ca_file.name
-        else:
-            self.temp_ca_file = None
+            with tempfile.NamedTemporaryFile(delete=True) as f:
+                f.write(bytes(ca_cert, "utf-8"))
+            self.temp_ca_file = config.ssl_ca_cert = f.name
 
         self.api_client = client.ApiClient(config)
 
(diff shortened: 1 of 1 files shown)


dimaqq:
Nevermind all integration tests failing, that's the case in the main branch too: https://github.com/juju/python-libjuju/commits/main/

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -29,21 +30,18 @@ def __init__(
...
         if ca_cert:
-            self.temp_ca_file = tempfile.NamedTemporaryFile()  # noqa: SIM115
-            self.temp_ca_file.write(bytes(ca_cert, "utf-8"))
-            self.temp_ca_file.flush()
-            config.ssl_ca_cert = self.temp_ca_file.name
-        else:
-            self.temp_ca_file = None
+            with tempfile.NamedTemporaryFile(delete=False) as f:
+                f.write(bytes(ca_cert, "utf-8"))
+            self.temp_ca_file = config.ssl_ca_cert = f.name
```
benhoyt: This would be clearer as two separate assignments:
```suggestion
            self.temp_ca_file = f.name
            config.ssl_ca_cert = f.name
```

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -67,13 +65,16 @@ def connect(self):
 
     def __del__(self):
         self.close()
+        try:
+            if self.temp_ca_file:
+                os.unlink(self.temp_ca_file)
+        except FileNotFoundError:
```
benhoyt: I don't think we need this exception handling. Python already ignores (and prints to stderr) exceptions that are raised in `__del__`, for example:

```python
Exception ignored in: <function C.__del__ at 0x72cfa9b4b6a0>
  File "<stdin>", line 4, in __del__
RuntimeError: problem!
```

dimaqq:
Asking @benhoyt to review while @james-garner-canonical is away.

juju/client/proxy/kubernetes/proxy.py
```diff
@@ -29,21 +30,18 @@ def __init__(
         self.namespace = namespace
         self.remote_port = remote_port
         self.service = service
+        self.temp_ca_file = None
```
benhoyt: Now that this is a filename/path and not a file object, how about we call this `temp_ca_path` to avoid confusion?

benhoyt (approved):
This looks good to me. Couple of nit comments. Let's just ensure we've tested this code path in real life (manually is okay).

dimaqq:
I've merged #1268 into this PR.

juju/client/proxy/kubernetes/proxy.py
dimaqq: done.

juju/client/proxy/kubernetes/proxy.py
dimaqq: done.

juju/client/proxy/kubernetes/proxy.py
dimaqq: done.

benhoyt (approved):

gfouillet (approved):
That's cleaner, LGTM.

dimaqq:
I've confirmed that the fix works manually.

Issue #1267, opened 2025-05-21
Integration tests are failing on main


dimaqq:
### Description

- `test_subordinate_units` fails across all Juju versions
   - complains of `ntp/0` in error / hook failed: "install"
   - I think the intention for `ntp` is to be subordinate to `ubuntu`
- eight more tests fail on Juju 3.3 specifically
   - all with the same `juju.errors.JujuError: base: ubuntu@24.04/stable`
   - perhaps the charm is resolved to noble, but noble is not supported?

```
tests/integration/test_application.py::test_deploy_charmhub_charm
tests/integration/test_application.py::test_trusted
tests/integration/test_application.py::test_app_destroy
tests/integration/test_application.py::test_app_remove_wait_flag
tests/integration/test_application.py::test_app_charm_name
tests/integration/test_secrets.py::test_grant_secret
tests/integration/test_unit.py::test_scp
tests/integration/test_unit.py::test_ssh
tests/integration/test_unit.py::test_subordinate_units
```

### Urgency

Annoying bug in our test suite

### Python-libjuju version

main

### Juju version

all

### Reproduce / Test

```python
https://github.com/juju/python-libjuju/commits/main/ starting in May this year
```

dimaqq:
A snapshot from production (IS only) shows that 3.0, 3.1 and 3.3 are no longer used.

PR #1268, opened 2025-05-21
ci: disable test that fails on all Juju versions

dimaqq:
#### Description

Disable the test that fails on all Juju versions.

I suspect that it's due base incompatibility:
- `ubuntu` goes first and is available on 24.04
- `ntp` goes next, but it tops out at 22.04 on the default channel

Ref: #1267

### DIFF
--- a/tests/integration/test_unit.py
+++ b/tests/integration/test_unit.py
@@ -254,6 +254,8 @@ async def test_unit_introspect():
         )
 
 
+# Probably a mismatch between `ubuntu` available for 24.04 and `ntp` only for 22.04
+@pytest.mark.xfail
 @base.bootstrapped
 async def test_subordinate_units():
     async with base.CleanModel() as model:


dimaqq:
This PR only addresses the one test that fails across all juju versions.
It does not address the other 8 tests that fail on Juju 3.3 specifically.

benhoyt (approved):
Change looks okay to me, thanks.

Integration tests are now failing with a couple of "juju.errors.JujuError: base: ubuntu@24.04/stable" errors. Boy, that's not a very helpful error message, is it? Are we missing part of the error?

gfouillet (approved):

dimaqq:
@benhoyt nope, that's that's the whole error that Juju 3.3 sends in the jRPC response. It basically means something like "field `base`, bad value `ubuntu@24.04/stable`".

PR #1269, opened 2025-05-23
ci: drop Juju 3.3

dimaqq:
Juju 3.3 has reached end of life.

> 🔸 Juju 3.3.7 - 10 September 2024
> NOTE: This is the last release of 3.3. There will be no more releases.

https://documentation.ubuntu.com/juju/latest/reference/juju/juju-roadmap-and-releases/index.html?dfghjkl=#juju-3-3

Meanwhile, 8 integration tests are failing against Juju 3.3 specifically #1267 

This PR removed Juju 3.3 from CI to close #1267

PR #1276, opened 2025-07-10
fix: don't strip valid characters from secret data

### LINKED ISSUE #1275: create_secret_data() strips valid trailing characters from secret values
### Description

The [create_secret_data](https://github.com/juju/python-libjuju/blob/d61ea7dde28f41766a917fd818816d53aa73e8c2/juju/secrets.py#L20) function removes the last character of the secret data if that's a capital "P", capital "Q" or an "=". 
I assume this is done to remove padding from base64 encoded strings. But the function doesn't seem to check whether the string it got is base64 encoded, it just assumes that.
But even if it gets a base64 encoded string it should not remove "P" nor "Q".
The root cause is the `stripped = val.rstrip(base64.b64encode(b"=").decode("utf-8"))` call which evaluates to `stripped = val.rstrip("PQ==")`.
I think this should be `stripped = val.rstrip("=")` if the data is a base64 encoded string.

Currently this is breaking the CI in multiple projects where a secret is created in integration tests then used in the charm under test. Here is an example in Vault where the CI tries to authorize the charm with a Vault token that ends with a "Q" and then it fails to do so when it fetches the token from a secret as the token won't have that "Q" at the end making it invalid.

I marked its urgency as "Annoying bug in our test suite" but I really believe it's more than that.

### Urgency

### Python-libjuju version

3.6.1.1

### Juju version

3.6.8

### Reproduce / Test

```python
import logging
import base64
from juju.errors import JujuError


import pytest

logger = logging.getLogger(__name__)


@pytest.mark.abort_on_fail
async def test_reproduce_error(
    ops_test: OpsTest
):
    assert ops_test.model
    root_token = "pbYXY9Aojkwsd6pwDbpfRMXQ"
    
    try:
        logger.info("Adding secret with token: %s", root_token)
        secret = await ops_test.model.add_secret(
            f"token", [f"token={root_token}"]
        )
        logger.info("Secret added: %s", secret)
        secret_id = secret.split(":")[-1]
        logger.info("Secret ID: %s", secret_id)

        list_secrets = await ops_test.model.list_secrets(show_secrets=True)
        
        for secret_obj in list_secrets:
            if secret_obj.label == "token" and secret_obj.value and secret_obj.value.data:
                token_base64 = secret_obj.value.data.get("token")
                if token_base64:
                    try:
                        decoded_token = base64.b64decode(token_base64).decode('utf-8')
                        logger.info("Decoded token: %s", decoded_token)
                    except Exception as decode_error:
                        logger.error("Error decoding token: %s", decode_error)
                break
        
    except JujuError as e:
        logger.error("Error adding secret: %s", e)
```

james-garner-canonical:
#### Description

`create_secret_data` unnecessarily attempts to strip trailing `=` symbols from secret data, and inadvertently strips all trailing `P`, `Q` and `=` characters instead. This PR drops the unnecessary removal of trailing characters, and adds unit tests covering cases that would previously have been handled badly (as well as some test cases validating existing behaviour that wasn't covered).

Fixes: #1275

#### QA Steps

Unit test should pass. Locally I have `tests/unit/test_jujudata.py::TestJujuData::test_verify_controller_uninitialized` failing both on `main` and on this branch.

### DIFF
--- a/tests/unit/test_secrets.py
+++ b/tests/unit/test_secrets.py
@@ -11,18 +11,29 @@ import pytest
 from juju.secrets import create_secret_data, read_secret_data
 
 
-class TestCreateSecretData(unittest.TestCase):
-    def test_bad_key(self):
+class TestCreateSecretData:
+    @pytest.mark.parametrize("keyval", ["foo", "=bar", "baz=", "f=bar", "fo=bar", "foo_bar=baz"])
+    def test_bad_key(self, keyval: str):
         with pytest.raises(ValueError):
-            create_secret_data(["foo"])
-        with pytest.raises(ValueError):
-            create_secret_data(["=bar"])
+            create_secret_data([keyval])
 
-    def test_good_key_values(self):
-        self.assertDictEqual(
-            create_secret_data(["foo=bar", "hello=world", "goodbye#base64=world"]),
-            {"foo": "YmFy", "hello": "d29ybGQ=", "goodbye": "world"},
-        )
+
+    @pytest.mark.parametrize(
+        "keyval,expected_key,expected_val",
+        [
+            ("foo=bar", "foo", "YmFy"),
+            ("hello=world", "hello", "d29ybGQ="),
+            ("goodbye#base64=world", "goodbye", "world"),
+            ("equalsign==", "equalsign", "PQ=="),
+            ("equalsign#base64=PQ==", "equalsign", "PQ=="),
+            ("pq-identity-theorem=P===Q", "pq-identity-theorem", "UD09PVE="),
+        ]
+    )
+    def test_goo_key_values(self, keyval: str, expected_key: str, expected_val: str):
+        actual = create_secret_data([keyval])
+        expected = {expected_key: expected_val}
+        assert actual == expected
+
 
     def test_key_content_too_large(self):
         with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


tests/unit/test_secrets.py
```diff
@@ -50,15 +61,12 @@ def test_secret_key_from_file(self):
 
             data = create_secret_data(["key1=value1", f"key2#file={file_path}"])
 
-            self.assertDictEqual(
```
james-garner-canonical, line 53: Just changing to `assert data ==`, which results in dedenting the assertion data.

tests/unit/test_secrets.py
```diff
@@ -11,18 +11,29 @@
 from juju.secrets import create_secret_data, read_secret_data
 
 
-class TestCreateSecretData(unittest.TestCase):
```
james-garner-canonical, line 14: Drop mostly unused `unittest.TestCase` base class so that `pytest.mark.parametrize` can work.

dimaqq (approved):
Code LGTM

let’s get clarity on failing integration tests vs main

james-garner-canonical:
Manual run on `main` [in progress](https://github.com/juju/python-libjuju/actions/runs/16207653428), let's see how it goes.

james-garner-canonical:
Summary: it's the same on main.

---

The non-quarantined integration test failures in this branch are all with Juju 3.3. #1267 noted multiple test failures in `main` with 3.3 specifically in May this year and #1269 proposed to drop 3.3 from without pushback (still unmerged due to ideas of dropping more). The same failures are present on `main` (and actually we get an extra failure on `main` [in one case](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493128), which I suspect is just random flakiness rather than anything to do with this PR).

Quarantined integration tests are failing `test_relate` on all Juju versions other than 3.1. The jobs are still running on `main`, but that specific test has already failed ([3.3](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493066), [3.4](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493074), [3.5](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493071), [3.6](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493072)). I've opened #1277 to document the `test_relate` failure.

Additionally, several more quarantined integration tests fail on 3.3 specifically (3.4, 3.5, and 3.6 only fail `test_relate`). This is also the case on [main](https://github.com/juju/python-libjuju/actions/runs/16207653428/job/45761493066).

PR #1278, opened 2025-07-11
chore: release 3.6.1.3

### LINKED ISSUE #1275: create_secret_data() strips valid trailing characters from secret values

james-garner-canonical:
#### Description

Version bump and changelog update for release.

Fixes: #1275

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,12 +1,17 @@
 Changelog
 ---------
 
+3.6.1.3
+^^^^^^^
+
+Friday 11th Jul 2025
+
+* fix: don't strip valid characters from secret data by @james-garner-canonical in https://github.com/juju/python-libjuju/pull/1276
+
 3.6.1.2
 ^^^^^^^
 
 Monday 26th May 2025
-
-## What's Changed
 
 * chore: hint machine addresses as list of dicts by @gboutry in https://github.com/juju/python-libjuju/pull/1258
 * ci: disable test that fails on all Juju versions by @dimaqq in https://github.com/juju/python-libjuju/pull/1268
@@ -16,8 +21,6 @@ Monday 26th May 2025
 ^^^^^^^
 
 Wednesday 19th Mar 2025
-
-## What's Changed
 
 * chore: fix changelog format by @dimaqq in https://github.com/juju/python-libjuju/pull/1248
 * docs: move docs from juju.is by @tmihoc in https://github.com/juju/python-libjuju/pull/1244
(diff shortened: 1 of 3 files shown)


dimaqq (approved):

james-garner-canonical:
Integration test failures are the same as on `main`.

PR #1282, opened 2026-07-22
chore: document deprecation in preparation for repo archival

james-garner-canonical:
This PR documents that `python-libjuju` is deprecated, and that charms should use Jubilant for integration tests.

- Github: `README.rst` is a symlink to `docs/readme.rst`, which now has a deprecation notice at the top.
- RTD: `docs/index.rst` includes the readme, so the notice shows on the docs homepage too ([preview](https://pythonlibjuju--1282.org.readthedocs.build/en/1282/))

Additionally the readme is update to fix a slight formatting error (missing newline) and document 24.04 as the last supported base.

---

Please use the text above this line as the commit message.

See below for additional context.

---

I experimented with having a deprecation notice on every docs page, which worked fine locally, but I couldn't get it past RTD's incremental docs builder to show up in the preview build, so I've gone with just the readme + homepage update.

---

Pushing this information to PyPI is not included in this PR, as that would require a new release of `python-libjuju`, which probably sends the wrong message. If we did want to push the notice to PyPI, we'd need to bump the library version and make a library release -- typically this would be a `.post0` release, as it's metadata only, though we'd want to dig and make sure that no one is implementing any funky version parsing (the current version is `3.6.1.3`, which appears to be `juju-major.juju-minor.juju-patch.python-libjuju-release-counter`, and is exposed as `juju.version.CLIENT_VERSION`).

If we did want to make a library release, we could consider updating the package docstring to include the deprecation notice, and perhaps even adding a runtime deprecation warning (though this is probably a bad idea since use of `python-libjuju` isn't confined to integration tests).

Ultimately I don't recommend making a library release at this time.

---

The integration test failures are certainly unrelated, and I refuse to look at them.

The [Jenkins check URL](https://jenkins.juju.canonical.com/job/github-check-merge-juju-python-libjuju/1154/)  returns "Bad Request" after 2FA. I don't think it's worth trying to fix Jenkins for this project (again), so if we can bypass that check to merge that would be great.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -305,30 +305,16 @@ texinfo_documents = [
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 # texinfo_no_detailmenu = False
 
-
 def setup(app):
     app.add_css_file("custom.css")
-    app.connect("source-read", _prepend_deprecation_notice)
-
 
 # Notice prepended to every built page so that visitors arriving on any page
 # (e.g. via a search engine or a deep link) see the deprecation warning.
-# We use a custom source-read hook rather than rst_prolog to avoid showing the
-# notice twice on the landing page, as index.rst includes readme.rst.
-_DEPRECATION_NOTICE = """\
+rst_prolog = """
 .. warning::
 
    **python-libjuju is deprecated.**
 
    Charms should migrate to `Jubilant <https://github.com/canonical/jubilant>`_, Canonical's recommended library for Juju integration tests.
    See `How to migrate integration tests from pytest-operator <https://canonical.com/juju/docs/ops/latest/howto/migrate/migrate-integration-tests-from-pytest-operator/>`_.
-
 """
-
-
-def _prepend_deprecation_notice(app, docname, content):
-    if docname in ("index", "readme"):
-        return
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -308,3 +308,27 @@
...
+
+"""
+
+
+def _prepend_deprecation_notice(app, docname, content):
+    if docname in ("index", "readme"):
+        return
+    # ``content`` is a list containing a single string to make in-place modification possible.
+    # https://www.sphinx-doc.org/en/master/extdev/event_callbacks.html#event-source-read
+    content[0] = _DEPRECATION_NOTICE + content[0]
```
james-garner-canonical: I tested this locally and it works perfectly. I thought it might be worthwhile since we're considering not taking down / redirecting the documentation site based on traffic. However, we could drop this and only have the warning in the `readme.rst`, which does render on the documentation homepage.

james-garner-canonical: Doesn't work on RTD due to incremental build shenanigans, simplifying.

dwilding (approved):
I agree with this approach. Thanks for the explanation.

benhoyt (approved):

